PR #5, opened 2026-07-14
feat(auth-service): admin user-creation API and role/admin seed bootstrap

Dharanishwaran-ND:
## What

Adds admin-provisioned user creation to `auth-service` plus a seed bootstrap for roles and the initial admin.

### API
- **`POST /api/v1/users`** — admin-only (`authenticate` + `requireRoles('ADMIN')`). Creates a user and their initial role assignment in one transaction.
  - Request: `{ mobileNumber, password, displayName, email?, roleCode, projectId?, geographyUnitId? }`
  - Fields follow the `users` table (ERD Appendix A.1); `roleCode`/`projectId`/`geographyUnitId` drive the `user_roles` insert.
  - Duplicate mobile/email → `409`; unknown/inactive role → `404`; `passwordHash` is never returned.
- Extracted the shared `mobileNumber` zod schema so login and create-user validate identically.

### Seed (`prisma/seed.ts`)
- **Roles**: seeded only when the `roles` table is empty (7 codes from the ERD `roles.role_code` enum). Left untouched once present.
- **Admin bootstrap**: initial admin created from `ADMIN_MOBILE_NUMBER` / `ADMIN_PASSWORD` env vars, only when that user doesn't already exist; never rotates an existing admin's password. Skips gracefully if the env vars are unset.
- **Test user**: distinct number + same create-if-absent guard.
- `.env.example` documents the new admin env vars (placeholder values only; `.env` stays gitignored).

## Verification
- `nx run-many -t build/test/lint` — **all 18 projects pass** (fresh Prisma clients, `--skip-nx-cache`).
- auth-service: **20/20** unit tests (5 new for `createUser`).
- Verified end-to-end against a local DB: roles + admin seeded, admin login returns a valid `ADMIN`-scoped JWT.

Related to #87

### DIFF
--- a/apps/auth-service/prisma/seed.ts
+++ b/apps/auth-service/prisma/seed.ts
@@ -3,6 +3,12 @@ import { PrismaClient } from '../../../node_modules/.prisma/client-auth-service'
 
 const prisma = new PrismaClient();
 
+interface SeedResult {
+  step: string;
+  created: boolean;
+  message: string;
+}
+
 /**
  * Role master data (docs/Arogya_Sakhi_Database_Design_ERD_Table_Definitions.docx.md,
  * Appendix A.1 "roles", line 337). This is real reference data required in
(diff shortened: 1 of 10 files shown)


apps/auth-service/prisma/seed.ts
```diff
@@ -43,16 +43,68 @@ const ROLES: { roleCode: string; roleName: string; description: string }[] = [
...
 
+/**
+ * Bootstraps the initial ADMIN user from environment variables so no admin
+ * credential is ever hardcoded in the repo. Runs in every environment
+ * (including production) when both ADMIN_MOBILE_NUMBER and ADMIN_PASSWORD are
+ * set; if either is missing it logs a notice and skips, so a fresh env is
+ * never blocked from seeding. Only creates the admin when no user with that
+ * mobile number exists yet — an existing user is left untouched (never
+ * re-created and never has its password rotated by the seed).
+ */
```
palani-ND, line 75: /**
 * TODO: Consider adding a return value (e.g. { created: boolean; message: string })
 * to help operators understand what actually ran, especially in CI/CD pipelines
 * where seed may run multiple times. Currently all functions return void and only
 * log output, which can make it unclear what actually executed.
 */

Dharanishwaran-ND: The three seed functions now return { step, created, message } instead of void, and main() prints a summary showing exactly which steps created data vs. skipped — verified working (0/3 step(s) created new data when everything already exists).

Pemmaiah544:
Tracked by #9.

PR #59, opened 2026-07-21
feat(beneficiary,visit-form,auth,gateway): geography-derived health block, form response naming, and prior review/docs fixes

Dharanishwaran-ND:
## Summary
This PR carries the full history of `feature/swagger-error-docs-and-modularization` (12 commits) into `develop`:

- **Health block enrollment fix**: `pii.healthBlockId` is now optional in `POST /beneficiaries` (the Sakhi app never captured it) and is derived server-side from `pii.phcId`'s parent Health Block via a new `GET /geography-units/:id` in auth-service, proxied through the gateway using the caller's own forwarded token — no new auth mechanism introduced.
- **Form submission response naming**: `POST /forms/{formCode}/submissions` now echoes the submitted data back as `formData` (was `formDataJson`), matching the request field name.
- **CI fix**: `geography.client.ts` now reads `AUTH_SERVICE_BASE_URL` directly instead of via the shared `appConfig` singleton, so importing it doesn't drag required-but-unset env vars into unit tests.
- **PR #58 review fixes**: project/funder active-status filtering, CHILD case DOB cross-check, visit-status enum→lookup migration rewritten with a real backfill script, ts-node devDependency + lockfile sync.
- **Docs/tooling**: error-response envelope alignment with the HLD, documented API response schemas, revised app form markdown reference, geography units seeding, large-file module split, lockfile regenerated under Node 20.11.0 to match CI.

## Test plan
- [x] `npx nx test beneficiary-service` — passing, including new tests for optional `healthBlockId` and the geography client
- [x] `npx nx test visit-form-service` — passing
- [x] `npx nx test auth-service` — passing, including new `geography.service.spec.ts`
- [x] `npx nx test api-gateway` — passing
- [x] CI green on `feature/swagger-error-docs-and-modularization` (lint + test + build)
- [ ] Frontend confirms enrollment succeeds without sending `healthBlockId`
- [ ] Frontend updates response parsing from `data.formDataJson` to `data.formData`

Related to #78


palani-ND (commented):
Automated review pass focused on security/scalability/performance per this repo's own `.claude/CLAUDE.md` standards, cross-checked against the SRS/ERD. Left inline notes on the items worth a look; everything else (migration+backfill safety, the `formDataJson`→`formData` rename, the `AUTH_SERVICE_BASE_URL` fix, project/funder filtering, geography-endpoint auth) checked out fine.

apps/auth-service/src/geography/geography.repository.ts
```diff
@@ -0,0 +1,10 @@
+import type { PrismaService } from '../prisma/prisma.service';
+
+/** Data access for geography_units master data (State/District/Block/PHC/Sub-centre/Village/Pada). */
+export class GeographyRepository {
+  constructor(private readonly prisma: PrismaService) {}
+
+  findById(id: string) {
+    return this.prisma.geographyUnit.findUnique({ where: { geographyUnitId: id } });
```
palani-ND: `findById` doesn't filter `isDeleted`/`status`, unlike every other master-data repo touched in this PR (e.g. `project.repository.ts` filters `isDeleted: false` consistently). Root `CLAUDE.md` §11 says "soft-delete where needed," and the response schema in `geography.controller.ts` even returns `status: ACTIVE|INACTIVE` — but the caller (`geography.client.ts`) never checks it. Right now a soft-deleted or INACTIVE PHC/Health Block can still be silently accepted when deriving `healthBlockId` for a new beneficiary. Worth adding a filter here (or an explicit active check in the caller).

apps/beneficiary-service/src/geography/geography.client.ts
```diff
@@ -0,0 +1,48 @@
...
+  authorizationHeader: string,
+): Promise<string> {
+  const url = `${AUTH_SERVICE_BASE_URL}/api/v1/geography-units/${phcId}`;
+  const res = await fetch(url, { headers: { Authorization: authorizationHeader } });
+
+  if (res.status === 404) {
+    throw unprocessable('pii.phcId does not refer to a known geography unit.');
+  }
+  if (!res.ok) {
+    throw notFound('Unable to resolve pii.phcId — geography lookup failed.');
```
palani-ND: Any non-404 failure from auth-service (5xx, timeout, network error) is thrown as `notFound(...)`, which becomes a 404 to the enrollment caller. That conflates "bad phcId" (a real data problem, already handled by the 404 branch above) with "our dependency is down" (an infra problem) — a Sakhi retrying enrollment during an auth-service blip sees a confusing "not found" instead of a retryable/server error, and it'll pollute 404-rate monitoring during outages. `geography.client.spec.ts:46` currently asserts this exact behavior, so it reads as intentional — flagging in case it wasn't. Consider a 502/503 (`serviceUnavailable` or similar) for the non-ok/non-404 branch instead.

apps/beneficiary-service/src/geography/geography.client.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  const body = (await res.json()) as { data: GeographyUnit };
+  const phc = body.data;
+
+  if (phc.geoType !== 'PHC') {
```
palani-ND, line 70: Nit: only `geoType === 'PHC'` is checked before returning `phc.parentId` as the Health Block — nothing verifies the parent row is actually `BLOCK`-typed. Low risk given seed/master data is curated, but a one-level data-entry error in `geography_units.parent_id` would be silently accepted rather than surfaced as a 422. Might be worth an explicit parent-type check if this is cheap to add (would need the parent's geoType, e.g. via a second lookup or by including it in the `GET /geography-units/:id` response).

apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
```diff
@@ -268,6 +122,11 @@ export class BeneficiaryService {
     const journeyStartDate = dto.case.registrationDate;
     const currentPhase = dto.case.caseType === 'MOTHER' ? 'ANC' : 'NN';
 
+    // Mobile never sends pii.healthBlockId (no field for it on the enrollment
+    // form) — derive it server-side from pii.phcId's parent Health Block
+    // instead of persisting null for every case.
+    const healthBlockId = await resolveHealthBlockIdFromPhc(dto.pii.phcId, authorizationHeader);
```
palani-ND, line 129: Scalability note (root `CLAUDE.md` §15: "Cache hot, rarely-changing data (rule packs, master data) in Redis with TTL + event-driven invalidation"): `geography_units` is exactly that kind of master data, yet this now does a synchronous gateway→auth-service round trip on every single `POST /beneficiaries`. Not a blocker for this PR, but worth a follow-up ticket to cache PHC→Health Block resolution rather than hitting auth-service per enrollment as enrollment volume grows.

apps/visit-form-service/src/visits/visitInstance.controller.ts
```diff
@@ -34,26 +35,26 @@ const createVisitInstanceRequestSchema = createVisitInstanceSchema.extend({
...
-  success: z.literal(false),
-  message: z.string(),
-  errorCode: z.string().openapi({ example: 'VALIDATION_ERROR' }),
-  details: z.record(z.unknown()).optional(),
+  actualVisitDate: z
+    .string()
+    .datetime()
+    .nullable()
+    .openapi({ example: '2026-07-20T00:00:00.000Z' }),
+  statusLookupValueId: z
```
palani-ND, line 53: This documents `statusLookupValueId` as a non-nullable `z.string().uuid()`, but `schema.prisma` makes the underlying column nullable for the transition window until `backfill-visit-status-lookup.ts` has run in every environment. Until backfill completes everywhere, `GET /visits` can genuinely return `statusLookupValueId: null` for legacy rows — a real (if temporary) break from what this OpenAPI schema promises to any client (mobile/frontend) that trusts it as non-null. Might be worth `.nullable()` here until backfill is confirmed complete in all envs, then tightening back up in a follow-up PR.

apps/beneficiary-service/src/beneficiary/dto/create-beneficiary.dto.ts
```diff
@@ -200,6 +206,21 @@ export const createBeneficiarySchema = z
         });
       }
 
+      // pii.dateOfBirth and childDetails.dateOfBirth are both required and
+      // both describe the child's DOB for a CHILD case. Reject a mismatch so
+      // the two tables can't persist conflicting dates — downstream code
+      // (e.g. buildSearchTokens) treats them as the same value.
+      if (
+        data.childDetails &&
+        data.pii.dateOfBirth.getTime() !== data.childDetails.dateOfBirth.getTime()
```
palani-ND: Nit: this cross-check compares `.getTime()` (exact timestamp equality), while `beneficiary.duplicate-detection.ts` elsewhere in this same PR compares dates via `.toISOString().slice(0, 10)` specifically to get calendar-day equality and avoid time-of-day/timezone mismatches. Two submissions naming the same calendar date with different time-of-day components would spuriously fail this check. Probably low real-world likelihood (both fields likely come from the same date-only picker), but inconsistent with the convention used a few functions over — might be worth aligning for consistency.

apps/auth-service/src/geography/geography.repository.ts
Dharanishwaran-ND: Good catch — valid. Fixed in 5d43384: `findById` (and `findAncestors`) now use `findFirst` with `isDeleted: false`, matching `project.repository.ts`. A soft-deleted PHC/Block now returns null → surfaces as a 404/422 rather than being silently resolved. I also added an explicit `status !== 'ACTIVE'` guard in the beneficiary-service caller (`geography.client.ts`) so INACTIVE units are rejected with a 422 when deriving `healthBlockId`.

apps/beneficiary-service/src/geography/geography.client.ts
Dharanishwaran-ND: Valid — it was not intentional, thanks for flagging. Fixed in 5d43384: non-404 responses (5xx, etc.) and network errors/timeouts from auth-service now throw a new `badGateway` (502) helper instead of `notFound`, so "dependency down" is distinct from "bad phcId" (still 422). Updated the spec (`geography.client.spec.ts`) to assert 502 for the 5xx and network-error cases.

apps/beneficiary-service/src/geography/geography.client.ts
Dharanishwaran-ND: Valid — implemented. `resolveHealthBlockIdFromPhc` now does a second lookup on `parentId` and asserts the parent is `geoType === 'BLOCK'` (and ACTIVE) before returning it as the Health Block; a one-level `parent_id` data-entry error now surfaces as a 422 instead of persisting a wrong `healthBlockId`. Added spec coverage for the non-BLOCK-parent and inactive-parent cases.

apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
Dharanishwaran-ND: Agreed, and noting it's explicitly a non-blocking follow-up as you framed it. Leaving the synchronous resolution as-is for this PR; the PHC→Health Block resolution is a good candidate for the Redis master-data cache with TTL + event-driven invalidation (§15). I'll raise a follow-up ticket to cache it rather than hitting auth-service per enrollment.

apps/visit-form-service/src/visits/visitInstance.controller.ts
Dharanishwaran-ND: Valid — fixed in 5d43384. `statusLookupValueId` in the response schema is now `.nullable()` for the backfill transition window (the Prisma column is `String?`), with a comment to tighten it back to non-null in a follow-up once `backfill-visit-status-lookup.ts` is confirmed run in every env.

apps/beneficiary-service/src/beneficiary/dto/create-beneficiary.dto.ts
Dharanishwaran-ND: Valid consistency fix — done. The child-DOB cross-check now compares `.toISOString().slice(0, 10)` (calendar day) on both sides, matching the convention in `beneficiary.duplicate-detection.ts`, so a same-date pair with differing time-of-day components no longer spuriously fails.

PR #60, opened 2026-07-22
feat(geography): add root-units and unit-children listing endpoints

Dharanishwaran-ND:
## Summary
- Adds `GET /geography-units/roots` — lists all top-level geography units (STATEs, no parent)
- Adds `GET /geography-units/:id/children` — lists the direct children of a unit (e.g. all districts under a state)
- Supports cascading State→District→Block dropdowns per SRS line 971 (previously only bottom-up ancestor resolution existed)
- No schema/migration changes — builds entirely on the existing `geography_units` table
- Documents the resolution of the ERD's two unreconciled geography models (`geography_units` vs. the unimplemented per-level Model B tables) — `geography_units` confirmed as the single source of truth

## Design & plan
- `docs/superpowers/specs/2026-07-22-geography-children-endpoint-design.md`
- `docs/superpowers/plans/2026-07-22-geography-children-endpoint.md`

## Test plan
- [x] `npx nx lint auth-service` — 0 errors, 0 warnings
- [x] `npx nx test auth-service` — 83/83 passing (13 new: 4 repository + 9 service), 0 regressions
- [x] `npx nx affected -t lint test` — only pre-existing, unrelated "No tests found" failures in `wrapper-api-service`/`cms-content-service`/`reporting-etl-service` (confirmed unrelated to this change)
- [ ] Live curl smoke test — deferred; the local dev auth-service instance needs a restart to pick up the change (it's serving live traffic via an active tunnel) and was intentionally not restarted mid-session

Related to #77


apps/visit-form-service/src/forms/beneficiary-duplicated-fields.ts
```diff
@@ -0,0 +1,53 @@
...
+ * stores every field verbatim — this list does not affect validation, the
+ * app's request payload, or the raw submission record (ERD line 19: the
+ * complete submitted payload must be kept, even for fields also promoted
+ * elsewhere).
+ *
+ * Source of the mapping: see the beneficiary-service create-beneficiary DTO
+ * (piiSchema / motherDetailsSchema) — each code below corresponds 1:1 to a
+ * field accepted there. Keep the two lists in sync if either form changes.
+ */
+export const BENEFICIARY_DUPLICATED_FIELD_CODES: ReadonlySet<string> = new Set([
```
palani-ND, line 26: This list is a manually-maintained mirror of beneficiary-service's `create-beneficiary.dto.ts` fields, linked only by a code comment ("Source of the mapping: see the beneficiary-service create-beneficiary DTO") — there's no compile-time or test-time check that keeps the two in sync. If beneficiary-service renames/removes a field, or this form's `question_code`s change, the two can silently drift: a genuinely-duplicated field stops being excluded (persists a second, independently-editable copy), or a non-duplicated field gets wrongly dropped from `form_answers`. Given the forklift rule (services shouldn't know each other's internal shapes), consider at minimum a shared contract test in visit-form-service that fails loudly if a listed code no longer maps to anything, so drift is caught rather than silent.

apps/visit-form-service/src/forms/form.mapper.ts
```diff
@@ -58,6 +60,150 @@ export interface FormSubmissionRow {
...
+/** An empty row with all typed columns null — one gets filled per field. */
+function emptyRow(fieldCode: string): FormAnswerRow {
+  return {
+    fieldCode,
+    answerValueText: null,
+    answerValueNumber: null,
+    answerValueDate: null,
+    answerValueBool: null,
+    answerValueJson: null,
+    isIndexed: false,
```
palani-ND, line 100: `isIndexed` is unconditionally `false` for every row `buildFormAnswers` produces — confirmed no other code path in this service ever sets it `true`, and nothing reads it yet either. The JSDoc above this function frames the whole point of `form_answers` as making fields "individually queryable" for "searchable/reportable/rule-critical" use (per the ERD's design stance), but as written every row is indistinguishable via this flag. If `isIndexed` is meant to gate which fields get indexed/searched later, is that intentionally deferred to a follow-up, or should this decomposer already know (e.g. from `form-field.dto.ts`'s schema) which fields are rule-critical vs. just along for the ride?

apps/auth-service/src/geography/geography.controller.ts
```diff
@@ -90,5 +110,28 @@ export function createGeographyRouter(service: GeographyService, signer: TokenSi
     }),
   );
 
+  doc.get(
+    '/geography-units/:id/children',
```
palani-ND, line 114: Root CLAUDE.md §15 says "Use pagination (cursor-based) on list endpoints," and neither `/geography-units/roots` nor `/:id/children` paginate. Probably fine today (STATE/DISTRICT/BLOCK counts are small), but a district's children (SUBCENTREs) or a sub-centre's children (VILLAGEs) could plausibly run into the hundreds in real seeded data — worth deciding now whether that's an accepted exception for geography master data (bounded, rarely-changing) or something to revisit before villages/padas are seeded at full scale.

palani-ND (commented):
Reviewed against root/auth-service CLAUDE.md conventions and the PR's own design/plan docs. Ran the full test suites on this commit: auth-service 65/65 non-DB tests pass, visit-form-service 49/49 non-DB tests pass — no regressions, matches the PR's claimed 83/83.

**Scope note (the main thing worth flagging):** the PR title/description and both linked design docs (`docs/superpowers/specs/...-design.md`, `docs/superpowers/plans/...md`) describe this as an **auth-service-only** change ("Files touched (auth-service only)"), matching the `geography.repository.ts`/`geography.service.ts`/`geography.controller.ts` + spec changes. But the diff *also* includes a second, unrelated, completely undocumented feature in `visit-form-service`: `form_answers` decomposition (`form.mapper.ts`'s new `buildFormAnswers`), a new `beneficiary-duplicated-fields.ts` cross-service field-exclusion list, and a transactional rewrite of `form.repository.ts#createSubmission`. None of this is mentioned in the PR body, the design doc, or the plan doc. It's good, tested work on its own, but bundling an undocumented feature into a PR whose own paper trail says it doesn't touch that service makes this hard to review against its stated intent, and violates root CLAUDE.md §13's "small, focused PRs." Worth splitting into its own PR with its own description, even post-hoc.

A few smaller items inline.

apps/visit-form-service/src/forms/beneficiary-duplicated-fields.ts
Dharanishwaran-ND: Good catch — and it turned out worse than described: the hardcoded codes didn't just lack a sync check, several had already drifted from the real seeded MOTHER_REGISTRATION form (`edd` vs. seeded `edd_date`, `gravida_total_number_of_pregnancies` vs. seeded `gravida`, `para_number_of_births_after_24_weeks` vs. seeded `para`, `still_births` vs. seeded `stillbirths`, `enter_the_beneficiary_address` vs. seeded `beneficiary_address`). As merged this exclusion was largely a no-op against the actual seed data. Also removed a handful of entries (name, DOB, geography, height/weight) that don't correspond to any field the form currently asks — kept the list to only what's real today.

Fixed in ff5e52b:
- Corrected `BENEFICIARY_DUPLICATED_FIELD_CODES` to match the real seeded question_codes.
- Extracted the seed form field data into a side-effect-free `registration-forms.seed-data.ts` module (no Prisma import, safe to import from a test) shared by `prisma/seed.ts` and a new test.
- Added a contract test in `form.mapper.spec.ts` ("every code in BENEFICIARY_DUPLICATED_FIELD_CODES matches a real seeded question_code") that fails loudly if a listed code stops matching a seeded field — catches the exact drift mode you flagged.

This doesn't reach across to beneficiary-service's DTO fields directly (still no compile-time cross-service link — that's the forklift rule, so I don't think we want one), but it does catch the more common/likely drift: this form's own question_codes changing without the exclusion list being updated.

apps/auth-service/src/geography/geography.controller.ts
Dharanishwaran-ND: Confirmed: no controller anywhere in this repo currently implements cursor-based pagination on a list endpoint (checked audit-service, projects, lookups, beneficiary, approvals, incentives, closures, referrals, media, notifications, rules, visit-instances, sync-batches) — every `z.array(...)`-returning endpoint in the codebase today returns its full result set unpaginated. So this isn't a gap this PR introduces; it's following the repo's existing (unpaginated) precedent exactly.

Leaving as-is for this PR rather than adding pagination to just these 2 routes, which would be inconsistent with every other list endpoint. Agreed these are bounded, rarely-changing master-data lists today (STATE/DISTRICT/BLOCK counts are small), but if VILLAGE/PADA-level children get seeded at real scale, this is worth a dedicated follow-up that establishes the pagination pattern repo-wide rather than a one-off on this endpoint.

apps/visit-form-service/src/forms/form.mapper.ts
Dharanishwaran-ND: Not fixing this one — it's intentionally deferred, not an oversight. Checked both source docs this PR is scoped against (SRS v3.0 and the ERD) plus `form-field.dto.ts`'s `formFieldSchema`: none of them name which fields are "rule-critical"/searchable vs. "just along for the ride" — `formFieldSchema` has no such flag today (no `indexed`/`searchable`/`ruleCritical` key), and `form_answers.is_indexed` itself defaults to `false` at the Prisma schema level with no doc specifying a gating rule.

Since there's no source-of-truth for which fields should be indexed, hardcoding a guess here would be inventing a rule rather than implementing one, and this PR doesn't have a mandate to design that (reporting/rules requirements for `form_answers` aren't otherwise in scope for the children/roots-endpoint work this PR title describes — this file's `buildFormAnswers` addition is adjacent, from the same PR). Left `isIndexed: false` unconditionally, matching the schema default, and left the question open in the code's own JSDoc rather than silently deciding it. Whoever owns the reporting/rules-engine requirements should specify which fields need it as a follow-up — happy to wire it up once there's a concrete list to work from.

Issue #28, opened 2026-07-16
Create form_definitions & form_versions


Dharanishwaran-ND:
| Field | Value |
|---|---|
| Category | Backend |
| Dependency | Prisma |
| Estimate (days) | 4 |
| Priority | High |
| Deliverable | Migration |
| Success Criteria | Versioning ready |

Issue #26, opened 2026-07-16
Create geography hierarchy


Dharanishwaran-ND:
| Field | Value |
|---|---|
| Category | Backend |
| Dependency | Prisma |
| Estimate (days) | 6 |
| Priority | High |
| Deliverable | Migration |
| Success Criteria | Hierarchy ready |

PR #61, opened 2026-07-22
docs+schema: ERD audit fixes, schema reconciliation, and migrate tooling

Dharanishwaran-ND:
## Summary
This branch grew to cover three related efforts (branch name reflects only the first, smallest one). Grouped here as one PR per request:

### 1. ERD document audit & fixes (documentation only — 10 commits, `docs(erd)`/`docs(beneficiary)`)
Fixed 55 of 57 issues found in a full audit of `docs/Arogya_Sakhi_Database_Design_ERD_Table_Definitions.docx.md`:
- **Naming**: table headings + FK targets standardized to plural snake_case
- **Dangling references**: `visits`→`visit_instances`, `beneficiaries`→`beneficiary_cases`, `panchayats`→`geography_units`
- **Column contradictions**: BIGINT→UUID FKs, missing data types, duplicate columns removed, malformed constraints, `sex` enum unified to `FEMALE/MALE/OTHER/INTERSEX`
- **Missing audit columns** added to 11 tables; broken index/constraint references fixed
- **Duplicate-column pairs** resolved in `project_registration_targets`, `approval_requests`, `reopen_requests`, `call_logs`, `media_assets`
- **Structural/heading** formatting fixes
- **Risk-summary model overlap** (#28) resolved
- Deferred (2 of 57): #26/#27 — the `geography_units` vs normalized-tables model conflict, folded into the beneficiary_pii design below.

### 2. Schema reconciliation — code made to match the fixed doc (3 commits, `feat`)
Three services had schemas that disagreed with the doc's resolved design:
- `approval-service`: `status` enum → `decision_status_lookup_id` (lookup FK); dropped `ApprovalRequestStatus` enum; updated DTO/controller/spec. No migration (never deployed).
- `closure-reopen-service`: dropped 3 redundant `supervisor_*` columns from `reopen_requests` (+ new migration).
- `risk-referral-service`: dropped 4 redundant baseline/latest/ever-at-risk columns from `risk_state_snapshots`, keeping only `ccv_state` (+ new migration).

### 3. Migration tooling + seed (1 commit, `feat(tooling)`)
- `npm run migrate` (+ `db:migrate` alias) → `prisma migrate deploy` across all 12 services via the existing `prisma-foreach.js`.
- Seeded `APPROVAL_STATUS` lookup values in auth-service (backs the new `decision_status_lookup_id`).
- README documents the append-only migration convention.

### Also included
- `docs/superpowers/specs/2026-07-22-beneficiary-geography-fk-refactor-design.md` — design spec for consolidating `beneficiary_pii`'s geography columns to `geography_units` (Model A source-of-truth). **Design only — not yet implemented.**

## Test plan
- [x] Lint + tests green for all touched services (approval, closure-reopen, risk-referral, auth)
- [x] Prisma clients regenerate cleanly; `migrate status` reaches the DB and enumerates the two new drop-column migrations correctly
- [ ] Full `npm run migrate` apply against a fresh DB — not run to completion here (remote DB, long runtime); verified via `migrate status`

## Notes for reviewer
- Migrations are **destructive** (DROP COLUMN) on `reopen_requests`/`risk_state_snapshots` — confirmed no app code references the dropped columns and no seed data exists, but worth a second look if any environment already holds data there.
- `#26/#27` geography model conflict remains open, pending the beneficiary_pii refactor implementation.

Related to #82


palani-ND (commented):
Verified the highest-risk item first: the two DROP COLUMN migrations (`reopen_requests`, `risk_state_snapshots`). Grepped the whole repo for every dropped column's camelCase Prisma field name and found zero references anywhere in application code, confirming both `ReopenRequest`/`RiskStateSnapshot` genuinely have no service/controller/DTO layer yet — the migrations are safe as described. Also ran tests+lint for every touched service: approval-service 4/4, closure-reopen-service 4/4, risk-referral-service 4/4, rules-service 9/9, all lint-clean — matches the PR's claims.

**Scope note:** the PR description covers three efforts (ERD audit, schema reconciliation, migration tooling) and is upfront that it's an intentional bundle. But there's a fourth, undisclosed one: the last commit (`feat(rules): add rule-pack version fetch and publish endpoints`) adds a full new admin feature — `ruleVersion.controller/service/repository/module` + DTO + specs — not mentioned anywhere in the PR body. Unlike the other three sections it gets zero description, so a reviewer skimming the summary would miss that a new ADMIN-gated write endpoint (`POST /admin/rules/:setId/publish`) shipped in this PR. Worth a line in the description at least, even post-hoc.

One concrete correctness finding inline on the new rules feature; everything else looks solid.

apps/rules-service/src/rules/ruleVersion.service.ts
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Publishes a new rule pack version for a set (create + publish in one step):
+   * auto-increments versionNo, computes the checksum, marks it PUBLISHED as of
+   * now, and retires the previously-published version. 404 if the set is unknown.
+   */
+  async publish(ruleSetId: string, dto: PublishRuleVersionInput, publishedByUserId: string | null) {
+    const set = await this.repository.findSetById(ruleSetId);
+    if (!set) throw notFound('Rule set not found.');
+
+    const existingCount = await this.repository.countVersions(ruleSetId);
```
palani-ND: `countVersions` is read here, outside the `publishNewVersion` transaction, then used to compute `versionNo` (`v${existingCount + 1}`) at line 55. Two concurrent `publish()` calls for the same rule set (double-click, or two admins publishing near-simultaneously) can both read the same count and both attempt to insert the same `versionNo` — which `@@unique([ruleSetId, versionNo])` on `RuleVersion` (schema.prisma) will reject as a Prisma P2002. Nothing here or in the controller catches that, so the loser gets an uncaught DB-constraint error surfaced as a generic 500 instead of a clean 409/retry. Given this is ADMIN-only and probably low-traffic, likely a minor real-world risk, but since `publishNewVersion` already wraps its two writes in `$transaction`, moving the `countVersions` read inside that same transaction (or deriving the next version number from a DB-side sequence/`MAX(version_no)` within it) would close the race entirely.

Dharanishwaran-ND: Confirmed valid — verified `@@unique([ruleSetId, versionNo])` on `RuleVersion` in `apps/rules-service/prisma/schema.prisma` and that `countVersions` was indeed read before the `$transaction` in `publishNewVersion`, so two concurrent `publish()` calls could read the same count and race to insert the same `versionNo`, which would throw an uncaught Prisma P2002 (500) since nothing caught it.

Fixed in 33da0e0:
- Removed `countVersions` from `RuleVersionRepository` and moved the count inside `publishNewVersion`'s transaction, now run at `SERIALIZABLE` isolation (`{ isolationLevel: 'Serializable' }`). If two `publish()` calls for the same rule set overlap, Postgres aborts one with a serialization failure (P2034) instead of letting both silently compute the same `v${existingCount + 1}`.
- `RuleVersionService.publish()` now catches both P2002 and P2034 and returns a clean `409` via the existing `conflict()` helper (matches the `isUniqueConstraintViolation` pattern already used in `auth-service/src/lookups/lookup.service.ts`), instead of leaking a raw DB error as a 500.
- Updated `ruleVersion.service.spec.ts`: removed the `countVersions` mock, and added cases asserting 409 on P2002, 409 on P2034, and that unrelated errors are rethrown unchanged.
- `npx nx lint rules-service` and `npx nx test rules-service` both pass (12/12 tests), plus `npx nx affected -t lint test --base=origin/develop` is green across all 6 affected projects.

Didn't add caller-side retry logic since this is ADMIN-only/low-traffic as you noted — the 409 gives the client a clean signal to retry manually rather than silently succeeding twice.

PR #62, opened 2026-07-23
feat(geography): filterable GET /geography-units master API

Dharanishwaran-ND:
## Summary
First of the missing **master-data APIs** described in the docs but not yet built. Adds `GET /geography-units` — the geography list/tree read supporting cascading dropdowns (SRS line 971; HLD `GeographyTree` admin component).

`GET /geography-units?geoType=DISTRICT&parentId=<stateId>`
- Both query params optional.
- **No filter → top-level units (STATEs) only** — an unfiltered call never dumps the whole multi-level tree; clients walk down one level per request (the cascade).
- `geoType` and/or `parentId` narrow the result.
- Excludes soft-deleted; ordered by `geoCode`; capped at 500 rows.
- Read-only, any authenticated role (same policy as the existing `/:id` and `/:id/ancestors` reads).
- Empty result → `200 []`, not 404 (it's a query, not a fetch-by-id).

## Test plan
- [x] New `geography.repository.spec.ts` — filter combinations + no-filter-defaults-to-roots
- [x] `geography.service.spec.ts` — list projection + empty-result cases
- [x] auth-service: 80/80 tests pass, lint clean

## Notes for reviewer
- **Overlap with PR #60** (`feature/geography-children-endpoint`, not yet merged): that PR added `GET /geography-units/roots` and `/:id/children` as path-based variants. This PR does the same top-down navigation via query params on a single `GET /geography-units`. They don't conflict at the code level (different routes) and can coexist, but if both merge you may want to consolidate on one style — flagging for a decision.
- This is PR 1 of 3 for the missing master APIs. Next: rule-pack versioning (`GET /admin/rules/:setId` + publish), then `GET /sync/deltas` aggregator.

Related to #80


apps/auth-service/src/geography/geography.controller.ts
```diff
@@ -47,6 +61,31 @@ function envelope<T extends z.ZodTypeAny>(data: T) {
 export function createGeographyRouter(service: GeographyService, signer: TokenSigner) {
   const doc = createDocumentedRouter();
 
+  // Registered before `/geography-units/:id` so the bare `/geography-units`
+  // list path isn't captured by the `:id` route.
```
palani-ND: Minor: the comment says this route is "registered before `/geography-units/:id` so the bare `/geography-units` list path isn't captured by the `:id` route" — but `/geography-units` (2 segments) and `/geography-units/:id` (3 segments) don't actually collide regardless of registration order, since Express matches by path depth, not just prefix. Unlike PR #60's `/roots` vs `/:id` (which genuinely do collide at the same depth and need the ordering), this one doesn't need it. Not a bug, just a slightly over-cautious/inaccurate comment worth tightening.

palani-ND (commented):
Small, clean PR. Ran the suite: 62/62 non-DB auth-service tests pass (matches the claimed 80/80 including the 18 pre-existing DB-integration skips), lint clean. Logic is correct: no-filter defaults to roots only (never dumps the tree), empty result is 200/[] not 404 (right semantics for a query endpoint), soft-delete filtering and the 500-row cap are both present and tested for the filter-combination matrix.

**On the PR #60 overlap you flagged for a decision:** I reviewed #60 as well. My take — this PR's single filterable `GET /geography-units?geoType=&parentId=` is strictly more general than #60's `GET /geography-units/roots` + `GET /geography-units/:id/children`: `parentId` alone covers "children of X" and no-filter covers "roots", but this endpoint *also* covers "all units of a given type across the tree" (`geoType` alone, e.g. all VILLAGEs regardless of parent), which #60 has no way to express at all. If both merge, #60's two routes become a redundant subset of this one. I'd recommend merging this PR's version and dropping #60's two routes rather than trying to keep both styles alive — happy to leave that PR a comment pointing here if useful.

One very minor nit inline, not worth blocking on.

apps/auth-service/src/geography/geography.controller.ts
Dharanishwaran-ND: Confirmed — you're right, and it's not just an inaccuracy in the comment, it was a real logic gap in the reasoning. Express matches routes by path segment count: `/geography-units` is 2 segments, `/geography-units/:id` is 3, so a request to `/geography-units` can never be captured by the `:id` route pattern (which requires a second segment to exist) regardless of registration order. That's genuinely different from PR #60's `/geography-units/roots` vs `/geography-units/:id`, which are both 3-segment paths and do collide if `:id` is registered first (it would swallow `roots` as `id=roots`).

Fixed in e4f1428: reworded the comment to state the real reason for the ordering (readability — list route grouped above single-item reads) and to explain explicitly why no collision hazard exists here, contrasting with PR #60's genuine case. Pushed to `feature/geography-list-endpoint`.

PR #63, opened 2026-07-23
feat(db): add ERD access/project tables, new supervisor-operations-service, and public-schema consolidation

Dharanishwaran-ND:
## Summary
- Adds the ERD's "Access, Project, and Geography" tables to auth-service (supervisor_profiles, device_registry, project_geographies, project_registration_targets, sakhi_assignments, sakhi_location_assignments)
- Scaffolds a new `supervisor-operations-service` owning supervisor_events, event_attendance, inventory_items, inventory_transactions, and call_logs, wired into the gateway
- Gives reporting-etl-service its first Prisma schema (etl_runs, report_exports)
- Adds missing init migrations for 5 services that had Prisma models but no migrations folder (approval-service, rules-service, incentive-wages-service, notification-escalation-service, media-service)
- Consolidates every service onto a single shared `public` Postgres schema — per-service schema targeting was silently misrouting migrations through Supabase's transaction pooler, which doesn't honor per-session `search_path`
- Adds `npm run seed` to run every service's `prisma/seed.ts` in one command

## Test plan
- [x] `npm run lint` and `npm run affected:test` pass across all touched services
- [x] `npm run db:generate` regenerates all Prisma clients cleanly
- [x] `npm run migrate` applies all 14 services' init migrations cleanly against a fresh `public` schema (verified live against Supabase)
- [x] Verified 62/62 expected tables exist post-migration, matching every service's `schema.prisma` exactly
- [x] Verified login + role-gated endpoint (`GET /forms/MOTHER_REGISTRATION/active-version`) work end-to-end after reseeding auth-service + visit-form-service master data
- [ ] Reviewer: confirm no other long-running dev service instances need restarting after this schema change lands in a shared environment

Related to #76


tools/prisma-foreach.js
```diff
@@ -3,12 +3,17 @@
...
+ * schema targeting was dropped after Supabase's pgbouncer transaction pooler
+ * proved unable to honor per-session `search_path`, which silently misrouted
+ * migrations to the wrong schema.
  *
- * For `db push`/`db pull`/`migrate`, this script derives a per-service
- * DATABASE_URL/DIRECT_URL by taking the base connection from the root .env and
- * appending each service's `schema=` (read from apps/<svc>/.env or .env.example).
+ * For `db push`/`db pull`/`migrate`, this script still supports an optional
+ * per-service `schema=` override (read from apps/<svc>/.env or .env.example)
+ * for anyone who reintroduces one — if none is set, it just uses the root
```
palani-ND, line 15: The `--accept-data-loss` auto-append for `db push` (further down in this file, not touched by this diff) predates this PR, but it becomes materially riskier now that every service shares one physical database — a `db push` for one service's schema.prisma run through this script carries platform-wide blast radius, with Prisma's own safety prompt silently forced off and no warning logged. Worth a follow-up: log a loud warning before appending the flag, or require an explicit `--yes`/`--force` on the wrapper script itself.

apps/auth-service/prisma/schema.prisma
```diff
@@ -289,42 +265,217 @@ model GeographyUnit {
...
+  PRIMARY
+  SECONDARY
+  TEMPORARY_COVERAGE
+  PAUSED
+  STOPPED
+
+}
+
+// sakhi_assignments — ERD Appendix A (§4.1). A Sakhi's project + geography scope.
+model SakhiAssignment {
```
palani-ND, line 442: No unique/partial-unique constraint prevents two concurrent `assignmentType: PRIMARY, activeTo: null` rows for the same `sakhiId` — nothing stops a Sakhi from ending up with two simultaneously-"current" primary assignments. The ERD's own §6 ("Required Indexes and Constraints") calls for an index on `sakhi_assignments(sakhi_id, project_id, geography_unit_id, active_to)`, which is absent from both this schema and the migration. Worth adding the index now (cheap) and considering a partial unique index (`WHERE assignment_type = 'PRIMARY' AND active_to IS NULL`) before any service starts writing to this table.

apps/supervisor-operations-service/src/operations/operations.service.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/** Supervisor operations domain logic. Data access is delegated to the repository. */
+export class OperationsService {
+  constructor(private readonly repository: OperationsRepository) {}
+
+  listEvents() {
+    return this.repository.findEvents();
+  }
+
+  createEvent(dto: CreateSupervisorEventInput) {
```
palani-ND, line 13: `createEvent` is a bare passthrough to the repository, but the DTO's own doc comment (`create-supervisorEvent.dto.ts:36`) says `photoMediaId` is "application-mandatory when `status = COMPLETED`... enforced in the service layer." Nothing here enforces that — a caller can create a `COMPLETED` event with no photo, contradicting the documented rule. Either add the check here or update the comment if this was deliberately deferred.

apps/supervisor-operations-service/src/operations/operations.repository.ts
```diff
@@ -0,0 +1,47 @@
...
+ * call_logs) — no cross-service joins (forklift rule).
+ */
+export class OperationsRepository {
+  constructor(private readonly prisma: PrismaService) {}
+
+  findEvents() {
+    return this.prisma.supervisorEvent.findMany({
+      where: { isDeleted: false },
+      orderBy: { eventDate: 'desc' },
+      take: 50,
```
palani-ND, line 16: All four list methods (`findEvents`/inventory items/transactions/call logs) use a hard-coded `take` cap with no cursor/offset exposed through the controller — once a project accumulates more rows than the cap, older rows become permanently unreachable via this API. Root CLAUDE.md §15 calls for cursor-based pagination on list endpoints; worth adding before this service has real usage, since retrofitting pagination onto a shipped API is more disruptive than building it in now.

palani-ND (commented):
Large PR (134 files) — focused review on the two highest-risk/highest-novelty pieces: the platform-wide Postgres schema consolidation, and the brand-new services/tables.

**Schema consolidation (all 12+ services → one shared `public` schema):** verified there are zero table-name collisions across every service's `@@map(...)` — confirmed by diffing all Prisma schemas. `tools/prisma-foreach.js` runs migrations sequentially (not in parallel), so no race on the now-shared `_prisma_migrations` history table either. The rationale (Supabase's transaction pooler not honoring per-session `search_path`) is technically sound and this is a reasonable fix. One real issue found in that same script, flagged inline.

**Test claim discrepancy:** the PR says "`npm run lint` and `npm run affected:test` pass across all touched services," but `reporting-etl-service` (new in this PR) has zero `*.spec.ts` files anywhere, so `npx nx test reporting-etl-service` fails outright ("No tests found, exiting with code 1") — this would fail `nx affected -t lint test` in CI as currently configured. Worth double-checking whether `affected:test` for this project was actually exercised, or whether it was silently skipped some other way.

Ran the rest myself: `auth-service` 71/71 (18 pre-existing DB-integration skips), `supervisor-operations-service` 7/7, all lint-clean. The 6 new auth-service tables (`supervisor_profiles`, `device_registry`, `project_geographies`, `project_registration_targets`, `sakhi_assignments`, `sakhi_location_assignments`) are schema-only with no API layer yet — consistent with how `geography_units` shipped originally, so that's expected, not a gap. `supervisor-operations-service`'s migration, gateway wiring, and bootstrap all match sibling-service conventions.

A few concrete items inline.

tools/prisma-foreach.js
Dharanishwaran-ND: Confirmed valid. Fixed in f2518e2 (fix(tools): require explicit opt-in before prisma-foreach forces db push).

The wrapper now refuses to run `db push` unless the caller passes an explicit `--yes` or `--force`, and logs a loud warning before appending `--accept-data-loss` when they do:

```
node tools/prisma-foreach.js db push
✗ Refusing to run "db push" without an explicit --yes/--force on this wrapper. Every
  service now shares one physical database, so this command has platform-wide blast
  radius. Re-run with "node tools/prisma-foreach.js db push --yes" once you have
  confirmed the change is safe to apply to every service.
```

While fixing this I found and fixed two related bugs the original code had: (1) the `args` array was being read and mutated inside the per-service loop instead of computed once up front, so the accept-data-loss append logic re-ran (harmlessly, but sloppily) on every iteration; (2) the new `--yes`/`--force` wrapper-only flags need to be stripped before forwarding to the real `prisma` CLI, since Prisma doesn't recognize either flag and would otherwise error out. Verified both `db push` (refuses without `--yes`) and `generate` (unaffected, ran cleanly across all 14 services) manually.

apps/auth-service/prisma/schema.prisma
Dharanishwaran-ND: Confirmed valid — checked the schema and the init migration, and `sakhi_assignments` had no `@@index`/`@@unique` at all, so the ERD §6 requirement was fully missing, not just the partial-unique part.

Fixed in 5d3be89 (fix(db): add missing composite index on sakhi_assignments): added `@@index([sakhiId, projectId, geographyUnitId, activeTo])` to the Prisma schema and the matching `CREATE INDEX` statement to the init migration SQL (this service's tables ship in a single not-yet-released init migration, so there's no live-migration compatibility concern with hand-editing it directly).

I deliberately did **not** add the partial unique index (`WHERE assignment_type = 'PRIMARY' AND active_to IS NULL`) you suggested as a follow-up — that changes actual data-integrity behavior (rejecting a second concurrent PRIMARY assignment outright), and depends on a business rule I can't verify from the code alone: whether an overlap window is ever valid during a Sakhi handover between supervisors/projects. Leaving that as a follow-up decision for whoever owns that workflow rather than guessing.

apps/supervisor-operations-service/src/operations/operations.service.ts
Dharanishwaran-ND: Confirmed valid — read `operations.service.ts` and it was exactly as described: `createEvent` was a bare passthrough with no check against the DTO's own documented rule.

Fixed in a059c9c (fix(supervisor-operations): enforce photo requirement on completed events): `createEvent` now throws the standard 422 `unprocessable()` (the same helper used for equivalent business-rule checks in `beneficiary-service` and `auth-service`'s lookup service) when `status === 'COMPLETED' && !photoMediaId`. Added two new spec cases covering both branches (rejects without hitting the repository / allows through when the photo is present) — all 9 tests in the file pass.

Side note: adding this import surfaced that this service's `jest.config.ts` was missing the `jose` ESM transform config that most other services already have (`auth-service`, `rules-service`, `visit-form-service`, `beneficiary-service`) — without it, importing anything from `@armman/service-commons` fails at module load with a `SyntaxError: Unexpected token 'export'` from the ESM-only `jose` package. Fixed that in the same commit since the new import needed it to actually run.

apps/supervisor-operations-service/src/operations/operations.repository.ts
Dharanishwaran-ND: Checked this against every other service in the repo before deciding: `take: 50` (or `take: 100`/`take: 500` for master-data lists) with no cursor/offset is the **existing repo-wide convention**, not something this PR introduced or deviated from. It's identical in `rules-service`, `risk-referral-service`, `closure-reopen-service`, `approval-service`, `incentive-wages-service`, `media-service`, `notification-escalation-service`, `audit-service`, `sync-service`, and `visit-form-service` — all use the exact same `findMany({ orderBy, take: 50 })` pattern with no cursor exposed through the controller. `operations.repository.ts` is copying that established pattern exactly, not adding a new gap.

You're right that `STANDARD.MD`'s "Architecture, API, Data & Ops" table does list cursor-based pagination as a target standard (I didn't find a numbered "§15" anywhere in root `CLAUDE.md` — the closest match is that STANDARD.MD row). But since literally every existing service has this same gap, retrofitting it only in `supervisor-operations-service` would leave every other list endpoint in the codebase still non-compliant, and would make this one service inconsistent with its siblings for no functional reason.

I'm treating "adopt cursor-based pagination repo-wide" as a cross-cutting follow-up that belongs in its own PR (likely as a shared helper in `service-commons` so all ~14 services pick it up consistently), not a fix scoped to this PR's new service. Leaving this as-is for now — flagging for a maintainer to confirm that's the right scope for the follow-up rather than guessing at prioritization.

Dharanishwaran-ND:
Addressed all 4 inline review comments from the review above (replies posted on each thread) plus the test-claim discrepancy raised in the review summary. Recap:

**Fixed (pushed to `feature/auth-access-tables`, 4 commits):**
1. `tools/prisma-foreach.js` — `db push` now refuses to run unless the caller passes an explicit `--yes`/`--force`, with a loud warning logged before `--accept-data-loss` is appended (f2518e2).
2. `apps/auth-service/prisma/schema.prisma` + init migration — added the missing `sakhi_assignments(sakhi_id, project_id, geography_unit_id, active_to)` composite index called for by ERD §6 (5d3be89).
3. `apps/supervisor-operations-service` — `OperationsService.createEvent` now actually enforces "`photoMediaId` required when `status = COMPLETED`", matching the DTO's own doc comment. Also fixed a missing `jose`-ESM jest transform this surfaced (a059c9c).
4. `apps/reporting-etl-service` — the review's summary noted `nx test reporting-etl-service` failed outright with "No tests found" since this project (new in this PR) had zero spec files. Verified that directly, then added a smoke test for its one route plus the same missing `jose`-ESM jest transform (a9bc5a9).

**Left as-is, with reasoning in the thread replies:**
- The pagination comment on `operations.repository.ts` — verified `take: 50` with no cursor is the existing convention across every other service in the repo (`rules-service`, `audit-service`, `sync-service`, etc. all do the same thing), so this isn't a regression specific to this PR. Proposed treating repo-wide cursor pagination as its own cross-cutting follow-up PR rather than fixing it in one service only.
- The partial-unique-index suggestion on `sakhi_assignments` (preventing two concurrent PRIMARY assignments) — added the plain index, but left the uniqueness constraint as a follow-up decision since it depends on a business rule (whether an overlap window is ever valid during a Sakhi handover) that needs a product owner's call, not a guess.

All affected projects (`auth-service`, `supervisor-operations-service`, `reporting-etl-service`) pass `nx test` and `nx lint` after these changes.

palani-ND:
All four review findings fixed and verified on `f2518e2`:

- **`--accept-data-loss` auto-force** — now requires explicit `--yes`/`--force`, refuses otherwise with a clear error, logs a warning when confirmed, and correctly strips the wrapper-only flags before forwarding to the real `prisma` CLI. Also cleaned up the args-array mutation.
- **`sakhi_assignments` missing index** — composite index `(sakhi_id, project_id, geography_unit_id, active_to)` added to both `schema.prisma` and the migration SQL, matching the ERD's §6 requirement. The partial-unique-constraint question was correctly deferred as a separate business-rule decision (overlap-during-handover), not bundled in as a schema fix — good call.
- **`photoMediaId` enforcement** — `OperationsService.createEvent` now throws `422 unprocessable()` when `status = COMPLETED` with no photo, with both branches covered in `operations.service.spec.ts`.
- **`reporting-etl-service` test gap** — real smoke test added for the one route it exposes, not a placeholder; correctly handles the `DATABASE_URL`-before-import subtlety.

Re-ran the full suites: `auth-service` 71/71 (18 pre-existing DB-integration skips), `supervisor-operations-service` 9/9, `reporting-etl-service` 1/1 — all lint-clean. Nothing outstanding from my review.

PR #64, opened 2026-07-24
fix(auth-service): assign real project/geography scope to seeded users

Dharanishwaran-ND:
## Summary

This PR now bundles 4 separate, self-contained commits:

**1. `fix(auth-service): assign real project/geography scope to seeded users`**
- Seeded SAKHI/SUPERVISOR/MANAGER users previously got `projectId: null, geographyUnitId: null` on their `user_roles` row on every environment — a fresh deploy's `test.sakhi` always logged in with no project/geography scope, even though the geography hierarchy itself was seeded correctly.
- Adds `seedTestProject()` (creates `TEST-PROJ-SEED-01` if missing) and threads the seeded geography chain's leaf unit id into `seedUsersFromEnv`, so newly seeded users get a real, working scope out of the box.
- Fixes a real type bug found while testing: `seedRoles`/`seedAdminUsers` (`src/prisma/startup-seed.ts`) were typed to accept only `PrismaService`, which compiled fine under `nx lint` but failed under `ts-node`'s stricter checking (`npm run prisma:seed`). Relaxed to a structural `SeedPrismaClient` type so both callers (the app's `PrismaService` and the standalone script's plain `PrismaClient`) satisfy it.

**2. `chore(auth-service): wire startup seeding into bootstrap + root-env passthrough`**
- `main.ts` now calls `seedOnStartup(prisma)` right after `prisma.connect()`, before the HTTP server starts listening, so roles/ADMIN self-heal on boot in every environment.
- `tools/prisma-seed-foreach.js` now passes the root `.env`'s `DATABASE_URL`/`DIRECT_URL` through to each service's seed script invocation.
- `.gitignore`: `docs/test-runs/` holds local test-case verification notes and is never committed.

**3. `feat(auth-service): add endpoint to update a user's profile`**
- New ADMIN-only `PATCH /users/:id` to update `displayName`, `mobileNumber`, `email`, and `status` on an existing user.
- `username`, `password`, and role/project/geography assignment are intentionally out of scope (separate concerns/flows).
- Partial, strict Zod schema requiring at least one field; maps a duplicate `mobileNumber`/`email` to 409, unknown user to 404.

**4. `feat(auth-service): add create/update/delete endpoints for geography units`**
- New ADMIN-only `POST /geography-units`, `PATCH /geography-units/:id`, `DELETE /geography-units/:id`, completing full CRUD alongside the existing GET/list/roots/children/ancestors reads.
- Create enforces the fixed 7-level hierarchy order (STATE > DISTRICT > BLOCK > PHC > SUBCENTRE > VILLAGE > PADA) and rejects skipped/mismatched levels.
- Update is limited to `name`/`geoCode`/`status` — `parentId`/`geoType` stay immutable after creation.
- Delete is a soft delete, blocked with 409 if the unit has any active (non-deleted) children.

## Test plan
- [x] `nx test auth-service` — 123/123 passing
- [x] `npm run prisma:seed` — compiles and runs without type errors
- [x] Manually verified live end-to-end for all 4 areas:
  - Seed fix: added a temporary new SAKHI entry, reran the seeder, confirmed the fresh user's JWT contained a real `projectId`/`geographyUnitId` on login; pre-existing seeded users correctly left untouched
  - User update: happy path, unknown-field rejection, empty body, 404 (missing user), 403 (non-ADMIN), 409 (duplicate mobile/email)
  - Geography CRUD: create STATE/DISTRICT, reject wrong hierarchy level, reject duplicate `(parentId, geoType, geoCode)`, update `name`, reject immutable-field update, delete blocked by active children, delete succeeds once children removed, non-ADMIN forbidden

Related to #71

### DIFF
--- a/apps/auth-service/src/geography/geography.service.ts
+++ b/apps/auth-service/src/geography/geography.service.ts
@@ -1,5 +1,25 @@
-import { notFound } from '@armman/service-commons';
+import { badRequest, conflict, notFound } from '@armman/service-commons';
 import type { GeographyRepository } from './geography.repository';
+import type { CreateGeographyUnitInput } from './dto/create-geography-unit.dto';
+import type { UpdateGeographyUnitInput } from './dto/update-geography-unit.dto';
+
+/** Prisma unique-constraint violation code (parentId + geoType + geoCode). */
+const PRISMA_UNIQUE_CONSTRAINT_CODE = 'P2002';
+
+/**
+ * Fixed 7-level hierarchy order (SRS/ERD) — a unit's geoType must be exactly
+ * one level below its parent's geoType. STATE is the only level with no
+ * parent.
+ */
+const GEO_TYPE_ORDER = [
+  'STATE',
+  'DISTRICT',
+  'BLOCK',
+  'PHC',
+  'SUBCENTRE',
+  'VILLAGE',
+  'PADA',
+] as const;
 
 /**
  * Response is projected to EXACTLY the fields the API documents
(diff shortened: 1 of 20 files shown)


apps/auth-service/src/main.ts
```diff
@@ -4,10 +4,12 @@ import { LocalKeypairTokenSigner } from '@armman/service-commons';
 import { appConfig } from './config/app-config';
 import { createApp } from './app.module';
 import { PrismaService } from './prisma/prisma.service';
+import { seedOnStartup } from './prisma/startup-seed';
 
 async function bootstrap(): Promise<void> {
   const prisma = new PrismaService();
   await prisma.connect();
+  await seedOnStartup(prisma);
```
palani-ND, line 12: Pre-existing gap, not introduced by this PR, but now more relevant: `void bootstrap()` has no `.catch()`. This PR adds a new throw path in that call chain (`seedOnStartup` → malformed `ADMIN` env JSON), and the whole point of that validation is to "fail fast, loud" on bad config — but without a catch here, the rejection becomes an unhandled promise rejection rather than a guaranteed, explicit `process.exit(1)`. Node's default unhandled-rejection behavior will likely still crash the process today, but that's implicit, not guaranteed across Node versions/flags. Worth a follow-up: `bootstrap().catch((err) => { console.error(err); process.exit(1); })`.

apps/auth-service/src/geography/geography.service.ts
```diff
@@ -64,4 +84,72 @@ export class GeographyService {
...
+    if (input.geoType === 'STATE') {
+      if (input.parentId) {
+        throw badRequest('parentId: Must be omitted for a STATE-level unit.');
+      }
+    } else {
+      if (!input.parentId) {
+        throw badRequest('parentId: Required for every geoType except STATE.');
+      }
+
+      const parent = await this.repository.findById(input.parentId);
```
palani-ND, line 98: Nit: `create()` checks the parent exists and isn't soft-deleted (via `findById`'s existing `isDeleted: false` filter), but doesn't check the parent's `status` is `ACTIVE`. A new child can be created under an `INACTIVE` parent (e.g. a new PADA under a deactivated VILLAGE). Probably an edge case that doesn't matter much in practice, but worth a deliberate call either way.

palani-ND (commented):
Strong PR overall — reviewed all 4 bundled areas (disclosed transparently in the description, unlike some earlier PRs in this repo). Ran the suite myself: 105/123 passing (18 pre-existing DB-integration skips), matches the claim exactly, lint clean.

Highlights of what's genuinely well done:
- The seed fix isn't just a null-check patch — it's a real, well-reasoned root-cause fix (missing project/geography scope on seeded `user_roles`), found via actual end-to-end testing, with a sensible ordering (geography + project seeded before users so the scope is real, not a placeholder).
- Removing the hardcoded `test.sakhi`/`Test@1234` credential from the codebase in favor of env-driven seed users is a real security improvement, not just a refactor.
- Geography CRUD hierarchy validation (`GEO_TYPE_ORDER`, one-level-below-parent check) and the soft-delete-blocked-by-active-children logic are both correct and well-tested — checked against the repository layer, no gaps found.
- The `P2002`-to-409 mapping pattern is applied consistently and correctly on both new endpoints, verified it can't accidentally swallow the `notFound()` thrown in the same try block (narrow `code` check).

Two minor, non-blocking items:

apps/auth-service/src/main.ts
Dharanishwaran-ND: Good catch — fixed in d5349aa: `bootstrap()` now has `.catch()` that logs and calls `process.exit(1)` explicitly, rather than relying on Node's default unhandled-rejection behavior.

apps/auth-service/src/geography/geography.service.ts
Dharanishwaran-ND: Agreed, this should be a deliberate rejection rather than silently allowed. Fixed in d5349aa: `create()` now throws 400 if the parent's `status` isn't `ACTIVE`. Added a test case and fixed two existing mocks that only stubbed `geoType` on the parent.

PR #65, opened 2026-07-25
feat(auth-service): extend PATCH /users/:id to update credentials, role scope, and Sakhi profile

Dharanishwaran-ND:
## Summary
- Extends the existing `PATCH /users/:id` endpoint (no new endpoints, per decision) to accept, in one request, at least one of:
  - **`users`**: `username`, `password` — added alongside the existing `displayName`/`mobileNumber`/`email`/`status`
  - **`user_roles`**: `projectId`/`geographyUnitId` on the caller's currently-active role row selected by `roleCode` (`roleCode` itself stays immutable — a role change is a new grant, not an edit)
  - **`sakhi_profiles`**: `employeeCode`, `supervisorId`, `phoneNumber`, `backupContact`, `ifscCode`, `panNumber`/`aadhaarNumber`/`bankAccountNumber` (plaintext in, encrypted at rest via `encryptPii`), `activeFrom`/`activeTo`
- All three tables update in a single Prisma transaction — either every requested field lands, or none does.
- Changing `username` or `password` revokes all of that user's active sessions, forcing re-login.
- Response now returns the full profile (same shape as `GET /me`) instead of the slim `CreatedUser` projection, since the endpoint can now touch role/project/geography/profile fields.

## Validation rules
- At least one field required.
- `projectId`/`geographyUnitId` require `roleCode` to be present (and vice versa) — a role-scope update needs both "which role" and "what to change".
- `roleCode` must match a role the user currently holds an active assignment for → 404 otherwise.
- `projectId` must reference an existing, non-deleted project → 404 otherwise.
- Any Sakhi-profile field requires the target user to already have a `SakhiProfile` row → 404 otherwise.
- `activeTo` must not be before `activeFrom`.

## Test plan
- [x] `nx test auth-service` — 162/162 passing
- [x] `nx build auth-service` — clean
- New tests cover: username/password update + session revocation, role/project/geography scope update (including the "no active role" and "no such project" 404 paths), Sakhi profile field updates (including PAN/Aadhaar/bank-account encryption and 404 for users without a profile), and combined multi-table updates in a single request.
- DTO-level tests (`update-user.dto.spec.ts`) cover the cross-field validation rules above.

Related to #72

### DIFF
--- a/apps/auth-service/src/auth/auth.service.ts
+++ b/apps/auth-service/src/auth/auth.service.ts
@@ -200,12 +200,9 @@ export class AuthService {
             }
           : undefined,
         sakhiProfile: sakhiProfileId ? { id: sakhiProfileId, data: sakhiProfileFields } : undefined,
+        revokeSessions: input.username !== undefined || input.password !== undefined,
       });
       if (!user) throw notFound('User not found.');
-
-      if (input.username !== undefined || input.password !== undefined) {
-        await this.repository.revokeAllSessionsForUser(id);
-      }
 
       return toUserProfile(user);
     } catch (err) {
--- a/apps/auth-service/src/auth/auth.service.ts
+++ b/apps/auth-service/src/auth/auth.service.ts
@@ -1,5 +1,5 @@
 import type { TokenSigner } from '@armman/service-commons';
-import { conflict, forbidden, notFound, unauthorized } from '@armman/service-commons';
+import { conflict, notFound, forbidden, unauthorized, encryptPii } from '@armman/service-commons';
 import type { AuthRepository } from './auth.repository';
 import { hashPassword, verifyPassword } from './password';
 import { generateRefreshToken, hashRefreshToken } from './refresh-token';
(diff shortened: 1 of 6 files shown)


apps/auth-service/src/auth/auth.service.ts
```diff
@@ -123,22 +123,96 @@ export class AuthService {
...
-        username: user.username,
-        mobileNumber: user.mobileNumber,
-        displayName: user.displayName,
-        email: user.email,
-        status: user.status,
-        createdAt: user.createdAt,
-      };
+
+      if (input.username !== undefined || input.password !== undefined) {
+        await this.repository.revokeAllSessionsForUser(id);
```
palani-ND: `revokeAllSessionsForUser` runs as a separate call *after* `updateUserTransaction` resolves — it's not part of that same `$transaction`. The PR's own framing is "either everything requested lands, or nothing does," but that atomicity guarantee doesn't extend to the session revocation, even though revocation is the actual security enforcement mechanism behind a username/password change. If this second call fails (a transient DB error, a process crash between the two calls, etc.) after the credential update already committed, the user's password/username has changed but their pre-existing sessions remain valid — silently undermining the "changing username or password revokes all active sessions, forcing re-login" guarantee described in the PR summary. Since Prisma's interactive `$transaction` callback can run arbitrary queries against the same `tx` client, moving the `userSession.updateMany` call inside `updateUserTransaction`'s existing transaction (gated on whether `username`/`password` are present in `fields.user`) would close this gap with a small change.

palani-ND (commented):
Reviewed carefully given the security-sensitive surface (credential changes, session revocation, PII encryption). Ran the suite myself: 144/162 passing (18 pre-existing DB-integration skips), matches the claim exactly, lint clean.

Verified solid:
- `panNumber`/`aadhaarNumber`/`bankAccountNumber` are genuinely encrypted via `encryptPii` before the data ever reaches the repository/transaction — confirmed no plaintext path, and the existing `toUserProfile` mapper (unchanged, pre-existing) never includes `passwordHash` in the response, even though the raw row returned from `updateUserTransaction` does carry it internally.
- The null-vs-undefined tri-state handling for `projectId`/`geographyUnitId`/`activeTo` (omit vs. explicit clear vs. set) is correctly threaded through both the DTO refinements and the service's field-building logic — good test coverage on the "clear to null" path specifically.
- Cross-field DTO validation (`roleCode` required with project/geography fields and vice versa, `activeTo >= activeFrom`) is correct and well-tested at the schema level.
- The three-table transaction (`users`/`user_roles`/`sakhi_profiles`) is genuinely atomic — confirmed it's one `$transaction` block, not three separate writes.

One real gap, inline below.

apps/auth-service/src/auth/auth.service.ts
Dharanishwaran-ND: Good catch — fixed in ab5ffd0. Session revocation now happens inside the same `$transaction` as the rest of the update (gated by a `revokeSessions` flag on `updateUserTransaction`), so it either lands atomically with the username/password change or not at all — no more window where a credential change commits without the accompanying forced-logout.

PR #66, opened 2026-07-27
feat(auth-service): auto-create Sakhi profile on PATCH /users/:id

Dharanishwaran-ND:
## Summary

**1. Auto-create a Sakhi profile on `PATCH /users/:id`**
- Previously 404'd ("User has no Sakhi profile") for any SAKHI whose `sakhi_profiles` row didn't exist yet — which was every SAKHI, since no code path (`POST /users`, the startup seeder) ever created one.
- Now, when a request includes Sakhi-profile fields and no profile exists, one is created atomically in the same transaction, provided:
  - the user holds an active `SAKHI` role (else 400)
  - that role has a `projectId` assigned, used as `primaryProjectId` (else 400)
  - the request includes `phoneNumber`, the profile's other required column not otherwise covered by role/request data (else 400)
- Also folds `revokeSessions` into the same `updateUserTransaction` call in `auth.repository.ts`, closing the non-atomic-session-revocation gap flagged on PR #65 — revocation now lands inside the transaction, not as a separate call after commit.

**2. Repo-wide `zod`/`zod-to-openapi` compatibility fix (unrelated to #1, bundled here)**
- Root `package.json` pins `zod` to `3.23.8` via an explicit dependency + `overrides` entry. npm had been hoisting `zod@3.25.76`, which `@asteasolutions/zod-to-openapi@3.4.0` (a peerDependency-only relationship) can't safely introspect — its `.isOptional()` check throws on newer Zod's internal schema shapes, crashing OpenAPI doc generation at startup.
- This was silently taking down 7 services' Swagger docs (`rules-service`, `risk-referral-service`, `approval-service`, `audit-service`, `supervisor-operations-service`, `media-service`, plus a related `z.coerce.bigint()` crash in `media-service`). Fixed with explicit `.openapi({ type: ... })` annotations on the affected fields, and a new `body` override added to `RouteDocOptions` in `service-commons/docs/documented-router.ts` for the one case (`media-service`'s `sizeBytes`) where no annotation could suppress the crash.
- Restores the full 41-endpoint aggregated Swagger spec at the gateway (previously only 29 endpoints were visible, since a crashed service silently drops out of the aggregation).
- Also includes a lockfile resync (Node 20/npm 10, matching the CI runner) needed after the `zod` pin.

**3. `seedRoles` now creates only missing roles, not all-or-nothing** (`startup-seed.ts`, used by both app boot and `npm run seed`)
- Previously: `if (any role row exists) skip the entire step` — a partially-seeded `roles` table (e.g. only `ADMIN` present, `SAKHI`/`SUPERVISOR`/`MANAGER` missing) would never self-heal, since one existing row was enough to skip creating the other three.
- Now: fetches existing `roleCode`s, diffs against the 4 expected roles, and creates only whichever are actually missing. Existing roles are still left untouched. The result message names exactly which roles were added, for clearer boot logs.

## Test plan
- [x] `nx test auth-service` — 168/168 passing
- [x] `nx build auth-service` — clean
- [x] `nx lint`/`test`/`build` for the 6 zod-to-openapi-affected services (`rules-service`, `risk-referral-service`, `approval-service`, `audit-service`, `supervisor-operations-service`, `media-service`) plus `service-commons` — all clean
- [x] Verified end-to-end against a real local Postgres instance: Sakhi profile row created with correctly encrypted PAN/Aadhaar/bank-account tokens; a follow-up `PATCH` on the same user correctly hits the existing update path (idempotent, no duplicate-row error)
- [x] Verified all 14 services respond `200` on `/docs.json` after the zod fix, and the gateway's aggregated `/api/v1/docs` spec grew from 29 to 41 paths
- [x] New/updated `seedRoles` tests cover: empty table (seeds all 4), all 4 present (skips), and a partially-seeded table (creates only the missing ones)
- [x] CI `verify` check passing on this branch

Related to #73

### DIFF
--- a/apps/approval-service/src/approvals/approvalRequest.controller.ts
+++ b/apps/approval-service/src/approvals/approvalRequest.controller.ts
@@ -13,6 +13,23 @@ import {
 
 extendZodWithOpenApi(z);
 
+// Request DTO annotated with examples for Swagger UI; validation behavior is
+// unchanged (`.openapi()` only attaches documentation metadata).
+// requestPayloadJson/decisionPayloadJson are built on z.lazy() (see
+// create-approvalRequest.dto.ts) — zod-to-openapi cannot introspect z.lazy()
+// on its own, so `type: 'object'` is required here to short-circuit its type
+// inference.
+const createApprovalRequestRequestSchema = createApprovalRequestSchema.extend({
+  requestPayloadJson: createApprovalRequestSchema.shape.requestPayloadJson.openapi({
+    type: 'object',
+    example: {},
+  }),
+  decisionPayloadJson: createApprovalRequestSchema.shape.decisionPayloadJson.openapi({
+    type: 'object',
+    example: {},
+  }),
+});
+
 const approvalRequestSchema = z.object({
   id: z.string().uuid(),
   requestType: z.enum([
@@ -102,7 +119,7 @@ export function createApprovalRequestRouter(service: ApprovalRequestService) {
     },
     trustGatewayIdentity,
     requireRoles('SAKHI', 'SUPERVISOR'),
-    validateBody(createApprovalRequestSchema),
+    validateBody(createApprovalRequestRequestSchema),
     asyncHandler(async (req, res) => {
       const created = await service.create(req.body);
       res.status(201).json(ok(created));
--- a/apps/audit-service/src/audit/auditLog.controller.ts
+++ b/apps/audit-service/src/audit/auditLog.controller.ts
@@ -17,6 +17,16 @@ const jsonValueSchema: z.ZodTypeAny = z.unknown().openapi({
   description: 'Arbitrary JSON value.',
 });
 
+// Request DTO annotated with examples for Swagger UI; validation behavior is
+// unchanged (`.openapi()` only attaches documentation metadata).
+// beforeJson/afterJson are built on z.lazy() (see create-auditLog.dto.ts) —
+// zod-to-openapi cannot introspect z.lazy() on its own, so `type: 'object'`
+// is required here to short-circuit its type inference.
+const createAuditLogRequestSchema = createAuditLogSchema.extend({
+  beforeJson: createAuditLogSchema.shape.beforeJson.openapi({ type: 'object', example: {} }),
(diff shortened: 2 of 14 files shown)


palani-ND (commented):
Ran the suite myself: 149/167 passing (18 pre-existing DB-integration skips), matches the claim exactly, lint clean.

The actual fix (auto-create a Sakhi profile when it doesn't exist) is well-designed and thoroughly tested: the three preconditions (active SAKHI role, that role has a project, `phoneNumber` supplied) are each independently tested for their 400 path, and the concurrent-double-create scenario is safe — `sakhi_profiles.userId` is `@unique` in the schema, so a TOCTOU race between the `findSakhiProfileByUserId` check and the `create` would surface as a clean 409 via the existing generic P2002 handler, not a raw 500 or a duplicate row.

Nice bonus I didn't expect: this also folds `revokeSessions` into the same `$transaction` in `updateUserTransaction` (`auth.repository.ts`) — that closes the exact non-atomic-session-revocation gap I flagged on PR #65. Confirmed it's now inside the transaction, not a separate call after commit.

**Scope note:** the diff also includes a repo-wide `zod` downgrade (3.25.76 → 3.23.8, pinned via a new root `overrides` in `package.json`) plus Swagger/OpenAPI documentation-schema fixes across 6 unrelated services (approval, audit, media, risk-referral, rules, supervisor-operations) — none of which is mentioned in the PR description, which is scoped to "auto-create Sakhi profile." The zod pin in particular is worth calling out explicitly: it's a real, deliberate dependency-version decision (presumably working around a `zod-to-openapi`/`.isOptional()` incompatibility with newer zod, per the `documented-router.ts` comment) that affects every service in the monorepo, not just auth-service — the kind of change a future `git blame` reader would want to find explained in this PR's description rather than only in a code comment three files away from the feature this PR claims to be about.

Nothing blocking — the core feature is solid.

Dharanishwaran-ND:
Good catch, and fair point — updated the PR description to cover both changes explicitly (the zod pin + the 6-service Swagger fix it was masking, plus the lockfile resync it required). Should've been in scope from the start rather than only explained in a code comment. Thanks for running the suite independently too.

PR #67, opened 2026-07-28
feat(auth-service): Add GET /master-data/deltas for offline sync

Dharanishwaran-ND:
## Summary

- Adds `GET /api/v1/master-data/deltas`, the endpoint the HLD specifies (as `GET /sync/deltas`) for Supervisor/Sakhi apps to pull geography, project, and funder master data for offline caching.
- Optional `since` query param (ISO-8601 datetime) filters to rows updated after that time; omitted or empty means a full snapshot.
- Soft-deleted rows are always included (never filtered out) so a client can tell "deleted" apart from "never existed" and prune its local cache.
- Response is `{ serverTime, geographyUnits[], projects[], funders[] }` — `serverTime` is meant to be stored by the client and replayed as the next `since`, avoiding client/server clock skew.
- No new role restriction — open to any authenticated role, matching the existing `/geography-units`, `/projects`, `/funders` GET endpoints this aggregates.
- Wired into the API gateway under `/master-data` (auth required, proxied to auth-service).

## Scope

Lives in `auth-service` since it already owns the `funders`/`projects`/`geography_units` master data tables (see `apps/auth-service/.claude/CLAUDE.md`) — no cross-service joins needed.

## Test plan

- [x] `nx lint auth-service` / `nx lint api-gateway` — clean
- [x] `nx test auth-service` — 185/185 passing (11 new/updated tests: `master-data.service.spec.ts`, `get-master-data-deltas.dto.spec.ts`, `geography.repository.spec.ts` additions)
- [x] `nx build auth-service` / `nx build api-gateway` — clean
- [x] Verified live against a running instance:
  - `GET /master-data/deltas` (no `since`) → 200, full snapshot with real data
  - `GET /master-data/deltas?since=<future date>` → 200, empty arrays
  - `GET /master-data/deltas?since=not-a-date` → 400
  - `GET /master-data/deltas?since=` (empty) → 200, treated as full snapshot
  - `GET /master-data/deltas` with no auth header → 401

Related to #75

### DIFF
--- a/apps/auth-service/src/master-data/master-data.service.ts
+++ b/apps/auth-service/src/master-data/master-data.service.ts
@@ -1,0 +1,88 @@
+import type { GeographyRepository } from '../geography/geography.repository';
+import type { ProjectRepository } from '../projects/project.repository';
+
+/**
+ * Response is projected to exactly the fields a delta client needs —
+ * including `isDeleted`/`deletedAt`, unlike the read APIs' projections
+ * (`toApiGeographyUnit`/`toApiProject`), which deliberately strip them since
+ * a soft-deleted row is never resolvable there in the first place. A delta
+ * client needs to see deletions to prune its local cache.
+ */
+function toApiGeographyUnit(u: Record<string, unknown>) {
+  return {
+    geographyUnitId: u.geographyUnitId,
+    parentId: u.parentId,
+    geoType: u.geoType,
+    geoCode: u.geoCode,
+    name: u.name,
+    status: u.status,
+    updatedAt: u.updatedAt,
+    isDeleted: u.isDeleted,
+    deletedAt: u.deletedAt,
+  };
+}
+
+function toApiFunder(f: Record<string, unknown> | null) {
+  if (!f) return null;
+  return {
+    funderId: f.funderId,
+    funderCode: f.funderCode,
+    funderName: f.funderName,
+    status: f.status,
+    updatedAt: f.updatedAt,
+    isDeleted: f.isDeleted,
+    deletedAt: f.deletedAt,
+  };
+}
+
+function toApiProject(p: Record<string, unknown>) {
+  return {
+    projectId: p.projectId,
+    funderId: p.funderId,
+    funder: toApiFunder(p.funder as Record<string, unknown> | null),
+    projectCode: p.projectCode,
+    projectName: p.projectName,
+    financialYear: p.financialYear,
+    startDate: p.startDate,
+    endDate: p.endDate,
+    status: p.status,
+    updatedAt: p.updatedAt,
+    isDeleted: p.isDeleted,
+    deletedAt: p.deletedAt,
+  };
+}
+
+/**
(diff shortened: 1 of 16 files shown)


apps/auth-service/src/master-data/master-data.service.ts
```diff
@@ -0,0 +1,88 @@
...
+    const sinceDate = since ? new Date(since) : undefined;
+
+    const [geographyUnits, projects, funders] = await Promise.all([
+      this.geographyRepository.findUpdatedSince(sinceDate),
+      this.projectRepository.findProjectsUpdatedSince(sinceDate),
+      this.projectRepository.findFundersUpdatedSince(sinceDate),
+    ]);
+
+    return {
+      serverTime: new Date(),
```
palani-ND: `serverTime` is captured *after* `await Promise.all([...])` resolves, not before the queries run. That ordering has a classic delta-sync race: if a geography/project/funder row is updated at some instant `T` that falls between when the queries actually executed against the DB and when `new Date()` is evaluated here, that row's `updatedAt` could be `< serverTime` even though it was never returned in *this* response (the query already ran before the update landed). The client then stores this response's `serverTime` as its next `since`, and because the filter is strict `updatedAt > since`, that update is permanently skipped — silently missing from every future delta response, not just delayed to the next sync.

The fix is straightforward: capture the timestamp *before* issuing the queries (`const capturedAt = new Date();` above the `Promise.all`, then return `serverTime: capturedAt`) so the snapshot time always precedes the query execution, guaranteeing any row updated during or after the query is still `updatedAt >= capturedAt` and will be picked up next time. Given this endpoint is the sync foundation for offline master data across both mobile apps, this is worth fixing before it ships, even though the real-world race window is narrow (master data changes infrequently) — the failure mode is a silent, hard-to-diagnose permanently-missed update, not a visible error.

palani-ND (commented):
Ran the suite myself: 167/185 passing (18 pre-existing DB-integration skips), matches the claim exactly, lint clean. File scope matches the description exactly this time, no undisclosed surprises.

Verified solid: soft-deleted rows are genuinely included (not filtered) in all three repository methods, confirmed against the actual `where` clauses; the `since` boundary is strict `gt` (correct — a client replaying its own `serverTime` as the next `since` won't re-fetch the row that produced it); the DTO's `.refine()`-based validation (rather than `.transform()`/`.pipe()`) is a sensible, well-explained workaround for the zod-to-openapi introspection issue already established elsewhere in this repo; the null-funder projection and independent-empty-arrays-per-entity cases are both explicitly tested.

One real correctness concern, inline below — everything else looks good.

apps/auth-service/src/master-data/master-data.service.ts
Dharanishwaran-ND: Confirmed and fixed in feb279c — `capturedAt` is now set before `Promise.all` runs and returned as `serverTime`, so any row updated during/after query execution stays >= that timestamp and is picked up on the next sync. Existing `serverTime`-timing test still passes since `capturedAt` is still bounded within the call; full suite (190/190) and lint clean.

PR #68, opened 2026-07-29
fix(api-gateway): Prevent doubled /api/v1 in aggregated Swagger servers

Dharanishwaran-ND:
## Summary
- The aggregated Swagger UI's "Servers" dropdown showed `https://api.armman.org/api/v1/api/v1` instead of `https://api.armman.org/api/v1`, since `createDocsRouter()` unconditionally appended `/api/v1` to every `PUBLIC_BASE_URLS` entry — but that env var already includes the suffix in some deployed environments (e.g. the live `api.armman.org` gateway).
- Extracted and exported `withApiPrefix()`: strips any trailing slash and skips appending `/api/v1` if it's already present, so the result is correct regardless of which convention an environment's `PUBLIC_BASE_URLS` follows.

## Test plan
- [x] New unit tests for `withApiPrefix` (bare URL, already-suffixed URL, trailing slash in both cases) — `apps/api-gateway/src/docs/docs.controller.spec.ts`
- [x] `nx test api-gateway` — 19/19 passing
- [x] `nx lint api-gateway` — clean
- [x] `nx build api-gateway` — clean

Related to #70

### DIFF
--- a/apps/api-gateway/src/docs/docs.controller.spec.ts
+++ b/apps/api-gateway/src/docs/docs.controller.spec.ts
@@ -1,0 +1,19 @@
+import { withApiPrefix } from './docs.controller';
+
+describe('withApiPrefix', () => {
+  it('appends /api/v1 to a bare base URL', () => {
+    expect(withApiPrefix('https://api.armman.org')).toBe('https://api.armman.org/api/v1');
+  });
+
+  it('does not double the suffix when it is already present', () => {
+    expect(withApiPrefix('https://api.armman.org/api/v1')).toBe('https://api.armman.org/api/v1');
+  });
+
+  it('strips a trailing slash before appending', () => {
+    expect(withApiPrefix('https://api.armman.org/')).toBe('https://api.armman.org/api/v1');
+  });
+
+  it('strips a trailing slash when the suffix is already present', () => {
+    expect(withApiPrefix('https://api.armman.org/api/v1/')).toBe('https://api.armman.org/api/v1');
+  });
+});
--- a/apps/api-gateway/src/docs/docs.controller.ts
+++ b/apps/api-gateway/src/docs/docs.controller.ts
@@ -2,6 +2,7 @@ import { Router, type NextFunction, type Request, type Response } from 'express'
 import swaggerUi from 'swagger-ui-express';
 import { appConfig } from '../config/app-config';
 import { createOpenApiAggregator, type DocsService } from './aggregate-openapi';
+import { withApiPrefix } from './with-api-prefix';
 
 /**
  * Every service that publishes an OpenAPI doc, in the order sections should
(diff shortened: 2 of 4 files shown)


palani-ND (commented):
Complete review. Ran the tests myself: 19/19 passing, matches the claim exactly, lint clean.

**The fix itself is correct.** Traced `withApiPrefix` against realistic inputs: bare URL, already-suffixed URL, one-or-more trailing slashes (the regex `/\/+$/` correctly handles multiple, not just one), and already-suffixed-with-trailing-slash — all four covered by the new tests and all behave correctly. Two theoretical, low-probability edge cases I don't think are worth changing: (1) `endsWith('/api/v1')` is case-sensitive, so a base URL with different casing like `/API/V1` would still double up — extremely unlikely for an ops-configured env var; (2) an empty-string element in `PUBLIC_BASE_URLS` would produce a bare `/api/v1` relative path, but that's only reachable if the array's own parsing lets an empty string through, which is outside this diff. Keeping `withApiPrefix` in its own module specifically to dodge `app-config.ts`'s `process.exit(1)`-on-missing-env-var side effect is a smart, well-reasoned call — matches an established pattern elsewhere in this codebase (e.g. `beneficiary-service`'s `geography.client.ts` does the same thing for the same reason).

**The one thing worth raising for completeness**, since this is a root-cause class of bug (inconsistent base-URL conventions across deployed environments), not just a one-off: the exact same unconditional-concatenation pattern exists in two other, more consequential places that this PR doesn't touch — flagged inline below. Those aren't cosmetic Swagger UI issues like this one; if `AUTH_SERVICE_BASE_URL` is ever configured with the `/api/v1` suffix already included in the same way `PUBLIC_BASE_URLS` apparently is on `api.armman.org`, those two call sites would 404 on every geography lookup in production, which is a real functional break, not just a wrong URL in a docs page. Worth a follow-up (possibly hoisting `withApiPrefix` into a shared lib like `libs/service-commons` so all three call sites — and any future ones — share one implementation instead of three copies of the same defensive logic, or three chances to forget it).

---

**Follow-up note (can't attach as an inline comment since these files aren't part of this diff):**

Same latent bug class this PR fixes for the Swagger "Servers" dropdown: `AUTH_SERVICE_BASE_URL` is unconditionally concatenated with `/api/v1/...` here, with no defense against the base URL already including that suffix. Since `AUTH_SERVICE_BASE_URL` is the same kind of externally-configured base URL as `PUBLIC_BASE_URLS` (this PR's own description says that var is "configured inconsistently across this platform's deployed environments — some already include the /api/v1 suffix, some don't"), the same misconfiguration here would produce a doubled path and a real 404 on every PHC/geography lookup in production — not just a cosmetic docs issue. Worth applying the same `withApiPrefix`-style guard (ideally via a shared helper rather than duplicating it, since `visit-form-service`'s `geography.client.ts` has the identical pattern at its own line 27).

Same pattern again at `apps/visit-form-service/src/geography/geography.client.ts:27`.

Dharanishwaran-ND:
Thanks for the thorough review!

**Fix confirmed correct** — re-ran the full test suite in a clean worktree with no `.env` present (matching CI exactly): `nx test api-gateway --coverage --passWithNoTests` → **19/19 passing**. `withApiPrefix()` behaves correctly for bare, already-suffixed, and trailing-slash variants (including multiple trailing slashes).

The isolation fix (2nd commit) is verified doing exactly what it claims: moving `withApiPrefix` into its own dependency-free module so tests don't crash via `app-config.ts`'s `process.exit(1)` on missing env vars in CI.

**Follow-up noted, intentionally left out of this PR's scope**: the same unguarded-concat pattern at `AUTH_SERVICE_BASE_URL` usage and `apps/visit-form-service/src/geography/geography.client.ts:27` is real and still open — agreed it's a more consequential (functional, not cosmetic) version of this bug. Will track as a separate follow-up rather than scope-creep this PR; happy to open an issue for it if that's useful.

The two low-probability edge cases (case-sensitive suffix match, empty-string array element) are accepted as-is per your review — not worth guarding against given they'd require an already-broken upstream config to trigger.

PR #69, opened 2026-07-29
feat: Sakhi lookups, inventory item/transaction CRUD, and ADMIN full access

Dharanishwaran-ND:
## Summary
- Add project-scoped Sakhi read endpoints (`GET /projects/:projectId/sakhis`, `GET /sakhis/:sakhiId`) in auth-service, routed through the gateway.
- Add inventory item and inventory transaction CRUD in supervisor-operations-service (`POST /inventory-items`, `GET/POST /inventory-transactions`, `GET /inventory-transactions/by-sakhi/:sakhiId`, `PUT`/`DELETE /inventory-transactions/:id`), backed by a cross-service `SakhiClient` that verifies Sakhi-Supervisor assignment via HTTP (no cross-service DB joins, per the forklift rule).
- Harden all new endpoints against IDOR: Supervisors can only act on Sakhis assigned to them and transactions they created; MANAGER is unrestricted on reads; **ADMIN now has full, unrestricted access across every inventory-item/inventory-transaction endpoint** (role gate + ownership-check bypass), matching how MANAGER already behaved on read paths.
- Fix a real gateway routing bug: `GET /sakhis/:sakhiId/inventory-transactions` was unreachable through the gateway proxy (parameterized prefixes aren't supported by the underlying `http-proxy-middleware` rewrite) — moved to `GET /inventory-transactions/by-sakhi/:sakhiId`, which falls under the existing static `/inventory-transactions` gateway route.
- Fix a data-integrity bug: duplicate Karnataka STATE rows in `geography_units` (Postgres NULL≠NULL breaks the unique index for STATE-level rows since `parentId` is always null there) — added an explicit app-level uniqueness check.
- Fix a race condition in `GET /master-data/deltas`: `serverTime` is now captured before the delta queries run, not after, so clients no longer miss updates that land in the query window.
- Fix seed-data bugs in `mother-registration.json` (children-under-5 validation range, split combined name field into first/middle/last, corrected `beneficiary_id`/`unique_id`/`registration_date` field config to use valid `computedFrom` enum values).
- Add per-level geography tables (`states` through `padas`) alongside the existing `geography_units` table (additive-only migration, old table kept as a rollback safety net) — Phase 1 of a larger schema migration, scoped down per explicit review.

## Test plan
- [x] `npx nx test supervisor-operations-service` — 34/34 passing, including new ADMIN-bypass coverage on all 4 ownership-checked service methods
- [x] `npx nx lint supervisor-operations-service` — clean
- [x] `npx nx build supervisor-operations-service` — succeeds
- [x] Live E2E curl testing of all inventory-transaction endpoints across success/failure/IDOR scenarios (SUPERVISOR, MANAGER, ADMIN, cross-Supervisor rejection)
- [x] Verified ADMIN can now list/create/update/delete inventory transactions and inventory items regardless of Sakhi assignment or transaction ownership
- [x] Verified the `/inventory-transactions/by-sakhi/:sakhiId` route is reachable end-to-end through the gateway (previously 400'd due to the routing bug above)

Related to #74

### DIFF
--- a/apps/api-gateway/src/proxy/routes.ts
+++ b/apps/api-gateway/src/proxy/routes.ts
@@ -54,6 +54,9 @@ export const SERVICE_ROUTES: readonly ServiceRoute[] = [
   { prefix: '/lookups', target: appConfig.AUTH_SERVICE_URL, requiresAuth: true },
   { prefix: '/geography-units', target: appConfig.AUTH_SERVICE_URL, requiresAuth: true },
   { prefix: '/master-data', target: appConfig.AUTH_SERVICE_URL, requiresAuth: true },
+  // Sakhi profile reads (list under a project, single lookup) — used by the
+  // Supervisor's Sakhi picker/detail header on inventory/meeting screens.
+  { prefix: '/sakhis', target: appConfig.AUTH_SERVICE_URL, requiresAuth: true },
   // NOTE: `/docs` is NOT proxied here. The gateway serves ONE aggregated
   // Swagger UI at `/api/v1/docs` (see docs/docs.controller.ts) that merges
   // every service's own `/docs.json` into a single page — there is no single
--- a/apps/auth-service/src/app.module.ts
+++ b/apps/auth-service/src/app.module.ts
@@ -19,6 +19,7 @@ import { createProjectModule } from './projects/project.module';
 import { createLookupModule } from './lookups/lookup.module';
 import { createGeographyModule } from './geography/geography.module';
 import { createMasterDataModule } from './master-data/master-data.module';
+import { createSakhiModule } from './sakhis/sakhi.module';
 import { buildAuthServiceOpenApiDocument } from './docs/openapi';
 
 // Re-export shared HTTP helpers so feature routers can import from a single place.
@@ -89,6 +90,7 @@ export function createApp(prisma: PrismaService, signer: TokenSigner, redis: Red
   const lookupModule = createLookupModule(prisma, signer);
   const geographyModule = createGeographyModule(prisma, signer);
   const masterDataModule = createMasterDataModule(prisma, signer);
+  const sakhiModule = createSakhiModule(prisma, signer);
 
   // All routes live under the global `api/v1` prefix.
   const api = express.Router();
@@ -104,6 +106,7 @@ export function createApp(prisma: PrismaService, signer: TokenSigner, redis: Red
         lookupModule.registry,
         geographyModule.registry,
         masterDataModule.registry,
+        sakhiModule.registry,
       ),
     ),
   );
@@ -114,6 +117,7 @@ export function createApp(prisma: PrismaService, signer: TokenSigner, redis: Red
   api.use(lookupModule.router);
   api.use(geographyModule.router);
   api.use(masterDataModule.router);
+  api.use(sakhiModule.router);
   app.use('/api/v1', api);
 
   app.use(notFoundHandler);
--- a/apps/auth-service/src/sakhis/sakhi.controller.ts
+++ b/apps/auth-service/src/sakhis/sakhi.controller.ts
@@ -1,0 +1,103 @@
+import { extendZodWithOpenApi } from '@asteasolutions/zod-to-openapi';
+import { z } from 'zod';
+import type { TokenSigner } from '@armman/service-commons';
+import type { SakhiService } from './sakhi.service';
+import {
+  asyncHandler,
+  authenticate,
+  createDocumentedRouter,
+  errorResponse,
+  ok,
+  requireRoles,
+  unauthorized,
+  validate,
+} from '../app.module';
+
+extendZodWithOpenApi(z);
+
+const projectIdParamsSchema = z
+  .object({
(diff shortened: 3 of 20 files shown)


palani-ND (commented):
**Please retarget this PR's base branch before merging.** It's currently opened against `main`, which is just this repo's original "Initial commit" and has never been updated since (`git log origin/develop..origin/main` is empty; `develop` is 216 commits ahead of it). That's why GitHub shows 591 changed files / +56337 — it's diffing this branch against essentially nothing, not against the actual codebase. I recomputed the diff against `develop` instead and it's a clean, reviewable **20 files, +1305/-8**, matching this PR's actual description. All findings below are from that correct diff — retargeting the base to `develop` on GitHub should immediately shrink the visible diff down to this same set with no code changes needed.

---

Modulo the base-branch mixup, the actual change is solid. Ran the tests myself: `supervisor-operations-service` 34/34 (matches claim exactly), `auth-service` 185/203 (18 pre-existing DB-integration skips), both lint-clean.

The IDOR-hardening work is well done — traced the full authorization path for all 4 ownership-checked methods:
- `createInventoryTransactions`/`listInventoryTransactionsBySakhi`: ownership verified via `SakhiClient` → auth-service's `GET /sakhis/:id` (cross-service, no DB join, per the forklift rule), correctly forwarding the *caller's own* token rather than inventing a service credential — consistent with the established pattern from `beneficiary-service`'s `geography.client.ts`. `sakhi.supervisorId !== caller.id` is a correct, simple equality check.
- `updateInventoryTransaction`/`deleteInventoryTransaction`: correctly check `existing.supervisorId !== caller.id`, bypassed for MANAGER/ADMIN.
- `supervisorId` is never client-suppliable in the create DTO — always the authenticated caller's own id, so a Supervisor can never write a transaction under someone else's name. Confirmed by reading the DTO's own doc comment and the schema (`supervisorId` genuinely absent from `createInventoryTransactionSchema`).
- Item existence/active-status validation happens *before* any DB write, and the per-item creates are wrapped in one `$transaction` so a partial failure can't half-write a multi-item submission.

Also verified the two previously-known bugs are correctly fixed:
- **`master-data.service.ts`**: `serverTime` (renamed `capturedAt`) is now captured *before* `Promise.all(...)` runs, not after — this closes the delta-sync race I flagged on PR #67 exactly as suggested.
- **Geography STATE duplicate bug**: correctly diagnosed and fixed — Postgres's `NULL ≠ NULL` means the existing `@@unique([parentId, geoType, geoCode])` constraint can never catch two STATE rows (both have `parentId = null`), so an explicit app-level uniqueness check was added for that case specifically. Good root-cause fix, not a band-aid.

One very minor observation, not a bug: `isPrivileged()` treats MANAGER as unrestricted for `updateInventoryTransaction`/`deleteInventoryTransaction`, but those routes' `requireRoles('SUPERVISOR', 'ADMIN')` gate means MANAGER can never actually reach that service method in the first place — harmless dead branch, not worth changing, just noting it's unreachable as written.

Dharanishwaran-ND:
Thanks for the thorough review.

**Base branch**: confirmed and fixed — retargeted this PR from `main` to `develop`. You were right: `main` was 216 commits stale (never updated past the initial commit), which is why GitHub was showing the diff against essentially nothing. The PR now correctly shows 20 files changed, +1305/-8, matching your recomputed diff exactly.

**IDOR/ownership review**: appreciate you tracing the full authorization path on all 4 methods — that matches the intent exactly (cross-service ownership check via `SakhiClient` using the caller's own forwarded token, `supervisorId` never client-suppliable, atomic multi-item creates). Good catch confirming the two previously-flagged bugs (delta-sync race, geography STATE duplicate) are properly fixed at the root cause rather than patched around.

**`isPrivileged()` dead branch on update/delete**: agreed, and noted — MANAGER can't reach `updateInventoryTransaction`/`deleteInventoryTransaction` today since those routes only allow `SUPERVISOR`/`ADMIN`. Leaving it as-is per your note (harmless, and keeps `isPrivileged()` consistent across all 4 methods rather than special-casing two of them) — happy to remove it if you'd rather it not exist at all.

Dharanishwaran-ND:
Thanks for catching the base-branch issue and for the detailed trace through the IDOR hardening!

**Base branch**: already retargeted to `develop` — diff is now the clean 20 files / +1305/-8 matching the PR description, confirmed via `gh pr view`.

**Re-verified independently**:
- `nx test supervisor-operations-service` → **34/34 passing**, including ADMIN-bypass coverage on all 4 ownership-checked methods (`createInventoryTransactions`, `listInventoryTransactionsBySakhi`, `updateInventoryTransaction`, `deleteInventoryTransaction`).
- Race-condition fix in `master-data.service.ts`: confirmed `capturedAt` (renamed from `serverTime`) is captured before `Promise.all(...)` runs — matches your PR #67 note exactly.
- Geography STATE duplicate fix: confirmed the explicit app-level uniqueness check is in place with a comment explaining the Postgres `NULL ≠ NULL` gap in `@@unique([parentId, geoType, geoCode])` — same fix applied consistently in both create and update paths.

**Dead-branch note on `isPrivileged()`/MANAGER** — agreed, left as-is per your call; not worth churning code for an unreachable branch guarded by the route-level role gate.

No further changes needed from my side — ready for merge whenever you're satisfied.

PR #92, opened 2026-07-31
Add call-sheet CRUD, meeting/training lifecycle APIs, and fix media upload bugs

### LINKED ISSUE #93: Supervisors can now log calls to Sakhis and see their full call history
This work has been completed.

Previously, Supervisors had no way to record a call sheet entry for a Sakhi, review a past call's details, update it after the call ended, or see a Sakhi's full call history — only a basic, unfiltered list existed.

What was delivered:
- Supervisors can now log a call to a Sakhi, recording when it happened, its outcome (connected, busy, switched off, etc.), notes, and any follow-up action needed.
- After a call ends, Supervisors can go back and update it — recording the final outcome, how long the call lasted, and any follow-up notes captured once the conversation was over.
- Supervisors can view the full call history for any Sakhi assigned to them, newest first, including all notes and follow-up actions.
- A "recently contacted" view was added so Supervisors can quickly see which Sakhis have already been called recently, supporting the call sheet's at-a-glance highlight feature.
- Security was built in from the start: a Supervisor can only log or view calls for Sakhis actually assigned to them, while Managers and Admins can see and manage call records across all Supervisors.

Together, these changes give Supervisors a complete, reliable call sheet — from logging a call through to reviewing history — closing a gap where only viewing a raw list of recent calls was possible before.

### LINKED ISSUE #94: Supervisors can now manage the full lifecycle of meetings and trainings, including attendance
This work has been completed.

Previously, Supervisors could only create a meeting or training event and see a basic list of them — there was no way to look up a single event's details, cancel it, mark it complete, filter the list, or record who attended.

What was delivered:
- The meetings/training list can now be filtered by status (scheduled, completed, cancelled) and by type (meeting or training), instead of always showing everything.
- Supervisors can now open a single event to see its full details.
- Supervisors can cancel a scheduled meeting or training that's no longer happening.
- Supervisors can mark a meeting or training as complete — the system requires a photo and at least one attendance record to be in place first, matching the program's existing rule that an event isn't done until both are captured.
- Supervisors can record and review attendance for each Sakhi at a meeting or training, including pre- and post-training scores where applicable. Submitting attendance again for the same event safely updates the existing record instead of creating duplicates.
- Security was built in from the start: a Supervisor can only manage events and attendance they created, while Managers and Admins can act across all Supervisors' events.

Together, these changes turn meeting/training scheduling into a complete workflow — Supervisors can now take an event all the way from creation through attendance-taking to completion, not just schedule it.

### LINKED ISSUE #95: Fixed photo/media upload not working, and a crash right after a successful upload
This work has been completed.

Photo and other media uploads (such as the event-completion photo required for meetings and trainings) were silently failing — every upload attempt was rejected, regardless of what was sent, due to a technical mismatch in how the file's verification checksum was being validated. On top of that, once the underlying issue was fixed, a second problem surfaced: the upload would actually succeed and save correctly, but the confirmation response back to the app would crash, because a numeric file-size field couldn't be converted to the format needed for the response.

What was delivered:
- Media/photo uploads now work correctly end-to-end — verified with a real upload request that completed successfully.
- Fixed the crash that happened immediately after a successful upload, so the app now receives a proper confirmation instead of an error.
- Added test coverage so this checksum validation can't silently break again in the future.

Together, these fixes unblock any feature that depends on photo/media upload — including completing meetings and trainings, which requires a photo before it can be marked done.

### LINKED ISSUE #96: Add role-based scoping to GET /beneficiaries (Sakhi/Supervisor/Admin)
## Problem

`GET /beneficiaries` returned all beneficiary cases to any caller with SAKHI/SUPERVISOR/MANAGER, with no scoping by role. Per SRS FR-S-9.2, access should be:

- **Sakhi**: only cases they personally created (`sakhi_id` on `beneficiary_cases` matches their own id).
- **Supervisor**: only cases belonging to Sakhis they supervise (resolved via auth-service's `GET /projects/:projectId/sakhis`, filtered by `supervisorId`).
- **Admin** (and Manager): all cases, unscoped.

## Fix

- `beneficiary.service.ts`'s `list()` now takes the authenticated caller and derives `BeneficiaryListFilters.sakhiId` (SAKHI) or `.sakhiIds` (SUPERVISOR, via a new `listSakhiIdsForSupervisor` client call to auth-service) before querying.
- `beneficiary.repository.ts` applies `where.sakhiId` / `where.sakhiId: { in: ... }` accordingly.
- `ADMIN` added to the route's `requireRoles(...)` list (previously only SAKHI/SUPERVISOR/MANAGER could call the endpoint at all).
- New `apps/beneficiary-service/src/sakhi/sakhi.client.ts` — thin HTTP client calling auth-service's existing `/projects/:projectId/sakhis`, no new auth-service endpoint needed.

## Verification

Verified live against a running gateway with real JWTs for all four scenarios:
- Sakhi A sees only her own cases.
- Sakhi B (different Sakhi, same Supervisor as A) sees only her own cases.
- Supervisor supervising both A and B sees the union of both their cases.
- A different Supervisor (supervising neither) correctly sees zero cases — no cross-supervisor leakage.
- Admin sees all cases unscoped.

### LINKED ISSUE #97: auth-service GET /sakhis returns sakhi_profiles PK instead of users.user_id, breaking Supervisor-scoped queries
## Bug

`GET /projects/:projectId/sakhis` and `GET /sakhis/:sakhiId` (`sakhi.service.ts`'s `toApiSakhi()`) returned `sakhiId: profile.id` — the `sakhi_profiles` table's own primary key.

But `beneficiary_cases.sakhi_id` is populated with the Sakhi's `users.user_id` (same value as their JWT `sub`), which is also what beneficiary-service's own SAKHI-role filter (`caller.id`) already used.

Two different id spaces were being compared for the same person, so `beneficiary.service.ts`'s Supervisor-scoping branch (which resolves Sakhi ids via this endpoint, see #96) always failed to match — a Supervisor's `GET /beneficiaries` call returned an empty list regardless of real Sakhi assignment.

## Fix

- `sakhi.service.ts`: `toApiSakhi()` now returns `sakhiId: user.id` instead of `profile.id`.
- `sakhi.repository.ts`: `findById()` now queries `sakhi_profiles` by `userId` instead of the profile's own `id`, since the `:sakhiId` route param is now user-id-shaped everywhere.
- `sakhi.controller.ts`: clarifying comment on the param schema.
- Updated `sakhi.service.spec.ts` / `sakhi.repository.spec.ts` fixtures/assertions to match.

No DB migration needed — this only changes which column backs an API response field. Existing data was unaffected (it already stored `user_id`).

## Verification

Confirmed live via the gateway with real Sakhi/Supervisor/Admin JWTs — see #96 for the full scenario matrix, all passing after this fix (previously the single/only Supervisor tested got 0 results incorrectly).

Dharanishwaran-ND:
Closes #93, closes #94, closes #95, closes #96, closes #97

## Summary
- **Call Sheet (FR-SV-3.1–3.4)** — closes #93: added `POST /call-logs`, `GET`/`PATCH /call-logs/:callLogId`, and `GET /call-logs/by-sakhi/:sakhiId(/recent)` — only list/no-filter existed before. Ownership mirrors the existing `inventory-transactions` pattern (`supervisorId` always server-derived, `SakhiClient`-checked, MANAGER/ADMIN unrestricted). The Sakhi-scoped routes sit under `/call-logs/by-sakhi/:sakhiId` rather than `/sakhis/:sakhiId/call-logs`, since the gateway's proxy can't rewrite a mid-path param into a different downstream service (`/sakhis` already routes to `auth-service`) — confirmed by reading `register-proxies.ts`.
- **Meeting/Training lifecycle (FR-SV-2.1–2.4)** — closes #94: added `GET /supervisor-events?status=&eventType=` (filtering), `GET /supervisor-events/:id`, `PATCH .../:id/cancel`, `PATCH .../:id/complete`, and `GET`/`PUT .../:id/attendance`. Cancel/complete are restricted to `SCHEDULED` events (409 otherwise); complete additionally requires a photo (existing rule) and at least one recorded attendance row (FR-SV-2.3's "mark attendance ... upload photo ... submit" flow). Attendance upserts by `(eventId, sakhiId)` since no DB unique constraint exists on that pair.
- **media-service fix** — closes #95: `POST /media` validated `checksum` as `z.instanceof(Buffer)`, which `express.json()` can never produce from a JSON body — the endpoint was unreachable over real HTTP with any client, always failing validation regardless of input. Now accepts a 64-char hex SHA-256 string and transforms it to a `Buffer` for the `Bytes` column.
- **media-service fix (found while testing the above)** — also part of #95: once `POST /media` could actually be reached, it 500'd — `sizeBytes` is a `BigInt` on the Prisma model, and `JSON.stringify` can't serialize it. `res.json()` was crashing after the DB write had already succeeded. Both `GET`/`POST /media` responses now stringify `sizeBytes`, matching what the response schema already documented.
- **Role-based scoping on `GET /beneficiaries` (SRS FR-S-9.2)** — closes #96: Sakhi callers now see only their own cases, Supervisor callers see only cases belonging to their own Sakhis (resolved via auth-service's existing `GET /projects/:projectId/sakhis`, filtered by `supervisorId` — no new auth-service endpoint), and Manager/Admin see everything unscoped. `ADMIN` was also missing from the route's allowed roles entirely. New `apps/beneficiary-service/src/sakhi/sakhi.client.ts` calls auth-service to resolve a Supervisor's Sakhi ids.
- **auth-service id-mismatch fix, found while verifying the above** — closes #97: `GET /projects/:projectId/sakhis` / `GET /sakhis/:sakhiId` returned `sakhiId: profile.id` (the `sakhi_profiles` row's own PK), but `beneficiary_cases.sakhi_id` is populated with the Sakhi's `users.user_id` (the JWT `sub`) — the same id beneficiary-service's own SAKHI-role filter already uses. The mismatched id spaces meant a Supervisor's `GET /beneficiaries` call always returned zero rows regardless of real Sakhi assignment. Fixed `toApiSakhi()` to return `user.id`, and `sakhi.repository.ts`'s `findById` to query by `userId`. No DB migration needed — only changes which column backs the API field.

## Test plan
- [x] `nx lint supervisor-operations-service` / `nx lint media-service` — clean
- [x] `nx test supervisor-operations-service` — 94/94 passing (29 new call-log tests, 31 new event/attendance tests)
- [x] `nx test media-service` — 12/12 passing (8 new checksum DTO tests)
- [x] `nx build` both services — clean, including OpenAPI doc generation (a real risk given `zod-to-openapi`'s inability to introspect `ZodEffects`/transform chains — caught and fixed a genuine crash-at-boot here)
- [x] Live end-to-end verification against a running gateway + both services with real JWTs (SUPERVISOR and SAKHI roles): full call-log CRUD cycle, full meeting/training lifecycle including the complete-with-photo-and-attendance happy path, and 400/403/404/409/422 error paths all confirmed with real HTTP requests, not just mocks
- [x] `nx affected -t lint test` (auth-service, beneficiary-service) — clean
- [x] Live end-to-end verification of `GET /beneficiaries` role-scoping against a running gateway with real JWTs across a 5-scenario matrix: Sakhi A sees only her own cases; Sakhi B (different Sakhi, same Supervisor as A) sees only her own; a Supervisor supervising both A and B sees the union of both; a different Supervisor (supervising neither) correctly sees zero cases (no cross-supervisor leakage); Admin sees all cases unscoped

### DIFF
--- a/apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
+++ b/apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
@@ -5,6 +5,7 @@ import {
   notFound,
   normalizeForSearch,
   unprocessable,
+  type AuthenticatedUser,
 } from '@armman/service-commons';
 import type {
   BeneficiaryRiskConditionSummary,
@@ -16,6 +17,7 @@ import { computeBmi, withDecryptedName } from './beneficiary.mapper';
 import type { BeneficiaryListFilters, BeneficiaryRepository } from './beneficiary.repository';
 import { deriveFullName, type CreateBeneficiaryInput } from './dto/create-beneficiary.dto';
 import { resolveHealthBlockIdFromPhc } from '../geography/geography.client';
+import { listSakhiIdsForSupervisor } from '../sakhi/sakhi.client';
 
 const GESTATION_DAYS = 280;
 
@@ -37,11 +39,20 @@ export class BeneficiaryService {
   constructor(private readonly repository: BeneficiaryRepository) {}
 
   /**
-   * Lists beneficiary cases per SRS FR-S-9.2 / HLD's filter set (scope
-   * enforcement added with the auth layer). Each row's name is decrypted
-   * server-side for display — the search hash itself is never returned.
+   * Lists beneficiary cases per SRS FR-S-9.2 / HLD's filter set, scoped by
+   * the caller's role: a SAKHI only ever sees their own cases (their own id
+   * always wins over anything else), a SUPERVISOR only sees cases belonging
+   * to their own Sakhis (resolved via auth-service's existing
+   * `/projects/:projectId/sakhis`, filtered by supervisorId — no new
+   * auth-service endpoint), and MANAGER/ADMIN see everything unscoped. Each
+   * row's name is decrypted server-side for display — the search hash
+   * itself is never returned.
    */
-  async list(query: ListBeneficiariesQuery) {
+  async list(
+    query: ListBeneficiariesQuery,
+    caller: AuthenticatedUser,
+    authorizationHeader: string,
+  ) {
     const filters: BeneficiaryListFilters = {
       projectId: query.projectId,
       villageId: query.villageId,
@@ -54,6 +65,16 @@ export class BeneficiaryService {
         ? hashForSearch(normalizeForSearch(query.mobileNumber))
         : undefined,
     };
+
+    if (caller.roles.includes('SAKHI')) {
+      filters.sakhiId = caller.id;
+    } else if (caller.roles.includes('SUPERVISOR')) {
+      filters.sakhiIds = await listSakhiIdsForSupervisor(
+        caller.projectId ?? '',
+        caller.id,
+        authorizationHeader,
+      );
+    }
 
     const cases = await this.repository.findMany(filters);
     return cases.map(withDecryptedName);
--- a/apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
+++ b/apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
@@ -1,6 +1,7 @@
 import { addDays } from '@armman/core';
 import {
   encryptPii,
+  forbidden,
   hashForSearch,
   notFound,
   normalizeForSearch,
@@ -69,8 +70,14 @@ export class BeneficiaryService {
     if (caller.roles.includes('SAKHI')) {
       filters.sakhiId = caller.id;
     } else if (caller.roles.includes('SUPERVISOR')) {
+      // Per the SRS, a Supervisor has exactly one project — a caller missing
+      // this claim is an invalid/inconsistent identity, not a "no project"
+      // case to silently degrade into a malformed `/projects//sakhis` path.
+      if (!caller.projectId) {
+        throw forbidden('Supervisor caller has no project scope.');
+      }
       filters.sakhiIds = await listSakhiIdsForSupervisor(
-        caller.projectId ?? '',
+        caller.projectId,
         caller.id,
         authorizationHeader,
       );
--- a/apps/beneficiary-service/src/sakhi/sakhi.client.ts
+++ b/apps/beneficiary-service/src/sakhi/sakhi.client.ts
@@ -18,6 +18,12 @@ interface ApiSakhi {
  * Called through the gateway (per AUTH_SERVICE_BASE_URL) so the gateway can
  * verify `authorizationHeader` — the original Supervisor caller's own bearer
  * token, forwarded unchanged, matching geography.client.ts's convention.
+ *
+ * Reads the full response body with no pagination handling — safe today
+ * because `SakhiRepository.findByProject` (auth-service) queries with no
+ * `take`/`skip`. If that endpoint is ever paginated, this must change too,
+ * or a Supervisor with Sakhis beyond the first page would silently see an
+ * under-scoped (incomplete) beneficiary list instead of an error.
  */
 export async function listSakhiIdsForSupervisor(
   projectId: string,
--- a/apps/beneficiary-service/src/sakhi/sakhi.client.ts
+++ b/apps/beneficiary-service/src/sakhi/sakhi.client.ts
@@ -1,0 +1,42 @@
+import { badGateway } from '@armman/service-commons';
+
+// Read directly (not via appConfig) so importing this client doesn't pull in
+// app-config's full schema — see geography.client.ts for the same rationale.
+const AUTH_SERVICE_BASE_URL = process.env.AUTH_SERVICE_BASE_URL ?? 'http://localhost:3000';
+
+interface ApiSakhi {
+  sakhiId: string;
+  supervisorId: string | null;
+}
+
+/**
+ * Resolves the Sakhi ids reporting to a given Supervisor, via auth-service's
+ * existing `GET /projects/:projectId/sakhis` (no new auth-service endpoint —
+ * that route already returns each Sakhi's `supervisorId`). Used to scope
+ * `GET /beneficiaries` to a Supervisor's own Sakhis' cases.
+ *
+ * Called through the gateway (per AUTH_SERVICE_BASE_URL) so the gateway can
+ * verify `authorizationHeader` — the original Supervisor caller's own bearer
+ * token, forwarded unchanged, matching geography.client.ts's convention.
+ */
+export async function listSakhiIdsForSupervisor(
+  projectId: string,
+  supervisorId: string,
+  authorizationHeader: string,
+): Promise<string[]> {
+  let res: Response;
+  try {
+    res = await fetch(`${AUTH_SERVICE_BASE_URL}/api/v1/projects/${projectId}/sakhis`, {
+      headers: { Authorization: authorizationHeader },
+    });
+  } catch {
+    throw badGateway('Unable to resolve Sakhis — the auth service is unreachable.');
+  }
+
+  if (!res.ok) {
+    throw badGateway('Unable to resolve Sakhis — the auth service returned an error.');
+  }
+
+  const body = (await res.json()) as { data: ApiSakhi[] };
+  return body.data.filter((sakhi) => sakhi.supervisorId === supervisorId).map((s) => s.sakhiId);
+}
--- a/apps/media-service/src/media/mediaAsset.controller.ts
+++ b/apps/media-service/src/media/mediaAsset.controller.ts
@@ -2,6 +2,7 @@ import { extendZodWithOpenApi } from '@asteasolutions/zod-to-openapi';
 import { z } from 'zod';
 import type { MediaAssetService } from './mediaAsset.service';
 import { createMediaAssetSchema } from './dto/create-mediaAsset.dto';
+import type { MediaAsset } from '../../../../node_modules/.prisma/client-media-service';
 import {
   asyncHandler,
   createDocumentedRouter,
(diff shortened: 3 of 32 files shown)


apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
```diff
@@ -55,6 +66,16 @@ export class BeneficiaryService {
         : undefined,
     };
 
+    if (caller.roles.includes('SAKHI')) {
+      filters.sakhiId = caller.id;
+    } else if (caller.roles.includes('SUPERVISOR')) {
+      filters.sakhiIds = await listSakhiIdsForSupervisor(
+        caller.projectId ?? '',
```
palani-ND: **Silent fallback on missing `projectId`.** If a SUPERVISOR caller's `projectId` is ever `null` (JWT missing the claim, migration gap, etc.), this falls back to `''` and calls `listSakhiIdsForSupervisor('', caller.id, ...)`, hitting `/api/v1/projects//sakhis` — a malformed path — instead of failing loudly. Per the SRS comment ("one project per Supervisor"), a SUPERVISOR without a `projectId` is really an invalid/inconsistent state; consider throwing a clear 403/500 here rather than silently degrading to an empty-string path segment, so the failure mode is obvious rather than a confusing downstream 404 from the gateway.

apps/beneficiary-service/src/sakhi/sakhi.client.ts
```diff
@@ -0,0 +1,42 @@
...
+  } catch {
+    throw badGateway('Unable to resolve Sakhis — the auth service is unreachable.');
+  }
+
+  if (!res.ok) {
+    throw badGateway('Unable to resolve Sakhis — the auth service returned an error.');
+  }
+
+  const body = (await res.json()) as { data: ApiSakhi[] };
+  return body.data.filter((sakhi) => sakhi.supervisorId === supervisorId).map((s) => s.sakhiId);
```
palani-ND, line 47: This reads the full response body from `GET /projects/:projectId/sakhis` and filters client-side by `supervisorId`, with no pagination handling. If that auth-service endpoint ever paginates (or caps results, as `findEvents`/`findCallLogsBySakhi` elsewhere in this repo do with `take: 50`), a Supervisor with Sakhis beyond the first page would silently see an incomplete (under-scoped) beneficiary list rather than an error. Worth confirming the endpoint is unbounded, or handling pagination here if it isn't.

apps/supervisor-operations-service/src/operations/operations.repository.ts
```diff
@@ -24,6 +32,75 @@ export class OperationsRepository {
...
+    entries: UpdateAttendanceInput['attendance'],
+    userId: string,
+  ) {
+    const existingRows = await this.prisma.eventAttendance.findMany({
+      where: { eventId, isDeleted: false, sakhiId: { in: entries.map((e) => e.sakhiId) } },
+    });
+    const existingBySakhiId = new Map(existingRows.map((row) => [row.sakhiId, row]));
+
+    return this.prisma.$transaction(
+      entries.map((entry) => {
```
palani-ND, line 77: **Possible duplicate rows within a single submission.** `existingBySakhiId` is built once from `findMany` *before* this loop and never updated as entries are processed. If the client's `attendance` array contains two entries for the same `sakhiId` (nothing in `updateAttendanceSchema`/`attendanceEntrySchema` rejects that), both would independently resolve to "not existing" (when there's no prior row) and both branch to `.create(...)` inside the same `$transaction`, producing two rows for the same `(eventId, sakhiId)` — since, as this file's own comment notes, there's no DB unique constraint on that pair to fall back on. Worth either de-duping `entries` by `sakhiId` before building the map, or rejecting duplicate `sakhiId`s in `updateAttendanceSchema` via `.refine()`.

apps/media-service/src/media/mediaAsset.controller.ts
```diff
@@ -2,6 +2,7 @@ import { extendZodWithOpenApi } from '@asteasolutions/zod-to-openapi';
 import { z } from 'zod';
 import type { MediaAssetService } from './mediaAsset.service';
 import { createMediaAssetSchema } from './dto/create-mediaAsset.dto';
+import type { MediaAsset } from '../../../../node_modules/.prisma/client-media-service';
```
palani-ND, line 11: Importing a type via a relative path four levels up into `node_modules/.prisma/client-media-service` is fragile — it'll break if the Prisma generator's output path, the package manager (npm/pnpm/yarn all lay out `node_modules` differently for generated clients), or this file's own directory depth changes. If there's an existing convention elsewhere in the repo for importing generated Prisma types across service boundaries (a package alias, a re-export from `prisma.service.ts`, etc.), prefer that instead.

apps/supervisor-operations-service/src/operations/operations.controller.ts
```diff
@@ -165,6 +197,126 @@ export function createOperationsRouter(service: OperationsService) {
...
+        200: { description: 'Supervisor event', schema: envelope(supervisorEventSchema) },
+        401: { description: 'Unauthenticated', schema: apiErrorSchema },
+        403: { description: 'Caller role not permitted', schema: apiErrorSchema },
+        404: { description: 'Event not found', schema: apiErrorSchema },
+      },
+    },
+    trustGatewayIdentity,
+    requireRoles('SUPERVISOR', 'MANAGER', 'ADMIN'),
+    validate(eventIdParamsSchema, 'params'),
+    asyncHandler(async (req, res) => {
```
palani-ND: Minor consistency question: `GET /supervisor-events/:id` has no ownership check — any SUPERVISOR can fetch any event's detail by id — whereas `cancel`/`complete`/`GET|PUT .../attendance` all restrict a SUPERVISOR to events they own (403 otherwise). Is unscoped read-by-id here intentional (e.g. Supervisors need visibility into each other's schedules), or should this match the ownership scoping used everywhere else on this resource?

palani-ND (commented):
## Review summary

This PR bundles a lot: call-log CRUD, meeting/training lifecycle + attendance, two real media-service bugfixes, beneficiary role-scoping, and an auth-service id-mismatch fix. The core logic (ownership checks mirroring the existing `inventory-transactions` pattern, the `checksum`/`sizeBytes` fixes, the `sakhiId` identity-space fix) is solid and well-tested — 94+31 new tests, and the PR description shows real end-to-end verification rather than just mocks.

**Scope note:** the `BeneficiarySocioDemographics` migration/model/schema and the new `mobileNumber`/`address` decryption on `GET /beneficiaries` (in `beneficiary.controller.ts`, `beneficiary.mapper.ts`, `beneficiary.repository.ts`) aren't mentioned anywhere in the PR description or the closed issues (#93-#97), and aren't related to call-sheets/meetings/media/role-scoping. Worth confirming this was intentionally bundled in rather than an accidental merge/rebase artifact from another branch — if intentional, splitting it into its own PR would make this one easier to review and revert independently.

A few specific issues/questions left as inline comments below — mostly edge cases and one real correctness risk in the attendance upsert.

---
🤖 Review generated with [Claude Code](https://claude.com/claude-code)

apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
Dharanishwaran-ND: Fixed in 54a8612: a SUPERVISOR caller with no `projectId` now gets a clear `forbidden()` (403) before `listSakhiIdsForSupervisor` is ever called, instead of silently degrading to `/api/v1/projects//sakhis`. Added a regression test asserting the 403 and that neither the Sakhi lookup nor the repository query are hit.

apps/beneficiary-service/src/sakhi/sakhi.client.ts
Dharanishwaran-ND: Confirmed `SakhiRepository.findByProject` (auth-service) has no `take`/`skip` — genuinely unbounded today, so no live bug. Added a comment in 54a8612 documenting this as an explicit coupling: if that endpoint is ever paginated, this client must change too, or a Supervisor with Sakhis beyond the first page would silently see an under-scoped list. Didn't add speculative pagination handling for an endpoint that doesn't paginate, to avoid over-engineering against a case that doesn't exist yet — the comment makes sure it can't be missed if that changes.

apps/supervisor-operations-service/src/operations/operations.repository.ts
Dharanishwaran-ND: Confirmed as a real bug — fixed in 40cec66. `updateAttendanceSchema` now has a `.refine()` rejecting a repeated `sakhiId` within one submission's `attendance` array, so the client-side duplicate can never reach `upsertAttendance`'s once-built existing-row map at all. Added a dedicated DTO spec (`update-attendance.dto.spec.ts`) covering both the accept and reject cases.

apps/media-service/src/media/mediaAsset.controller.ts
Dharanishwaran-ND: Checked — this relative-path-into-generated-client import is actually the repo-wide convention, not a one-off: every service's own `prisma.service.ts` imports `PrismaClient` the same way (e.g. `apps/media-service/src/prisma/prisma.service.ts`), and other domain files do it directly too (`auth-service/src/geography/geography.repository.ts`, `beneficiary-service/src/beneficiary/beneficiary.service.ts`). No alternative convention (package alias, re-export) exists anywhere in the repo to switch to. Added a comment in 40cec66 explaining this so it reads as intentional rather than overlooked fragility — the path depth is fixed by this file's own directory nesting, not by Prisma/package-manager layout.

apps/supervisor-operations-service/src/operations/operations.controller.ts
Dharanishwaran-ND: Good catch — this was inconsistent and not intentional. Fixed in 40cec66: `GET /supervisor-events/:id` now uses the same ownership check as cancel/complete/attendance (a SUPERVISOR may only fetch their own events; 403 otherwise, MANAGER/ADMIN unrestricted). Added tests for the 403 case and for MANAGER/ADMIN unrestricted access.

Dharanishwaran-ND:
All 5 inline review comments addressed and replied to individually:

- **Missing `projectId` fallback** (`beneficiary.service.ts`) — fixed in 54a8612, now a clear 403 instead of a malformed path.
- **Unbounded Sakhi-list read** (`sakhi.client.ts`) — confirmed no live bug (the auth-service endpoint has no pagination today); documented the coupling in 54a8612 so it can't silently regress if that changes.
- **Attendance duplicate-`sakhiId` risk** (`operations.repository.ts`) — confirmed real, fixed in 40cec66 via schema-level rejection, with a new DTO spec.
- **Prisma type import path** (`mediaAsset.controller.ts`) — confirmed this matches the repo-wide convention (every service's `prisma.service.ts` does the same); documented as intentional in 40cec66.
- **Unscoped `GET /supervisor-events/:id`** — confirmed inconsistent, fixed in 40cec66 to match the ownership scoping used by every other single-event route, with new tests.

On the scope note about `BeneficiarySocioDemographics`/decrypted `mobileNumber`/`address` on `GET /beneficiaries`: that predates this session's changes and isn't something I added — flagging here rather than silently leaving it since you asked, but I don't have the context to say whether it should be split out. @Dharanishwaran-ND can confirm whether that was an intentional bundle or should move to its own PR.

All affected projects (`beneficiary-service`, `auth-service`, `supervisor-operations-service`, `media-service`) pass lint and tests after these fixes.

PR #98, opened 2026-08-03
fix(auth-service): Scope GET /projects and GET /projects/:id/sakhis to the caller

Dharanishwaran-ND:
## Summary

Both `GET /projects` and `GET /projects/:projectId/sakhis` previously returned unscoped data to any authenticated caller:

- **`GET /projects`** listed every active project in the system regardless of the caller's role or assigned project.
- **`GET /projects/:projectId/sakhis`** returned every Sakhi in a project, checking only project membership — never actual Supervisor ownership. Any Supervisor sharing a project with other Supervisors saw every other Supervisor's Sakhi roster, not just their own.

## Fix

- `apps/auth-service/src/projects/project.service.ts` — `list()` now takes the caller and scopes SAKHI/SUPERVISOR callers to only their own `projectId`; MANAGER/ADMIN remain unrestricted (they legitimately oversee multiple projects).
- `apps/auth-service/src/sakhis/sakhi.service.ts` — `listByProject()` now additionally filters to `supervisorId === caller.id` for SUPERVISOR callers, on top of the existing project-membership check. MANAGER/ADMIN remain unrestricted.

## Why this matters

A client that fetches the project list and picks one (e.g. "the first result") for a SAKHI/SUPERVISOR caller could previously land on a project it has no access to at all, and would then 403 on the follow-up Sakhi-list call. Separately, any Supervisor sharing a project with peers saw every peer's Sakhis in the "Assign Item" / Sakhi-picker flows — both reproduced live against a real deployed environment during manual verification.

## Test plan

- [x] `nx test auth-service` — 210/210 passing, including new coverage:
  - `project.service.spec.ts`: SUPERVISOR/SAKHI scoped to own project; MANAGER/ADMIN unrestricted; empty result when own project isn't in the active list
  - `sakhi.service.spec.ts`: SUPERVISOR scoped to own Sakhis only; MANAGER/ADMIN see every Sakhi in the project
- [x] Live verification against a running local gateway + auth-service: 5 Supervisors sharing one project each correctly see only their own Sakhis after the fix (previously all 5 saw all 10); a Supervisor's `GET /projects` correctly excludes a second project created by another party

### DIFF
--- a/apps/auth-service/src/projects/project.service.ts
+++ b/apps/auth-service/src/projects/project.service.ts
@@ -43,6 +43,19 @@ function toApiProject(p: Record<string, unknown>) {
   };
 }
 
+/**
+ * MANAGER and ADMIN are unrestricted across all project-scoping checks —
+ * checked as the absence of an elevated role, not the presence of a
+ * restrictive one (SAKHI/SUPERVISOR), since a caller can hold multiple
+ * role assignments at once (see auth.service.ts's issueTokens) and must
+ * not be scoped down just because one of their roles is restrictive.
+ * Matches the same isPrivileged() pattern in
+ * supervisor-operations-service/operations.service.ts.
+ */
+function isPrivileged(caller: CallerScope): boolean {
+  return caller.roles.includes('MANAGER') || caller.roles.includes('ADMIN');
+}
+
 /** Business logic for project/funder master data. */
 export class ProjectService {
   constructor(private readonly repository: ProjectRepository) {}
@@ -50,19 +63,18 @@ export class ProjectService {
   /**
    * A caller with a project scope on their JWT (SAKHI/SUPERVISOR — one
    * project per person per the SRS) only ever sees their own project. A
-   * caller with no single-project scope (MANAGER/ADMIN, who oversee
-   * multiple projects) is unrestricted. This prevents a SAKHI/SUPERVISOR
-   * client from picking a project it has no access to out of an unscoped
-   * list and then 403ing on a follow-up call (e.g. GET /projects/:id/sakhis).
+   * privileged caller (MANAGER/ADMIN, who oversee multiple projects) is
+   * unrestricted. This prevents a SAKHI/SUPERVISOR client from picking a
+   * project it has no access to out of an unscoped list and then 403ing
+   * on a follow-up call (e.g. GET /projects/:id/sakhis).
    */
   async list(caller: CallerScope) {
     const projects = await this.repository.findManyActiveProjects();
     const mapped = projects.map((p) => toApiProject(p as unknown as Record<string, unknown>));
-    const isScoped = caller.roles.includes('SAKHI') || caller.roles.includes('SUPERVISOR');
-    if (isScoped) {
-      return mapped.filter((p) => p.projectId === caller.projectId);
+    if (isPrivileged(caller)) {
+      return mapped;
     }
-    return mapped;
+    return mapped.filter((p) => p.projectId === caller.projectId);
   }
 
   async getById(id: string) {
--- a/apps/auth-service/src/projects/project.service.ts
+++ b/apps/auth-service/src/projects/project.service.ts
@@ -4,6 +4,12 @@ import type { CreateFunderInput } from './dto/create-funder.dto';
 import type { CreateProjectInput } from './dto/create-project.dto';
 import type { UpdateProjectInput } from './dto/update-project.dto';
 
+/** The calling principal's own scope, as carried on their JWT/trusted-identity headers. */
+export interface CallerScope {
+  readonly roles: string[];
+  readonly projectId: string | null;
+}
+
 /** Prisma unique-constraint violation code (projectCode/funderCode). */
 const PRISMA_UNIQUE_CONSTRAINT_CODE = 'P2002';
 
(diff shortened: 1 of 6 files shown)


apps/auth-service/src/projects/project.service.ts
```diff
@@ -41,9 +47,22 @@ function toApiProject(p: Record<string, unknown>) {
...
+   * caller with no single-project scope (MANAGER/ADMIN, who oversee
+   * multiple projects) is unrestricted. This prevents a SAKHI/SUPERVISOR
+   * client from picking a project it has no access to out of an unscoped
+   * list and then 403ing on a follow-up call (e.g. GET /projects/:id/sakhis).
+   */
+  async list(caller: CallerScope) {
     const projects = await this.repository.findManyActiveProjects();
-    return projects.map((p) => toApiProject(p as unknown as Record<string, unknown>));
+    const mapped = projects.map((p) => toApiProject(p as unknown as Record<string, unknown>));
+    const isScoped = caller.roles.includes('SAKHI') || caller.roles.includes('SUPERVISOR');
```
palani-ND: This (and the analogous `caller.roles.includes('SUPERVISOR')` check in `sakhi.service.ts`) scopes based on the *presence* of a restrictive role (SAKHI/SUPERVISOR) rather than the *absence* of an elevated one, unlike the `isPrivileged(caller)`-style checks used in supervisor-operations-service. If `roles` can ever contain more than one value for a caller (e.g. `['SUPERVISOR', 'MANAGER']`), this would incorrectly scope them down despite also holding an unrestricted role. Worth confirming roles are always single-valued in this system — if not, flipping this to check for the absence of MANAGER/ADMIN would be more robust.

palani-ND (commented):
## Review summary

Small, focused fix for a real authorization gap: `GET /projects` was returning every project to any authenticated caller, and `GET /projects/:projectId/sakhis` only checked project membership rather than actual Supervisor ownership (so peer Supervisors on the same project could see each other's Sakhi rosters). Both fixes follow the existing scoped-vs-unrestricted (SAKHI/SUPERVISOR vs MANAGER/ADMIN) pattern already used elsewhere in this codebase, and are backed by new tests including the empty-result edge case. Two things worth double-checking below, neither blocking.

apps/auth-service/src/sakhis/sakhi.service.ts
```diff
@@ -43,13 +45,23 @@ export class SakhiService {
...
    */
   async listByProject(projectId: string, caller: CallerScope) {
     if (caller.projectId && caller.projectId !== projectId) {
       throw forbidden('You do not have access to this project.');
     }
     const profiles = await this.repository.findByProject(projectId);
-    return profiles.map((p) => toApiSakhi(p as unknown as Record<string, unknown>));
+    const mapped = profiles.map((p) => toApiSakhi(p as unknown as Record<string, unknown>));
+    if (caller.roles.includes('SUPERVISOR')) {
+      return mapped.filter((s) => s.supervisorId === caller.id);
```
palani-ND: Worth double-checking: this assumes `sakhi_profiles.supervisor_id` already stores the Supervisor's `users.user_id` (the JWT `sub`/`caller.id`), rather than some other Supervisor-side profile PK. That's exactly the class of identity-space mismatch that `toApiSakhi()`'s `sakhiId` field had before the recent fix in #92 (`sakhi_profiles.id` vs `users.user_id`). If `supervisorId` here is populated the same way `sakhi_profiles.id` used to be (a different id space than the caller's JWT `sub`), this filter would silently return zero Sakhis for every Supervisor instead of just their own. Given the test plan's live verification did cover this ("5 Supervisors ... correctly see only their own Sakhis"), this is likely already confirmed correct — flagging mainly because it's the same bug shape that slipped through once already in this area.

apps/auth-service/src/projects/project.service.ts
Dharanishwaran-ND: Confirmed real, not hypothetical — `auth.service.ts`'s `issueTokens` explicitly says "A user can hold multiple role assignments" and the JWT's `roles` array carries the full role-code set. Fixed in 8a40ad4: both `project.service.ts` and `sakhi.service.ts` now use the same `isPrivileged(caller)` pattern already established in `supervisor-operations-service/operations.service.ts` (checks for MANAGER/ADMIN's absence, not SAKHI/SUPERVISOR's presence). Added a regression test for a caller holding `['SUPERVISOR', 'ADMIN']` (or `['SUPERVISOR', 'MANAGER']`) simultaneously, confirming they're correctly unrestricted.

apps/auth-service/src/sakhis/sakhi.service.ts
Dharanishwaran-ND: Good catch on flagging this explicitly rather than trusting the earlier fix carried over correctly. Traced it through: `sakhi_profiles.supervisor_id` is a bare untyped scalar with no relation, and the *only* real write path (`auth.service.ts`'s `updateUser`, via `PATCH /users/:id`) stores whatever `supervisorId` the caller supplies — which in every real usage in this codebase is the Supervisor's own `users.user_id` (same as their JWT `sub`), never a `supervisor_profiles.supervisor_id`. That table exists in the schema but has zero application code anywhere reading or writing it — it's dead. The existing schema comment claiming the FK relationship was itself stale/incorrect; fixed in 8a40ad4 to document the actual behavior and flag the reconciliation risk if `supervisor_profiles` is ever wired up later. No runtime change needed since the live verification already confirmed the assumption holds — this was a documentation-only fix.

Dharanishwaran-ND:
Both review comments addressed in 8a40ad4:

- **Multi-role scoping bug** (`project.service.ts` / `sakhi.service.ts`) — confirmed real: `auth.service.ts` explicitly supports a caller holding multiple role assignments simultaneously. Both files now use the `isPrivileged(caller)` pattern already established in `supervisor-operations-service/operations.service.ts` (absence of MANAGER/ADMIN, not presence of SAKHI/SUPERVISOR), with new tests for a dual-role caller.
- **`supervisorId` identity-space assumption** — traced through the only real write path (`PATCH /users/:id`) and confirmed `sakhi_profiles.supervisor_id` does store `users.user_id` in practice; `supervisor_profiles` is dead schema with no code touching it. Corrected the stale schema comment that claimed otherwise, documenting the actual contract and the reconciliation risk if that table is ever wired up.

`nx lint auth-service` and `nx test auth-service` (212/212) both pass. Re-verified live locally that the original scoping behavior (a Supervisor sees only their own Sakhis) still holds with no regression.

Issue #25, opened 2026-07-16
Create projects & funders


Dharanishwaran-ND:
| Field | Value |
|---|---|
| Category | Backend |
| Dependency | Prisma |
| Estimate (days) | 3 |
| Priority | High |
| Deliverable | Migration |
| Success Criteria | Migration successful |

PR #99, opened 2026-08-04
feat(supervisor-operations-service): Add Meeting & Training flow, revert path renames, overhaul call status

Dharanishwaran-ND:
## Summary

Three changes in this PR:

1. **New Meeting & Training flow endpoints** (`supervisor-operations-service`) — implements the 8 endpoints flagged as missing in the Supervisor App API requirements doc, none of which existed in any form before.
2. **Path renames, reverted** — 2 of the 6 endpoints originally renamed to match the requirements doc were reverted back to their pre-existing names per explicit request; the other 4 stay renamed.
3. **Call log status overhaul** — replaced the call-sheet's `callStatus` enum with the field team's actual 8-value taxonomy, and added a `responder` field to capture who picked up the call.

## 1. New Meeting & Training flow endpoints

| Method | Path | Purpose |
|---|---|---|
| GET | `/training-topics` | List topic-picker catalog |
| POST | `/training-topics` | Create a topic (ADMIN) |
| POST | `/supervisor-events/:id/reschedule` | Move a SCHEDULED event's date |
| POST | `/supervisor-events/:id/photos` | Add a photo to an event's gallery |
| POST | `/supervisor-events/:id/gatherings` | Create a Training session under an event |
| GET | `/gatherings/:gatheringId/topics` | List a gathering's topics |
| GET/PUT | `/gatherings/:gatheringId/attendance` | Session-level attendance |
| GET/PUT | `/topics/:topicId/marks` | Read/save one Pre or Post score |
| POST | `/topics/:topicId/marks/complete` | Lock a score so it can't be edited again |

### New tables (additive only — no existing table modified)

- `training_topics` — reusable topic catalog. The old schema only had a freeform `topicsJson` field per event, no structured list existed.
- `event_photos` — multi-photo gallery, additive to the existing single mandatory `photoMediaId` completion photo (untouched).
- `event_gatherings` — one row per Training session under a TRAINING event. The old schema treats Training as a single one-time event; Meetings never have gatherings (application-enforced, not a DB constraint).
- `gathering_topics` — join table: topics per gathering.
- `gathering_attendance` — session-level attendance, distinct from the existing `event_attendance` (which Meetings keep using, completely unchanged).
- `topic_marks` — per-topic Pre/Post score with an `isLocked` flag. No unlock endpoint — matches the existing one-way `SCHEDULED → COMPLETED` event-status pattern. The old attendance table's flat `preTrainingScore`/`postTrainingScore` columns are untouched and still used by Meetings/legacy event-level attendance.

Migration: `20260804045058_add_meeting_training_flow` — pure `CREATE TABLE`/`CREATE TYPE`/`ALTER TABLE`, no drops, no changes to existing columns.

## 2. Path renames — final state

| # | Path | Method | Status |
|---|---|---|---|
| 1 | `GET /call-logs/by-sakhi/:sakhiId` | GET | **Reverted to original** (was briefly `/call-logs/sakhi/:sakhiId`) |
| 2 | `GET /call-logs/by-sakhi/:sakhiId/recent` | GET | **Reverted to original** |
| 3 | `GET /inventory-items` | GET | **Reverted to original** (was briefly `/items`) |
| 4 | `POST /inventory-items` | POST | **Reverted to original** |
| 5 | `GET /inventory-transactions/by-sakhi/:sakhiId` | GET | **Reverted to original** (was briefly `/inventory-transactions/sakhi/:sakhiId`) |
| 6 | `GET`/`PUT /supervisor-events/:id/attendance` | GET/PUT | **Stays reverted to original** — real gatherings are now their own resource with their own id (see the new `/gatherings/:gatheringId/attendance` above), so keeping this alias pointed at the old event-level data would have been misleading |

Net effect: only the `/supervisor-events/:id/attendance` pair differs from where this branch started (because gatherings genuinely became a new, separate resource) — every other endpoint is back to its original pre-existing name.

## 3. Call log status overhaul

`call_logs.call_status` no longer uses the placeholder 6-value set (`CONNECTED`, `NOT_CONNECTED`, `FOLLOWUP_REQUIRED`, `BUSY`, `SWITCHED_OFF`, `WRONG_NUMBER`). It now uses the field team's actual 8-value taxonomy:

`PICKED_UP_TALKED`, `PICKED_UP_NO_ONE_TALKING`, `PICKED_UP_CUT_MIDWAY`, `CALL_BACK`, `NOT_PICKED_UP`, `RINGING`, `PHONE_OFF`, `OUT_OF_NETWORK`

A new nullable `responder` field was added (`RELATIVE`, `HUSBAND`, `SAKHI`, `PERSON_WHO_DOES_NOT_KNOW_WOMAN`) to capture who actually answered, on both create and update.

Migration `20260804081500_update_call_status_and_add_responder` remaps every existing row onto the closest new value rather than discarding data:

| Old value | New value |
|---|---|
| `CONNECTED` | `PICKED_UP_TALKED` |
| `NOT_CONNECTED` | `NOT_PICKED_UP` |
| `FOLLOWUP_REQUIRED` | `CALL_BACK` |
| `BUSY` | `RINGING` |
| `SWITCHED_OFF` | `PHONE_OFF` |
| `WRONG_NUMBER` | `OUT_OF_NETWORK` |

**Any client (mobile app, Postman collections, etc.) currently sending the old 6 values must switch to the new 8** — the old values are now rejected with a 400 validation error, not silently accepted.

Tracked separately: #100 (client migration off the old `callStatus` values).

## Test plan

- [x] `nx lint supervisor-operations-service api-gateway` — clean
- [x] `nx test supervisor-operations-service` — 144/144 passing
- [x] `nx test api-gateway` — 19/19 passing
- [x] `nx build supervisor-operations-service` — clean, including OpenAPI generation
- [x] Live verification against a running local gateway + supervisor-operations-service with real JWTs: all reverted paths return 200 on real data; renamed-then-reverted path variants correctly 404; new call log created with the new `callStatus` + `responder`; old `callStatus` value correctly rejected; existing call log confirmed remapped to its new value after migration

## Status update for reviewers (this morning → latest commit)

A plain-language summary of what happened on this branch today, in case it's useful context on top of the commit log:

- **Route renames were tried, then rolled back.** Early today, 6 endpoints were renamed to match the API requirements doc. That broke 2 endpoints the mobile app already calls under their old names (flagged in review by @palani-ND), so those 2 — plus 2 more of the 6 — were reverted back to their original names before this PR was finalized. **No action needed from anyone** — this is already fixed in the current code, and both review threads are replied-to and resolved.
- **New Meeting & Training endpoints were added.** These are net-new — nothing existing changed shape, so there's no migration risk for other clients.
- **The call log "status" values changed** (see section 3 above). This is the one change that needs action from any other client of this API: if anything outside this repo sends the old 6 `callStatus` values, it needs to switch to the new 8 before this deploys, or those calls will start failing validation. Tracked in #100.


palani-ND (commented):
## Review summary

Small, purely mechanical rename — route path strings and the gateway's proxy prefixes only, no handler/auth/DB changes. The path-collision reasoning in the description checks out: I verified in `apps/api-gateway/src/proxy/routes.ts` that the new `/items` and `/gatherings` prefixes don't collide with any existing entry, and the `sakhi/:sakhiId` segment choice avoids colliding with the existing single-segment `:callLogId`/`:id` routes in the same router (different segment counts, so Express can't confuse them) — matches the PR's own stated rationale.

**One real cross-repo concern, not inline-anchorable to this diff alone:** two of these six renamed paths are already hardcoded as Retrofit endpoints in the `supervisor-mobile-app` repo's `InventoryApi.kt` (added in that repo's PR #25, which I reviewed separately):
- `@GET("inventory-items")` → this PR renames the backend route to `/items`
- `@GET("inventory-transactions/by-sakhi/{sakhiId}")` → this PR renames it to `/inventory-transactions/sakhi/{sakhiId}` (drops `by-`)

If this PR merges and deploys before the mobile app's Retrofit interface is updated to match, both of those calls start 404ing against the real backend — `getInventoryItems()` (used by Assign Item's item catalog) and `getTransactionsBySakhi()` (used by the Assign Item detail screen's transaction history). Worth confirming the mobile-side update is sequenced together with this deploy, or landing a redirect/alias for a transition window, rather than assuming both repos ship in lockstep. The other four renamed paths (`/gatherings/...`, `/call-logs/sakhi/...`) aren't yet called from the mobile app per that PR's own description (call-logs and attendance are still local-only there), so those are safe to rename now.

apps/supervisor-operations-service/src/operations/operations.controller.ts
```diff
@@ -319,7 +319,7 @@ export function createOperationsRouter(service: OperationsService) {
   );
 
   doc.get(
-    '/inventory-items',
+    '/items',
```
palani-ND, line 322: `GET /items` — this is the exact path `supervisor-mobile-app`'s `InventoryApi.getInventoryItems()` (`@GET("inventory-items")`) currently calls under its old name. See the top-level review comment for the cross-repo sequencing concern.

apps/supervisor-operations-service/src/operations/operations.controller.ts
```diff
@@ -381,7 +381,7 @@ export function createOperationsRouter(service: OperationsService) {
   );
 
   doc.get(
-    '/inventory-transactions/by-sakhi/:sakhiId',
+    '/inventory-transactions/sakhi/:sakhiId',
```
palani-ND, line 384: `GET /inventory-transactions/sakhi/:sakhiId` — this is the exact path `supervisor-mobile-app`'s `InventoryApi.getTransactionsBySakhi()` (`@GET("inventory-transactions/by-sakhi/{sakhiId}")`) currently calls under its old name (with `by-`). See the top-level review comment for the cross-repo sequencing concern.

apps/supervisor-operations-service/src/operations/operations.controller.ts
Dharanishwaran-ND: Resolved — commit 1de7882 (landed before the latest push) already reverted this rename back to `GET /inventory-items`, its original pre-rename path, specifically so it keeps matching `supervisor-mobile-app`'s `InventoryApi.getInventoryItems()` (`@GET("inventory-items")`). Current code: apps/supervisor-operations-service/src/operations/operations.controller.ts:400.

apps/supervisor-operations-service/src/operations/operations.controller.ts
Dharanishwaran-ND: Resolved — commit 1de7882 (landed before the latest push) already reverted this rename back to `GET /inventory-transactions/by-sakhi/:sakhiId`, its original pre-rename path, specifically so it keeps matching `supervisor-mobile-app`'s `InventoryApi.getTransactionsBySakhi()` (`@GET("inventory-transactions/by-sakhi/{sakhiId}")`). Current code: apps/supervisor-operations-service/src/operations/operations.controller.ts:462.

Dharanishwaran-ND:
## Review follow-up

palani-ND's review flagged two route renames (`/items`, `/inventory-transactions/sakhi/:sakhiId`) that would 404 against `supervisor-mobile-app`'s existing Retrofit calls.

Both are already resolved: commit 1de7882 (pushed before this review, but landed as part of this same PR) reverted those two renames — along with `/call-logs/by-sakhi/:sakhiId(/recent)` — back to their original pre-rename paths, specifically so they keep matching the mobile app's `InventoryApi.kt` calls. No further code change was needed; replied on both inline threads with the exact current line numbers and resolved them.

PR #101, opened 2026-08-04
fix(supervisor-operations-service): Annotate gt parameter type in upsertTopicMark

Dharanishwaran-ND:
## Summary

- `upsertTopicMark`'s `gatheringTopics.some((gt) => gt.topicId === topicId)` had no type annotation on `gt`, which failed `noImplicitAny` (`TS7006`) on the dev box's stricter build config — this was blocking dev deploys of `supervisor-operations-service` after PR #99 merged.
- `gt` now takes an explicit type derived from `findGatheringTopics`'s own return type (`(typeof gatheringTopics)[number]`), so it can't drift from the repository's actual return shape.

This fix was originally pushed to the `feature/call-sheet-and-supervisor-events` branch, but landed *after* PR #99 was already merged, so it never made it into `develop`. Cherry-picked onto a fresh branch here.

## Test plan

- [x] `nx build supervisor-operations-service` — clean, no TS errors
- [x] `nx lint supervisor-operations-service` — clean
- [x] `nx test supervisor-operations-service` — 144/144 passing

### DIFF
--- a/apps/supervisor-operations-service/src/operations/operations.service.ts
+++ b/apps/supervisor-operations-service/src/operations/operations.service.ts
@@ -530,7 +530,9 @@ export class OperationsService {
     }
 
     const gatheringTopics = await this.repository.findGatheringTopics(dto.gatheringId);
-    const belongsToGathering = gatheringTopics.some((gt) => gt.topicId === topicId);
+    const belongsToGathering = gatheringTopics.some(
+      (gt: (typeof gatheringTopics)[number]) => gt.topicId === topicId,
+    );
     if (!belongsToGathering) {
       throw unprocessable('topicId: this topic is not part of the referenced gathering.');
     }


palani-ND (commented):
## Review summary

Tiny, low-risk fix — a cherry-picked `noImplicitAny` compile fix that missed `develop` due to merge-order timing with #99. The annotation itself is good practice: deriving `gt`'s type from `(typeof gatheringTopics)[number]` instead of hand-writing a duplicate interface means it can't drift from `findGatheringTopics`'s actual return shape, exactly as the description says.

One thing that doesn't quite add up and might be worth a quick look (not blocking, just curious): `findGatheringTopics` (`operations.repository.ts`) returns `this.prisma.gatheringTopic.findMany({ include: { topic: true } })` — a concretely-typed Prisma result, not `any[]`. And this repo's own checked-in `tsconfig.base.json` already has `strict: true` and `noImplicitAny: true` — the same config `nx build` uses. Under a normal, correctly-configured TS setup, `Array.prototype.some((gt) => ...)` infers `gt`'s type from the array automatically regardless of how complex the element type is (Prisma intersection types included) — there's usually no way to get an implicit-`any` TS7006 out of a `.some()` callback on an already-typed array. That makes me wonder whether "the dev box's stricter build config" is actually a TypeScript version/lockfile drift between that machine and CI, rather than a config difference — if so, it's worth finding and fixing at the version level, since the same latent gap could be sitting on other `.some`/`.map`/`.filter` calls over repository results elsewhere in this service that just haven't been hit on that stricter box yet. If it *is* a genuine, reproducible `tsc` error under the checked-in config, that'd be surprising and worth a one-line note on how it reproduces, since it doesn't match how TS inference normally works here.

apps/supervisor-operations-service/src/operations/operations.service.ts
```diff
@@ -530,7 +530,9 @@ export class OperationsService {
     }
 
     const gatheringTopics = await this.repository.findGatheringTopics(dto.gatheringId);
-    const belongsToGathering = gatheringTopics.some((gt) => gt.topicId === topicId);
+    const belongsToGathering = gatheringTopics.some(
```
palani-ND, line 533: Confirmed `findGatheringTopics` returns a concretely-typed Prisma array (not `any[]`), so this annotation correctly mirrors the real shape. See the top-level review comment for a question about why plain inference didn't already cover this under the checked-in `strict`/`noImplicitAny` tsconfig — worth a quick sanity check in case it points to a TS version mismatch elsewhere rather than something specific to this line.

Dharanishwaran-ND: Good catch — investigated this properly rather than assuming. Reverted just the annotation locally (`gt` back to plain inference) and ran both `nx build supervisor-operations-service` and `tsc --noEmit` directly against this repo's checked-in tsconfig: **neither reproduces TS7006.** `findGatheringTopics`'s Prisma-typed array infers `gt`'s type through `.some()` exactly as you'd expect — confirms plain inference does cover this here.

Checked the version-drift angle too: local `node_modules/typescript` is 5.9.3, and `package-lock.json` itself pins `typescript` to resolve to 5.9.3 — so it's not a TypeScript/lockfile mismatch between this machine and what CI would install. The one real difference I can find is Node itself: this machine runs Node 24.18.0, while `.nvmrc`/CI pin 20.11.0. I can't fully rule that out without actually running `tsc` under Node 20 here, but I don't have a way to reproduce the original TS7006 on this machine as configured.

Net: I can't confirm this was a genuine, reproducible compile error under the checked-in config — it may well have been specific to whatever toolchain state the original "dev box" was in at the time. That said, the annotation itself is still worth keeping independent of whether it fixes a real error: deriving `gt`'s type from `(typeof gatheringTopics)[number]` ties it to `findGatheringTopics`'s actual return shape, so it can't silently drift if that method's Prisma `include` ever changes — a defensive improvement either way, not just a workaround.

Given I can't reproduce the original failure, I won't go hunting for "the same latent gap" elsewhere in the service on the theory that it's a real, systemic issue — there's no confirmed bug to chase yet. Flagging this as something to watch for if it resurfaces on an actual CI run, rather than something to fix further right now.

Dharanishwaran-ND:
## Review follow-up

@palani-ND's question about whether TS7006 actually reproduces under this repo's checked-in config was worth investigating properly rather than dismissing.

**Findings:** reverted the annotation locally and ran both `nx build supervisor-operations-service` and `tsc --noEmit` — neither reproduces the error. `typescript` resolves to 5.9.3 both locally and per `package-lock.json`, so it's not a TS/lockfile mismatch. The one real difference from CI is Node itself (this machine: 24.18.0; `.nvmrc`/CI: 20.11.0) — can't fully rule that out without testing under Node 20, but there's no confirmed reproduction here.

**Conclusion:** the original TS7006 likely wasn't a genuine, reproducible error under this repo's checked-in config — probably specific to whatever toolchain state the original dev box was in. The annotation itself stays in the PR regardless: deriving `gt`'s type from `(typeof gatheringTopics)[number]` is a real improvement (ties it to `findGatheringTopics`'s actual shape, can't silently drift), just not something confirmed to be fixing an active build failure. Replied on the inline thread with the full investigation and resolved it — not chasing "the same latent gap elsewhere" since there's no confirmed bug to chase.

Dharanishwaran-ND:
Tracking issue: #107

PR #102, opened 2026-08-04
feat(rules-service): Add non-admin rule version lookup and seed script

Dharanishwaran-ND:
## Summary

- `GET /rules/versions/:versionId` — new, open to any authenticated role (unlike the existing `GET /admin/rules/:setId`, which is ADMIN-only and looks up by `ruleSetId`, not `versionId`). Needed so other services (e.g. the upcoming visit-schedules bulk-upload endpoint) can verify a `generatedByRuleVersionId` is real and `PUBLISHED` without needing rules-admin access themselves.
- `apps/rules-service/prisma/seed.ts` — new. This service had no seeding mechanism at all. Seeds one `SCHEDULE` `RuleSet` + a `v1-hardcoded` `PUBLISHED` `RuleVersion`, on fixed UUIDs (not `@default(uuid())`), so every environment's seed run produces the same id — required since a downstream mobile client hardcodes this UUID and it must match across dev/SIT/UAT.
- Deliberately narrow response shape (`id`/`ruleSetId`/`status` only) — omits `rulesJson`/`checksum`/audit columns, which a non-admin caller has no reason to see.

## Test plan

- [x] `nx lint rules-service` — clean
- [x] `nx test rules-service` — 14/14 passing (2 new tests for `getById`)
- [x] `nx build rules-service` — clean, builds independently of any other in-flight change
- [x] Live verification against a running local gateway + rules-service with a real SUPERVISOR JWT (not just ADMIN): `GET /rules/versions/22222222-2222-4222-8222-222222222222` → 200, `{status: "PUBLISHED"}`

## Notes

The seeded UUID (`22222222-2222-4222-8222-222222222222`) is what the mobile team needs to hardcode as `generatedByRuleVersionId`.

### DIFF
--- a/apps/rules-service/prisma/seed.ts
+++ b/apps/rules-service/prisma/seed.ts
@@ -1,0 +1,70 @@
+import { createHash } from 'node:crypto';
+import { PrismaClient } from '../../../node_modules/.prisma/client-rules-service';
+
+const prisma = new PrismaClient();
+
+// Fixed, hardcoded UUIDs (not @default(uuid()) at insert time) — the Sakhi
+// mobile app hardcodes SCHEDULE_RULE_VERSION_ID in HardcodedRuleSource and
+// must match exactly across dev/SIT/UAT, so every environment's seed run has
+// to produce the same id rather than a fresh one per run.
+const SCHEDULE_RULE_SET_ID = '11111111-1111-4111-8111-111111111111';
+const SCHEDULE_RULE_VERSION_ID = '22222222-2222-4222-8222-222222222222';
+const SCHEDULE_RULE_VERSION_NO = 'v1-hardcoded';
+
+/**
+ * Seeds the SCHEDULE rule set + its v1-hardcoded published version, so
+ * VisitSchedule.generatedByRuleVersionId (NOT NULL, no default) has a real
+ * row to reference for M2 — actual scheduling rules live in the Kotlin app's
+ * HardcodedRuleSource today and move to GoRules packages in M3 (CR-032).
+ */
+async function seedScheduleRuleVersion(): Promise<void> {
+  // update: {} is deliberate on both upserts — this only ever creates the row
+  // on a fresh environment. Once it exists, re-running the seed must NOT
+  // touch it: dev/SIT/UAT may already have real data (schedules, submissions)
+  // referencing this exact row, and silently rewriting rulesJson/checksum/
+  // effectiveFrom out from under them would be a live-data mutation disguised
+  // as a seed. A real change to this rule version's content is a new
+  // versionNo (a new row), not an edit of this one — matching the
+  // create/immutable-then-supersede pattern used elsewhere in rules-service
+  // (see ruleVersion.repository.ts's publishNewVersion).
+  await prisma.ruleSet.upsert({
+    where: { id: SCHEDULE_RULE_SET_ID },
+    create: {
+      id: SCHEDULE_RULE_SET_ID,
+      ruleCategory: 'SCHEDULE',
+      ruleSetName: 'Arogya Sakhi Visit Scheduling',
+      status: 'ACTIVE',
+    },
+    update: {},
+  });
+
+  const rulesJson = {
+    note: 'Rules implemented in sakhi-mobile-app HardcodedRuleSource. Replaced by GoRules packages in M3 (CR-032).',
+  };
+
+  await prisma.ruleVersion.upsert({
+    where: { id: SCHEDULE_RULE_VERSION_ID },
+    create: {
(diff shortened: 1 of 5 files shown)


Dharanishwaran-ND:
Tracking issue: #108

apps/rules-service/prisma/seed.ts
```diff
@@ -0,0 +1,61 @@
...
+async function seedScheduleRuleVersion(): Promise<void> {
+  await prisma.ruleSet.upsert({
+    where: { id: SCHEDULE_RULE_SET_ID },
+    create: {
+      id: SCHEDULE_RULE_SET_ID,
+      ruleCategory: 'SCHEDULE',
+      ruleSetName: 'Arogya Sakhi Visit Scheduling',
+      status: 'ACTIVE',
+    },
+    update: {},
```
palani-ND, line 38: `update: {}` on both upserts means a re-run of this script never syncs an *existing* row to match edits made to this file later — only a brand-new environment (or one where the row doesn't exist yet) would pick up a change to `rulesJson`/`checksum`/`effectiveFrom` etc. If that's intentional (protecting whatever's already live from being silently altered by a re-seed), worth a one-line comment saying so; otherwise it's a bit of a trap the next time this file needs a real edit and someone expects `nx run seed` to converge dev/SIT/UAT to the new content.

apps/visit-form-service/prisma/seed-data/mother-registration.json
```diff
@@ -680,6 +680,11 @@
       "required": true,
       "input_type": "number",
       "numericRange": { "max": 14, "min": 0 },
+      "visibleWhen": {
```
palani-ND: This `visibleWhen` addition (here and the 3 sibling occurrences below, gating `para`/`living_children`/`abortions`/`still_births` on `gravida >= 2`) isn't mentioned in this PR's description, and it's not part of #105's Gravida-fix content either (different lines, different change) — it looks like it belongs with #105's Gravida cross-total fix (hiding these fields exactly for the Gravida=1 case #105 makes valid) rather than this rules-service PR. Worth confirming this is meant to be here, or whether it should move to #105 where the rest of the Gravida-related changes live — as a heads up, this also doesn't fully close the `gravida = 0` edge case I flagged on #105 (that boundary is >= 2 here, so it also hides these fields for `gravida = 0`, but only if the mobile client actually respects `visibleWhen` and skips sending values for hidden fields — the server-side cross-check itself doesn't know about this UI hint).

palani-ND (commented):
## Review summary

Housekeeping note first: this diff (against `develop`) also includes the `create-beneficiary.dto.ts`/`.spec.ts` Gravida fix and the `mother-registration.json` `husbands_name` addition — identical content to PR #105 (same before/after hashes). That's branch stacking (this branch sits on top of #105's commit, which hasn't merged to `develop` yet), not something added by this PR's own work — I already reviewed that content on #105, so I'm not repeating it here, and it should disappear from this diff once #105 merges first.

The actual rules-service work is clean and correctly scoped to what the summary describes: `GET /rules/versions/:versionId` is properly narrowed to `id`/`ruleSetId`/`status` (omitting `rulesJson`/`checksum`/audit columns a non-admin caller has no reason to see, with a test asserting exactly that), open to any authenticated role as intended, and reuses the existing `/rules` gateway prefix without needing a new route entry. `getById`'s 404 path and the summary-projection are both covered by tests.

Two smaller notes below — one on the new seed script, one flagging what looks like an out-of-scope/undocumented addition that isn't part of #105's changes either.

apps/rules-service/prisma/seed.ts
Dharanishwaran-ND: Valid — added a comment explaining this is deliberate: this only ever creates the row on a fresh environment, and once it exists a re-seed must NOT touch it, since dev/SIT/UAT may already have real data referencing this exact row. A real content change to a rule version is a new versionNo (a new row), not an edit — matches the create/immutable-then-supersede pattern this service already uses for publishNewVersion.

apps/visit-form-service/prisma/seed-data/mother-registration.json
Dharanishwaran-ND: Confirmed and fixed — this was accidental leakage from an earlier, separate task (adding visibleWhen gating to MOTHER_REGISTRATION fields 46-56 on gravida>=2), never properly committed anywhere, and it ended up in this commit by mistake during branch surgery. It doesn't belong here or on #105 (verified #105's branch doesn't have it either). Stripped it out of this PR entirely and rebased onto current develop to also drop the unrelated CI/env-example diff noise you flagged — this PR is now scoped to just the rules-service work. The visibleWhen work will come back as its own dedicated PR.

PR #103, opened 2026-08-04
feat(visit-form-service): Add POST /visit-schedules/bulk

Dharanishwaran-ND:
## ⚠️ Depends on #102

This PR is stacked on `feat/rules-service-version-lookup` (#102) — it calls the new `GET /rules/versions/:versionId` endpoint added there. Base branch is set to that PR's branch, not `develop`, so the diff shown here is scoped to just this PR's own commit; merge #102 first, then retarget/merge this one into `develop`.

## Summary

Implements `POST /visit-schedules/bulk` per the Sakhi App Backend Requirements doc (§3). The Sakhi app generates a beneficiary's full visit schedule on-device, offline, at enrollment (SRS FR-S-2.2/2.2A) — this endpoint is the upload path; the server never generates a schedule itself.

- **Idempotent** by `localScheduleUuid` (new `@unique` column + migration) — a retried upload returns the existing row unchanged rather than creating a duplicate.
- **Transactional per batch** — all-or-nothing; a partially-written schedule would leave gaps in a woman's care plan.
- **Anchor resolution** for `*_HR` visit types — resolves `anchorVisitLocalUuid` against already-stored rows AND against sibling rows in the same batch (two-pass insert, since a sibling's real id isn't known until it's inserted). Rejects (422) rather than silently storing null if unresolvable.
- **Conflict detection** (409) — same `(beneficiaryId, visitCode, generatedByRuleVersionId)` under a different `localScheduleUuid` is a real conflict, not an update; fixing a schedule goes through supersede (P2, not built here), never a re-upload.
- **Date-only enforcement** — `scheduledDate`/`windowStartDate`/`windowEndDate` reject a full ISO datetime (400), since a timezone-bearing timestamp could shift a real visit to the wrong day.
- **413** on batches over 100 rows (new `ErrorCode.PAYLOAD_TOO_LARGE`, added to the shared error envelope — no 413 mapping existed before this).

## New cross-service clients

- `beneficiary.client.ts` — existence check via `GET /beneficiaries/:id`, through the gateway with the caller's own token forwarded. **Note:** that endpoint has no ownership scoping today (any SAKHI/SUPERVISOR/MANAGER can fetch any beneficiary by id) — this is existence-only, not an ownership guarantee. Flagged as a follow-up, not fixed here.
- `ruleVersion.client.ts` — calls #102's new endpoint to verify `generatedByRuleVersionId` is real and `PUBLISHED`.

## Gateway

Added `/visit-schedules` → visit-form-service.

## Test plan

- [x] `nx lint visit-form-service api-gateway service-commons` — clean
- [x] `nx test visit-form-service` — 86/86 passing
- [x] `nx test service-commons` — clean (413 mapping didn't break anything)
- [x] `nx build visit-form-service` — clean
- [x] Live verification against a running local gateway + visit-form-service + rules-service with a real SUPERVISOR JWT:
  - Unknown `beneficiaryId` → 404 ✓
  - Full ISO datetime for `scheduledDate` → 400 with a clear field error ✓
  - 101-row batch → 413 `PAYLOAD_TOO_LARGE` ✓
  - Server boots and OpenAPI doc generates without crashing (see note below)

## Bug found and fixed during live verification

The first version of `dateOnlySchema` used a custom `z.transform()`. This crashed the **entire service at startup** with `UnknownZodTypeError` — `zod-to-openapi` cannot introspect a custom transform, and grepping the repo confirmed no existing DTO uses one for exactly this reason. Fixed by keeping the date fields as validated strings through the Zod schema and converting to `Date` only in the service layer, after validation (`toDateOnly()`).

## Not built (per the doc's own "do NOT build for M2" list)

`GET /visit-schedules` (read), `PATCH /visit-schedules/supersede`.


Dharanishwaran-ND:
Tracking issue: #109

apps/visit-form-service/src/visit-schedules/visitSchedule.controller.ts
```diff
@@ -0,0 +1,105 @@
...
+        }),
+        413: errorResponse(413, { message: 'schedules cannot exceed 100 rows (got 101).' }),
+        422: errorResponse(422, {
+          message: 'anchorVisitLocalUuid "..." is not in this batch and not already stored.',
+        }),
+        500: errorResponse(500),
+      },
+    },
+    trustGatewayIdentity,
+    requireRoles('SAKHI', 'SUPERVISOR'),
```
palani-ND, line 93: This endpoint has no ownership check at all — `beneficiaryExists` (existence-only, as its own doc comment says) is the *only* gate between a `SAKHI`/`SUPERVISOR` caller and writing a full visit schedule for `dto.beneficiaryId`. The PR description flags the read-side limitation (`GET /beneficiaries/:id` has no ownership scoping) as a tracked follow-up, but that framing undersells what it means for *this* endpoint specifically: since there's no check here beyond "does this beneficiary exist," any authenticated Sakhi can currently submit (and, via the idempotency/conflict logic, effectively lock in) a visit schedule for a beneficiary that isn't hers. That's a stronger gap than a read-only existence check — every other write endpoint in this codebase I've seen (call-logs, inventory-transactions, supervisor-events) enforces an explicit "is this beneficiary/Sakhi actually assigned to the caller" check before allowing a write, even where the read-side dependency has the same existence-only limitation. Worth confirming whether this is meant to be closed before this ships, or whether it's genuinely deferred by design for M2.

apps/visit-form-service/src/beneficiaries/beneficiary.client.ts
```diff
@@ -0,0 +1,33 @@
...
+export async function beneficiaryExists(
+  beneficiaryId: string,
+  authorizationHeader: string,
+): Promise<boolean> {
+  const url = `${GATEWAY_BASE_URL}/api/v1/beneficiaries/${beneficiaryId}`;
+  const res = await fetch(url, { headers: { Authorization: authorizationHeader } });
+
+  if (res.status === 404) return false;
+  if (!res.ok) {
+    throw notFound('Unable to verify the beneficiary — beneficiary lookup failed.');
```
palani-ND: This throws `notFound()` (404) for *any* non-ok, non-404 response from beneficiary-service — including a genuine 500/502 from that service being down or erroring internally. That mislabels a backend outage as "beneficiary case not found" to the Sakhi uploading a schedule, which is misleading (she may know the beneficiary was just enrolled) and denies the client the ability to distinguish "stop, this beneficiary doesn't exist" from "retry later, something's down." `beneficiary-service`'s own `sakhi.client.ts` (`listSakhiIdsForSupervisor`) already establishes the right pattern for this exact situation — `badGateway()` (502) for a non-ok/unreachable downstream call, reserving 404 for the actual not-found case. Worth aligning with that.

apps/visit-form-service/src/rules/ruleVersion.client.ts
```diff
@@ -0,0 +1,35 @@
...
+export async function findRuleVersion(
+  ruleVersionId: string,
+  authorizationHeader: string,
+): Promise<RuleVersionSummary | null> {
+  const url = `${GATEWAY_BASE_URL}/api/v1/rules/versions/${ruleVersionId}`;
+  const res = await fetch(url, { headers: { Authorization: authorizationHeader } });
+
+  if (res.status === 404) return null;
+  if (!res.ok) {
+    throw new Error('Unable to verify the rule version — rules-service lookup failed.');
```
palani-ND: Similar to `beneficiary.client.ts`: this throws a plain `new Error(...)` (not an `HttpError`) for a non-ok response from rules-service, so it'll surface as an opaque, unlabeled 500 rather than the `badGateway()` (502) that already exists in `@armman/service-commons` and is used for this exact "downstream call failed" case elsewhere (e.g. `beneficiary-service`'s `sakhi.client.ts`). Worth using that here too, for both the correct status code and consistency with the sibling client above.

palani-ND (commented):
## Review summary

Solid implementation of FR-S-2.2/2.2A's upload path, and correctly stops at the doc's own "do NOT build for M2" boundary (no read endpoint, no supersede) rather than scope-creeping into P2 work. The transactional two-pass anchor resolution is sound — resolving against already-stored rows first, then same-batch siblings via a second UPDATE pass once real ids exist, with a hard reject (422) rather than a silent null on an unresolvable anchor. Idempotency-by-`localScheduleUuid` and the separate SCHEDULE_CONFLICT check (different uuid, same beneficiary+visitCode+ruleVersion) are both correctly distinguished from each other and covered by dedicated tests. The `dateOnlySchema`-not-a-`.transform()` workaround and the new 413/`PAYLOAD_TOO_LARGE` wiring through `service-commons` are clean, minimal, and don't leak into unrelated status-code paths.

One gap that goes a bit further than what the PR itself flags, plus two smaller inconsistencies in the new cross-service clients.

apps/visit-form-service/src/visit-schedules/visitSchedule.controller.ts
Dharanishwaran-ND: Confirmed and fixed — this was a real, more serious gap than the PR description let on, agreed. Added a proper ownership check in visitSchedule.service.ts: a SAKHI caller must own the beneficiary directly (beneficiary.sakhiId === caller.id, 403 otherwise); a SUPERVISOR caller must have that beneficiary's Sakhi actually assigned to them, checked via a new sakhi.client.ts that mirrors supervisor-operations-service's SakhiClient/listCallLogsBySakhi pattern exactly (visit-form-service doesn't own sakhi_profiles either, so this can only be an API call). MANAGER/ADMIN stay unrestricted, matching every other service's convention. findBeneficiaryById() (renamed from beneficiaryExists) now returns sakhiId so the check has something to compare against. Covered by 6 new tests (SAKHI-owns/doesn't-own, SUPERVISOR-assigned/not-assigned, unresolvable-Sakhi, MANAGER/ADMIN-bypass).

apps/visit-form-service/src/beneficiaries/beneficiary.client.ts
Dharanishwaran-ND: Fixed — now uses badGateway() (502) for both the unreachable-fetch case and any non-ok/non-404 response, matching sakhi.client.ts's convention exactly. Also renamed beneficiaryExists to findBeneficiaryById since it now needs to return the beneficiary's sakhiId for the ownership check (see the other comment on this PR).

apps/visit-form-service/src/rules/ruleVersion.client.ts
Dharanishwaran-ND: Fixed — now uses badGateway() (502) for both the unreachable-fetch case and any non-ok/non-404 response, matching sakhi.client.ts's convention, same fix as the sibling beneficiary.client.ts comment.

PR #104, opened 2026-08-04
fix(visit-form-service): Make POST /visits idempotent

Dharanishwaran-ND:
## Summary

`VisitInstanceService.create` called `prisma.visitInstance.create` directly with no existence check, despite `localVisitUuid` being `@unique`. A retried offline visit upload — normal on rural connections — hit `P2002` and surfaced as an unhandled 500 instead of returning the original row.

Flagged as a live P1 bug in the Sakhi App Backend Requirements doc (§6): "Not required for M2, but it is a live bug in shipped code and it is ~2 hours."

## Fix

- `visitInstance.repository.ts` — added `findByLocalVisitUuid` and `findScheduleById`.
- `visitInstance.service.ts` — `create()` now looks up by `localVisitUuid` first and returns the existing row unchanged on a replay, mirroring the same pattern already used by `form.service.ts`'s `createSubmission` and `beneficiary.service.ts`'s enroll flow. Also validates `scheduleId` resolves before insert, returning a typed 409 instead of a raw FK-violation error.

## Test plan

- [x] `nx test visit-form-service` — 59/59 passing (3 new/updated cases: idempotent replay returns the original row; unknown `scheduleId` → typed 409; normal create path still works)
- [x] `nx build visit-form-service` — clean, independent of any other in-flight PR

### DIFF
--- a/apps/visit-form-service/src/visits/visitInstance.service.ts
+++ b/apps/visit-form-service/src/visits/visitInstance.service.ts
@@ -1,4 +1,4 @@
-import { conflict } from '@armman/service-commons';
+import { unprocessable } from '@armman/service-commons';
 import type { VisitInstanceRepository } from './visitInstance.repository';
 import type { CreateVisitInstanceInput } from './dto/create-visitInstance.dto';
 
@@ -23,7 +23,13 @@ export class VisitInstanceService {
 
     const schedule = await this.repository.findScheduleById(dto.scheduleId);
     if (!schedule) {
-      throw conflict('scheduleId does not reference an existing visit schedule.');
+      // 422, not 409 — this is "the referenced scheduleId doesn't exist"
+      // (a bad reference), the same class of error supervisor-operations-
+      // service's createCallLog uses unprocessable() for on an unknown
+      // sakhiId — not a state conflict, which conflict()/409 is reserved for
+      // elsewhere in this codebase (form.service.ts's concurrent-DRAFT/
+      // wrong-status cases).
+      throw unprocessable('scheduleId does not reference an existing visit schedule.');
     }
 
     return this.repository.create(dto);
--- a/apps/visit-form-service/src/visits/visitInstance.service.ts
+++ b/apps/visit-form-service/src/visits/visitInstance.service.ts
@@ -1,3 +1,4 @@
+import { conflict } from '@armman/service-commons';
 import type { VisitInstanceRepository } from './visitInstance.repository';
 import type { CreateVisitInstanceInput } from './dto/create-visitInstance.dto';
 
@@ -9,7 +10,22 @@ export class VisitInstanceService {
     return this.repository.findMany();
   }
 
-  create(dto: CreateVisitInstanceInput) {
+  /**
+   * Idempotent by localVisitUuid (@unique) — mirrors form.service.ts's
+   * createSubmission and beneficiary.service.ts's enroll pattern. Without
+   * this, a retried offline visit upload hit P2002 on the unique constraint
+   * and surfaced as an unhandled 500 rather than returning the original row;
+   * on rural connections a retry is the norm, not the exception.
+   */
+  async create(dto: CreateVisitInstanceInput) {
+    const existing = await this.repository.findByLocalVisitUuid(dto.localVisitUuid);
+    if (existing) return existing;
+
+    const schedule = await this.repository.findScheduleById(dto.scheduleId);
+    if (!schedule) {
+      throw conflict('scheduleId does not reference an existing visit schedule.');
+    }
+
     return this.repository.create(dto);
   }
 }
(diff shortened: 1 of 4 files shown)


Dharanishwaran-ND:
Tracking issue: #110

apps/visit-form-service/src/visits/visitInstance.service.ts
```diff
@@ -9,7 +10,22 @@ export class VisitInstanceService {
...
+   * and surfaced as an unhandled 500 rather than returning the original row;
+   * on rural connections a retry is the norm, not the exception.
+   */
+  async create(dto: CreateVisitInstanceInput) {
+    const existing = await this.repository.findByLocalVisitUuid(dto.localVisitUuid);
+    if (existing) return existing;
+
+    const schedule = await this.repository.findScheduleById(dto.scheduleId);
+    if (!schedule) {
+      throw conflict('scheduleId does not reference an existing visit schedule.');
```
palani-ND: Status-code convention check: this file's own sibling, `form.service.ts`, uses `notFound()` (404) for "referenced resource doesn't exist" (`No published form version found...`, `Unknown form code...`) and reserves `conflict()` (409) for genuine state conflicts (concurrent creation, wrong DRAFT status). `supervisor-operations-service` uses `unprocessable()` (422) for the equivalent "referenced Sakhi not found" case in `createCallLog`. A non-existent `scheduleId` reads more like the same "referenced resource doesn't exist" case those two conventions cover, not a state conflict — `notFound()` (or `unprocessable()`, matching the other service) seems like the better fit than `conflict()` here, both for internal consistency and so a client doesn't have to special-case this one 409 differently from every other "bad reference" error in the codebase.

apps/visit-form-service/src/visits/visitInstance.service.ts
```diff
@@ -9,7 +10,22 @@ export class VisitInstanceService {
...
 
-  create(dto: CreateVisitInstanceInput) {
+  /**
+   * Idempotent by localVisitUuid (@unique) — mirrors form.service.ts's
+   * createSubmission and beneficiary.service.ts's enroll pattern. Without
+   * this, a retried offline visit upload hit P2002 on the unique constraint
+   * and surfaced as an unhandled 500 rather than returning the original row;
+   * on rural connections a retry is the norm, not the exception.
+   */
+  async create(dto: CreateVisitInstanceInput) {
```
palani-ND, line 20: This mirrors the existing check-then-insert idempotency pattern in `beneficiary.service.ts`'s `create()` (same `findByLocalCaseUuid`-then-`create` shape, no `P2002` catch-and-recover), so it's consistent with the rest of the codebase rather than a new gap. Worth flagging anyway since it's exactly the scenario motivating this fix: two *concurrent* retries with the same `localVisitUuid` (not sequential — e.g. a client that times out and fires a second request while the first is still mid-flight, plausible on the flaky rural connections this PR is written for) can both pass the `findByLocalVisitUuid` null-check before either has inserted, and the second `repository.create` still hits the same `P2002` this PR is meant to eliminate. Given it matches an established repo-wide pattern, this is more a "worth confirming acceptable" note than a blocker for this PR specifically.

palani-ND (commented):
## Review summary

Correctly scoped to the P1 bug the requirements doc §6 flagged, and nothing more — no unrelated cleanup, matches the `findByLocalCaseUuid`-style idempotency pattern already established in `beneficiary.service.ts`/`form.service.ts`, and the three new test cases (replay-returns-existing, unresolved-scheduleId-409, normal-create-still-works) cover the stated behavior. Two things worth a look before merge, both about matching existing conventions rather than new functionality.

One more gap: `visitInstance.controller.ts`'s `POST /visits` OpenAPI `responses` block (`400/401/403/500` only) wasn't touched by this PR, so the new `409` this adds isn't documented — the Swagger spec is now out of sync with actual behavior for this endpoint's error paths.

apps/visit-form-service/src/visits/visitInstance.service.ts
Dharanishwaran-ND: Confirmed and fixed — switched to unprocessable() (422): an unresolvable scheduleId is a bad reference (the same class of error supervisor-operations-service's createCallLog uses unprocessable() for on an unknown sakhiId), not a state conflict, so conflict()/409 was the wrong fit. Also added the 422 to POST /visits' documented responses, which addresses the top-level doc-gap note too.

apps/visit-form-service/src/visits/visitInstance.service.ts
Dharanishwaran-ND: Agreed this is a real gap, but leaving it as-is for this PR — you're right that it matches beneficiary.service.ts's create() exactly (same check-then-insert shape, no P2002 catch-and-recover), so fixing it here alone would be inconsistent with the rest of the codebase rather than closing a gap specific to this change. If this needs a real fix (e.g. catching P2002 and re-querying, or a DB-level upsert), it should probably be done consistently across all three call sites (here, beneficiary.service.ts, form.service.ts) in its own follow-up rather than diverging just for visits.

Dharanishwaran-ND:
Fixed the status-code convention mismatch (409 → 422, matching form.service.ts/supervisor-operations-service) and added the missing 422 to POST /visits' documented responses. Left the concurrent-retry race as-is per the reviewer's own framing — it matches an established repo-wide pattern (beneficiary.service.ts's create() has the identical gap), so it's a separate follow-up affecting 3 call sites, not something to diverge on just for this PR.

PR #105, opened 2026-08-04
fix(beneficiary-service): Let first-time pregnancies pass the Gravida cross-total rule

Dharanishwaran-ND:
## Summary

Two unrelated fixes, bundled here because both are small MOTHER_REGISTRATION-form fixes flagged in the same requirements review:

**1. Gravida cross-total rule (blocker)**

`create-beneficiary.dto.ts` required `liveBirths + stillbirths + abortions == gravida`. For Gravida = 1 (a first-time pregnant woman), all three fields must be 0 to be otherwise valid — but 0 ≠ 1, so the rule could never pass. **Sakhis could not enroll any first-time pregnant woman.**

Fixed by changing the rule to `== gravida - 1` — the sum only covers pregnancies that have already ended; the current, still-ongoing pregnancy (counted in Gravida) has no outcome yet, so subtracting 1 excludes it. This is correct for every Gravida value, not just 1, and matches what the code's own prior comment already claimed the rule was supposed to do.

**2. Husband's name (schema gap)**

The Beneficiary Profile design has a "Husband's Name" row, but `MOTHER_REGISTRATION` had no such question — "husband" only appeared as an option value on `who_owns_the_phone`. Added a new optional text field, `husbands_name`, next to the woman's own personal-info fields.

## Test plan

- [x] `nx lint beneficiary-service visit-form-service` — clean
- [x] `nx test beneficiary-service` — 91/91 passing (existing M13 test's numbers updated to the new formula; added M14 covering Gravida=1 with all-zero outcomes, previously impossible, now passes)
- [x] Validated `mother-registration.json` still parses correctly after the new field

## Note for whoever reviews this

The para ≤ gravida / abortions ≤ gravida rules were left unchanged — they're correct either way, per the original report that flagged this bug.

### DIFF
--- a/apps/beneficiary-service/src/beneficiary/dto/create-beneficiary.dto.ts
+++ b/apps/beneficiary-service/src/beneficiary/dto/create-beneficiary.dto.ts
@@ -79,7 +79,15 @@ export function deriveFullName(
 const motherDetailsSchema = z
   .object({
     lmpDate: z.coerce.date(),
-    gravida: z.number().int().min(0).max(14).optional(),
+    // min(1), not min(0) — this is a MOTHER_REGISTRATION case (a currently
+    // pregnant woman), and Gravida counts the current pregnancy by
+    // definition, so 0 is never a legitimate value here. Also closes a real
+    // bug the min(0) allowed: with the sum == gravida - 1 rule above,
+    // gravida: 0 makes gravida - 1 evaluate to -1, which
+    // liveBirths+stillbirths+abortions (each itself min(0)) can never reach —
+    // every submission with gravida: 0 would fail the cross-check
+    // unconditionally, with no way to satisfy it.
+    gravida: z.number().int().min(1).max(14).optional(),
     parity: z.number().int().min(0).max(14).optional(),
     liveBirths: z.number().int().min(0).max(14).optional(),
     stillbirths: z.number().int().min(0).max(14).optional(),
--- a/apps/beneficiary-service/src/beneficiary/dto/create-beneficiary.dto.ts
+++ b/apps/beneficiary-service/src/beneficiary/dto/create-beneficiary.dto.ts
@@ -72,7 +72,9 @@ export function deriveFullName(
 /**
  * Cross-field consistency rules per SRS Category 3 (line 401): Para ≤
  * Gravida; Abortions ≤ Gravida; Dead children ≤ Live births; Live births +
- * Stillbirths + Abortions = Gravida (including the current pregnancy).
+ * Stillbirths + Abortions = Gravida - 1 (the sum covers only pregnancies that
+ * have already ended — the current pregnancy, counted in Gravida, has no
+ * outcome yet, so subtracting 1 excludes it).
  */
 const motherDetailsSchema = z
   .object({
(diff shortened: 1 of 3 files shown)


Dharanishwaran-ND:
Tracking issue: #111

palani-ND (commented):
## Review summary

Both fixes are small, well-scoped, and matched to what the report actually flagged — nothing here reaches beyond the SRS's Category 3 cross-total rule or the Beneficiary Profile's documented "Husband's Name" field. The `gravida - 1` correction is the right formula (Gravida counts the current, still-ongoing pregnancy, which by definition has no live-birth/stillbirth/abortion outcome yet), the new M14 test concretely proves the previously-impossible Gravida=1 case now passes, and the M13 fixture was correctly updated in place rather than left stale. The `husbands_name` field addition is clean: unique `question_code` (checked against all 74 codes in the file), correctly optional/text, consistent with sibling entries' shape.

One boundary case the new formula reintroduces, flagged inline below.

apps/beneficiary-service/src/beneficiary/dto/create-beneficiary.dto.ts
```diff
@@ -122,11 +124,11 @@ const motherDetailsSchema = z
     }
     if (liveBirths !== undefined && stillbirths !== undefined && abortions !== undefined) {
       const sum = liveBirths + stillbirths + abortions;
-      if (sum !== gravida) {
+      if (sum !== gravida - 1) {
```
palani-ND, line 135: `gravida` is declared as `z.number().int().min(0).max(14).optional()` (line 82) — 0 is a permitted value. With the new `sum !== gravida - 1` check, a submission with `gravida: 0` and all three outcome fields answered (e.g. all 0, the natural way to fill this out if you believe you've had zero pregnancies) now *always* fails, since `gravida - 1` evaluates to `-1` and `liveBirths + stillbirths + abortions` can never be negative (each field is itself `min(0)`). The Sakhi would see "liveBirths + stillbirths + abortions must equal gravida - 1" for an input that has no way to satisfy it — arguably worse than the old bug for this specific value, since before, `gravida: 0` with all outcomes 0 actually passed (0 = 0).

Since this is a `MOTHER_REGISTRATION` case (a currently pregnant woman), Gravida — which by this field's own definition includes the current pregnancy — should never legitimately be 0. Given this PR is already touching this exact field and rule, it seems like the more complete fix is to also bump the lower bound to `min(1)` on `gravida` here, rather than leaving 0 as a schema-valid value that the cross-check can never actually satisfy.

Dharanishwaran-ND: Confirmed and fixed exactly as suggested — bumped gravida's lower bound to min(1). You're right that Gravida=0 is not just an unaddressed boundary but a genuinely worse regression than the original bug: with the new gravida - 1 rule, gravida: 0 makes the target sum -1, which liveBirths+stillbirths+abortions (each min(0)) can never reach, so every gravida: 0 submission failed unconditionally with no way to satisfy it. Added M15 asserting gravida: 0 is now rejected at the schema level, before the cross-check even runs.

PR #106, opened 2026-08-04
feat(supervisor-operations-service): Add GET /call-sheet-stats (by-sakhi and batch)

Dharanishwaran-ND:
## ⚠️ Partial implementation — 1 of 7 stat kinds is real data

Per the mobile team's call-sheet-stats spec, this adds:

- `GET /call-sheet-stats/by-sakhi/:sakhiId` — single Sakhi
- `GET /call-sheet-stats?sakhiIds=id1,id2,id3` — batch, for a list-view card grid without one call per card

Auth: `trustGatewayIdentity` + `requireRoles('SUPERVISOR', 'MANAGER', 'ADMIN')` (the spec said `SUPERVISOR`/`MANAGER` only, but every other Sakhi-scoped endpoint in this file — `call-logs/by-sakhi/*`, `inventory-transactions/by-sakhi/*` — also allows ADMIN, so this follows that existing convention rather than the narrower spec list). Ownership check: 403 if the Sakhi isn't assigned to the calling Supervisor (MANAGER/ADMIN unrestricted) — reuses the exact pattern already in `listCallLogsBySakhi`.

## What's real vs. stubbed

Investigated before writing any code — **6 of the 7 "kind" rows have no existing backing data anywhere in the codebase**:

| Kind | Status |
|---|---|
| `FOLLOWUP_PENDING` | **Real.** `CallLog.callStatus === 'CALL_BACK'` count. |
| `VISIT_DUE`, `VISIT_3_DAYS_TO_EXPIRE`, `MISSED_VISIT` | Stubbed `{count: 0, updated: 0}`. visit-form-service has the right date fields but zero existing "due soon"/"missed" query logic — would need to be built there, then this service would need a cross-service client to call it. |
| `CLOSURE_FORM_PENDING` | Stubbed. closure-reopen-service tracks approval status of *already-submitted* closures, not "should be closed but isn't yet" — no matching semantic exists. |
| `HIGH_RISK_ANC`, `HIGH_RISK_PNC` | Stubbed. **"PNC" doesn't map to anything in this codebase** — the phase enum has `ANC`/`PP` (postpartum)/`NN`/etc., no `PNC`. Needs clarification from whoever wrote the spec before real logic can be built. Closest existing data (`BeneficiaryRiskConditionSummary` in beneficiary-service) has a `phase` field but no single "high-risk AND this phase" flag. |

This repo forbids cross-service DB joins (forklift rule) — even once the missing data models exist, correctly computing all 7 kinds would need real query logic built in 3-4 different services, not just an aggregator here. Scoped this PR to what's real today rather than guessing at undefined business rules.

## Test plan

- [x] `nx lint supervisor-operations-service api-gateway` — clean
- [x] `nx test supervisor-operations-service` — 150/150 passing (6 new: real FOLLOWUP_PENDING count + stub zeros, 403/404 ownership checks, batch omits inaccessible/unknown ids, batch returns `[]` rather than erroring when none are accessible)
- [x] `nx build supervisor-operations-service` — clean
- [x] Live verification against a running local gateway + supervisor-operations-service with a real SUPERVISOR JWT: single-Sakhi call returns 200 with real `FOLLOWUP_PENDING` count (matching an existing `CALL_BACK` call log) and correct zero-stubs for the other 6; batch call with one real + one fake id returns only the real one; unknown sakhiId → 404

## Follow-up needed before the other 6 kinds can be built

Whoever owns the mobile spec should clarify: (1) what "PNC" is supposed to map to, (2) where "visit due/expiring/missed" and "closure form pending" data should live and whether it needs to be modeled from scratch, (3) which service should own the cross-service aggregation once those exist.

### DIFF
--- a/apps/supervisor-operations-service/src/operations/operations.repository.ts
+++ b/apps/supervisor-operations-service/src/operations/operations.repository.ts
@@ -241,6 +241,17 @@ export class OperationsRepository {
   }
 
   /**
+   * Call-sheet stats' FOLLOWUP_PENDING count — a call logged as CALL_BACK
+   * (the caller asked to be rung again) with no newer call logged since.
+   * The only one of the 7 call-sheet-stats kinds backed by real data today.
+   */
+  countPendingFollowups(sakhiId: string) {
+    return this.prisma.callLog.count({
+      where: { sakhiId, isDeleted: false, callStatus: 'CALL_BACK' },
+    });
+  }
+
+  /**
    * Only ever writes the fields captured after a call ends (status, end
    * time, duration, notes, followup) — sakhiId/projectId/supervisorId/
    * callStartAt/callDatetime are immutable, matching this repo's
--- a/apps/supervisor-operations-service/src/operations/operations.repository.ts
+++ b/apps/supervisor-operations-service/src/operations/operations.repository.ts
@@ -241,14 +241,22 @@ export class OperationsRepository {
   }
 
   /**
-   * Call-sheet stats' FOLLOWUP_PENDING count — a call logged as CALL_BACK
-   * (the caller asked to be rung again) with no newer call logged since.
-   * The only one of the 7 call-sheet-stats kinds backed by real data today.
+   * Call-sheet stats' FOLLOWUP_PENDING count — 1 if this Sakhi's most recent
+   * call is still CALL_BACK (unresolved), 0 otherwise (including "never
+   * called"). NOT a count of every CALL_BACK ever logged: call_logs has no
+   * beneficiary/thread grouping (it's one Supervisor-to-Sakhi check-in call
+   * per row, per FR-SV-3.3), so a later call of any status supersedes an
+   * earlier CALL_BACK — it doesn't stay "pending" once the Supervisor has
+   * called again, whatever the outcome of that later call. The only one of
+   * the 7 call-sheet-stats kinds backed by real data today.
    */
-  countPendingFollowups(sakhiId: string) {
-    return this.prisma.callLog.count({
-      where: { sakhiId, isDeleted: false, callStatus: 'CALL_BACK' },
+  async countPendingFollowups(sakhiId: string): Promise<number> {
+    const latest = await this.prisma.callLog.findFirst({
+      where: { sakhiId, isDeleted: false },
+      orderBy: { callDatetime: 'desc' },
+      select: { callStatus: true },
     });
+    return latest?.callStatus === 'CALL_BACK' ? 1 : 0;
   }
 
   /**
--- a/apps/supervisor-operations-service/src/operations/operations.service.ts
+++ b/apps/supervisor-operations-service/src/operations/operations.service.ts
@@ -15,6 +15,8 @@ import type { CreateGatheringInput } from './dto/create-gathering.dto';
 import type { UpdateGatheringAttendanceInput } from './dto/update-gathering-attendance.dto';
 import type { CompleteTopicMarkInput, CreateTopicMarkInput } from './dto/create-topic-mark.dto';
 import type { TopicMarkQuery } from './dto/topic-mark-query.dto';
+import type { CallSheetStatKind } from './dto/call-sheet-stats.dto';
+import { CALL_SHEET_STAT_KINDS } from './dto/call-sheet-stats.dto';
 import type { SakhiClient } from './sakhi.client';
 
 /** Default recency window for the FR-SV-3.4 "recently called" card highlight. */
@@ -42,6 +44,26 @@ export interface CallerIdentity {
 /** MANAGER and ADMIN are unrestricted across all inventory ownership checks. */
 function isPrivileged(caller: CallerIdentity): boolean {
   return caller.roles.includes('MANAGER') || caller.roles.includes('ADMIN');
+}
+
+/** Today's date as YYYY-MM-DD, for call-sheet-stats' lastDataSyncDate. */
+function todayDateOnly(): string {
+  return new Date().toISOString().slice(0, 10);
+}
+
+/**
+ * Builds all 7 call-sheet-stats rows in the fixed kind order. Only
+ * FOLLOWUP_PENDING carries a real count; every other kind is a
+ * { count: 0, updated: 0 } placeholder — see call-sheet-stats.dto.ts.
+ */
+function buildCallSheetStatRows(
+  pendingFollowups: number,
+): Array<{ kind: CallSheetStatKind; updated: number; count: number }> {
+  return CALL_SHEET_STAT_KINDS.map((kind) => ({
+    kind,
+    updated: 0,
+    count: kind === 'FOLLOWUP_PENDING' ? pendingFollowups : 0,
+  }));
 }
 
 /** Supervisor operations domain logic. Data access is delegated to the repository. */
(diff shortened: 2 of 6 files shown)


Dharanishwaran-ND:
Tracking issue: #112

palani-ND (commented):
## Review summary

Appreciate the honesty here — 6 of 7 stat kinds are explicitly documented as stubs rather than guessed-at logic, and the PR is upfront about needing spec clarification (PNC mapping, visit-due semantics, cross-service ownership of the aggregation) before building the rest. That's the right call given the forklift rule (no cross-service DB joins) and undefined business rules for those 6 kinds — nothing here oversteps into scope the SRS/spec hasn't actually defined. Auth/ownership follows the existing `listCallLogsBySakhi` pattern correctly, and the `ADMIN`-vs-spec role deviation is called out and justified by existing convention in the same file.

One real correctness bug in the one stat that *is* claimed to be real, plus two lower-severity notes on the batch endpoint below.

apps/supervisor-operations-service/src/operations/operations.repository.ts
```diff
@@ -240,6 +240,17 @@ export class OperationsRepository {
     });
   }
 
+  /**
+   * Call-sheet stats' FOLLOWUP_PENDING count — a call logged as CALL_BACK
+   * (the caller asked to be rung again) with no newer call logged since.
+   * The only one of the 7 call-sheet-stats kinds backed by real data today.
+   */
+  countPendingFollowups(sakhiId: string) {
```
palani-ND: The doc comment above this method says FOLLOWUP_PENDING is "a call logged as CALL_BACK ... with no newer call logged since," but the query itself is just `count({ where: { sakhiId, isDeleted: false, callStatus: 'CALL_BACK' } })` — it counts *every* CALL_BACK row ever logged for this Sakhi, with no check for whether a later call already resolved it. Concretely: Supervisor calls on Jan 1 (CALL_BACK), calls again Jan 5 (CONNECTED — issue resolved), calls again Jan 10 (CALL_BACK). This returns 2, but only the Jan 10 call is actually still pending — the Jan 1 one was already followed up on. Since this is the single kind this PR claims is real (not a stub), the count a Supervisor sees on their call-sheet card would be inflated for any Sakhi with more than one historical CALL_BACK, which seems worth fixing before merge rather than shipping as the one "real" number in an otherwise-stubbed feature. Would need something like: only the *latest* call log per Sakhi counts, and only if its status is CALL_BACK — e.g. filtering to rows with no later `callDatetime` for the same `sakhiId`, or restructuring to look at each Sakhi's most recent call only.

apps/supervisor-operations-service/src/operations/operations.service.ts
```diff
@@ -353,6 +375,54 @@ export class OperationsService {
...
+  async getCallSheetStatsBatch(
+    sakhiIds: string[],
+    caller: CallerIdentity,
+    authorizationHeader: string,
+  ) {
+    const results = await Promise.all(
+      sakhiIds.map(async (sakhiId) => {
+        try {
+          return await this.getCallSheetStats(sakhiId, caller, authorizationHeader);
+        } catch {
```
palani-ND: This bare `catch { return null }` swallows everything from `getCallSheetStats` for this id — not just the documented "unauthorized or unknown id" (403/404) cases, but also e.g. a `badGateway` from `sakhiClient.findById` if auth-service is unreachable, or any other unexpected error. A genuine infra failure on one id would silently render as "this Sakhi has no card," indistinguishable from a legitimately inaccessible Sakhi, with nothing surfaced to the caller or logged. Worth narrowing this to the specific expected statuses (403/404) and letting anything else propagate (or at least get logged) so a real backend problem doesn't masquerade as an empty result.

apps/supervisor-operations-service/src/operations/operations.service.ts
```diff
@@ -353,6 +375,54 @@ export class OperationsService {
...
+  }
+
+  /**
+   * Batch variant of getCallSheetStats for a list-view's card grid, so the
+   * caller isn't forced into one request per Sakhi card. A sakhiId not
+   * assigned to the caller is silently omitted from the result rather than
+   * failing the whole batch — the caller's other, legitimate cards should
+   * still render.
+   */
+  async getCallSheetStatsBatch(
```
palani-ND, line 413: Minor, given the PR's own stated purpose ("without one call per card"): this still does one `sakhiClient.findById` HTTP round-trip to auth-service *per* `sakhiId` in the batch (parallelized via `Promise.all`, but still N calls), plus one `countPendingFollowups` DB query per id. For a list-view with a Supervisor's full roster, that's N auth-service round-trips fired concurrently rather than the one-roster-fetch-then-filter-locally pattern already used elsewhere in this codebase for the same "batch-verify ownership of many Sakhis" problem (e.g. beneficiary-service's `listSakhiIdsForSupervisor`, which fetches the Supervisor's Sakhi list once and filters in memory). Also, `listCallSheetStatsBatchQuerySchema` has no cap on how many ids can be comma-separated, so this fans out unbounded. Not necessarily blocking for a first cut, but worth a look before this gets used against a large roster.

apps/supervisor-operations-service/src/operations/operations.repository.ts
Dharanishwaran-ND: Confirmed and fixed — this was a real bug, agreed it's the most important one to get right since it's the one kind this PR claims is real. Rewrote to check only the Sakhi's single most recent call log (findFirst, orderBy callDatetime desc): count is 1 if that latest call is still CALL_BACK, 0 otherwise. Since call_logs has no beneficiary/thread grouping — it's one Supervisor-to-Sakhi check-in call per row — a later call of any status genuinely supersedes an earlier CALL_BACK, so 'most recent call's status' is the correct semantic, not 'count of every CALL_BACK ever logged'. Updated the doc comment and the test's mock value (was 3, now 1, matching the new 0-or-1 contract).

apps/supervisor-operations-service/src/operations/operations.service.ts
Dharanishwaran-ND: Fixed — narrowed the catch to only swallow HttpError with status 403 or 404 (the documented unauthorized/unknown cases); anything else now rethrows and fails the whole batch rather than silently rendering as an empty result. Added a test asserting a badGateway-style error propagates instead of being swallowed.

apps/supervisor-operations-service/src/operations/operations.service.ts
Dharanishwaran-ND: Added a cap (100 ids, matching the visit-schedules bulk-upload endpoint's own cap) so the batch can't fan out unbounded — quick fix for the immediate risk. Left the N-calls-per-batch architecture itself as-is per your framing (not blocking for a first cut) — switching to a fetch-once-then-filter roster pattern like listSakhiIdsForSupervisor would be a bigger change worth its own follow-up rather than folding into this PR.

Dharanishwaran-ND:
Fixed all 3 findings: FOLLOWUP_PENDING now correctly checks only the Sakhi's most recent call (1 if still CALL_BACK, 0 otherwise) instead of counting every historical CALL_BACK row; the batch endpoint's error handling now only swallows 403/404 and propagates genuine infra failures; and the batch's sakhiIds is capped at 100. Live-verified through the running gateway: a Sakhi with a CALL_BACK as her most recent call now correctly shows count:1, and a 101-id batch is correctly rejected with 400.

PR #113, opened 2026-08-05
feat(quick-response): Add Quick Response API for EDD_NEARING and REOPEN cards

Dharanishwaran-ND:
## Summary

Implements the Quick Response API's core surface:
- `GET /api/v1/quick-response` — merges `approval_requests` (approval-service) and `escalation_events` (notification-escalation-service) into one cursor-paginated, `createdAt DESC` feed, spanning all 8 card types documented in the spec.
- `POST /api/v1/quick-response/:cardId/decision` — fully wired for **EDD_NEARING** (acknowledge-only, no audit/notify per spec) and **REOPEN** (delegates to a new `PATCH /reopen-requests/:id/decision` in closure-reopen-service, then writes an audit entry and notifies the Sakhi). The other 6 card types (`LMP_CHANGE`, `MISSED_VISIT`, `CLOSURE_REVIEW`, `REFERRAL_INCOMPLETE`, `DATA_RESTORE`, `ACCOMPANIED_REFERRAL`) are listed but respond `501 Not Implemented` on decision — their real side effects (ANC schedule regen, incentive triggers, closure/referral status writes) need write endpoints that don't exist yet in visit-form-service, risk-referral-service, and incentive-wages-service. Scoped this way deliberately, as agreed, rather than guessing at undefined cross-service behavior.

## Why the scope is split this way

Every other card type's approve/reject action needs a **new write endpoint in a service that currently only has GET/POST**. Building all 8 in one PR isn't reviewable; this PR proves the end-to-end architecture (merge, cursor pagination, decision dispatch, audit/notify) on the two card types where the underlying schema already supports it.

## Notable decisions (confirmed during implementation)

- `ApprovalRequestType.REFERRAL_SKIP` renamed to `REFERRAL_INCOMPLETE` to match the spec — naming drift only, migration included, 0 affected rows confirmed before writing it.
- `EscalationType` gained a new `EDD_NEARING` value (nothing today represents "EDD approaching," only "already missed" via `POST_EDD_MISSED`); `MISSED_VISIT` is a grouping of the 10 existing `*_MISSED` enum values, not a schema change.
- REOPEN's REJECT is persisted as `ReopenRequest.supervisorStatus = REJECTED` — this already matches the spec's "Cannot re-open" state; no new column needed.
- `POST /notifications` (notification-escalation-service) and `POST /audit` (audit-service) are widened from `ADMIN`-only to also allow `SUPERVISOR`, since approval-service calls them directly (service-to-service, forwarding the deciding Supervisor's own JWT — same pattern as supervisor-operations-service's existing `SakhiClient`). **Security note:** an automated review flagged this widening as a potential impersonation/IDOR risk if left unconstrained. Fixed before merge:
  - `notification-escalation-service`: a non-ADMIN caller can only notify a Sakhi actually assigned to them (new ownership check via a `SakhiClient`, mirroring the existing pattern).
  - `audit-service`: a non-ADMIN caller's `actorUserId` is always forced server-side to their own id (never the client-supplied value), and they're restricted to `QUICK_RESPONSE_*` actions.
  - Both fixes verified live via real API calls (403 on the abuse path, 200/201 on the legitimate path).

## Test plan

- `nx lint` + `nx test` clean across all 5 touched projects (notification-escalation-service, closure-reopen-service, audit-service, approval-service, api-gateway) — 71 unit tests covering merge/pagination, decision dispatch per card type, 404/409/501 edge cases, and the two security fixes.
- Live-verified via real HTTP calls: the notification ownership check and the audit actorUserId/action constraints, both the abuse path (403) and the legitimate path (200/201).
- **Not live-verified:** `GET /quick-response` and closure-reopen-service's new `PATCH /reopen-requests/:id/decision` end-to-end — blocked by pre-existing schema drift on the shared dev DB (`approval_requests` missing `decision_status_lookup_id`, `reopen_requests` table absent), unrelated to this PR. Flagging for whoever owns that environment; the code is built and tested against the current `schema.prisma`.

## Follow-ups (out of scope here, tracked separately)

- #114 — Wire the remaining 6 card types once their backing services grow the needed write endpoints.
- #115 — Define a permanent service-to-service auth credential (replace JWT-forwarding pattern).
- #116 — EDD_NEARING acknowledgement doesn't persist a status change on escalation_events.
- #117 — Dev Supabase schema drift blocks live verification of approval-service and closure-reopen-service.

### DIFF
--- a/apps/approval-service/.env.example
+++ b/apps/approval-service/.env.example
@@ -8,3 +8,11 @@ CORS_ORIGINS=http://localhost:5173
 # https://staging-api.example.com,https://api.example.com. Leave empty in
 # local dev to fall back to http://localhost:$PORT.
 PUBLIC_BASE_URLS=
+# Base URLs Quick Response calls directly (service-to-service, forwarding the
+# caller's own Authorization header — same pattern as supervisor-operations-
+# service's SakhiClient): resolving APPROVAL_STATUS lookup values, fetching
+# escalation-sourced cards, deciding REOPEN cards, and writing audit entries.
+AUTH_SERVICE_BASE_URL=http://localhost:3000
+NOTIFICATION_ESCALATION_SERVICE_BASE_URL=http://localhost:3009
+CLOSURE_REOPEN_SERVICE_BASE_URL=http://localhost:3006
+AUDIT_SERVICE_BASE_URL=http://localhost:3013
--- a/apps/approval-service/.env.example
+++ b/apps/approval-service/.env.example
@@ -8,11 +8,3 @@ CORS_ORIGINS=http://localhost:5173
 # https://staging-api.example.com,https://api.example.com. Leave empty in
 # local dev to fall back to http://localhost:$PORT.
 PUBLIC_BASE_URLS=
-# Base URLs Quick Response calls directly (service-to-service, forwarding the
-# caller's own Authorization header — same pattern as supervisor-operations-
-# service's SakhiClient): resolving APPROVAL_STATUS lookup values, fetching
-# escalation-sourced cards, deciding REOPEN cards, and writing audit entries.
-AUTH_SERVICE_BASE_URL=http://localhost:3000
-NOTIFICATION_ESCALATION_SERVICE_BASE_URL=http://localhost:3009
-CLOSURE_REOPEN_SERVICE_BASE_URL=http://localhost:3006
-AUDIT_SERVICE_BASE_URL=http://localhost:3013
--- a/apps/approval-service/src/quick-response/escalation.client.ts
+++ b/apps/approval-service/src/quick-response/escalation.client.ts
@@ -1,0 +1,58 @@
+import { badGateway, HttpError } from '@armman/service-commons';
+import { appConfig } from '../config/app-config';
+
+export interface EscalationCard {
+  cardId: string;
+  cardType: 'MISSED_VISIT' | 'EDD_NEARING';
+  cardSource: 'escalation_events';
+  beneficiaryId: string;
+  visitId: string | null;
+  referralId: string | null;
+  escalationType: string;
+  status: string;
+  raisedAt: string;
+}
+
+interface EscalationEventsResponse {
+  cards: EscalationCard[];
+  nextCursor: string | null;
+}
+
+/**
+ * Fetches escalation-sourced Quick Response cards from
+ * notification-escalation-service — called directly (not through the
+ * gateway), forwarding the caller's own Authorization header, same pattern
+ * supervisor-operations-service's SakhiClient uses.
+ */
+export class EscalationClient {
+  async list(
+    status: string,
+    cursor: string | undefined,
+    limit: number,
+    authorizationHeader: string,
+  ): Promise<EscalationEventsResponse> {
+    const params = new URLSearchParams({ status, limit: String(limit) });
+    if (cursor) params.set('cursor', cursor);
+
+    let res: Response;
+    try {
+      res = await fetch(
+        `${appConfig.NOTIFICATION_ESCALATION_SERVICE_BASE_URL}/api/v1/escalation-events?${params.toString()}`,
+        { headers: { Authorization: authorizationHeader } },
+      );
+    } catch {
+      throw badGateway('Unable to fetch escalation events — the service is unreachable.');
+    }
+
+    if (!res.ok) {
+      if (res.status >= 400 && res.status < 500) {
+        const body = (await res.json().catch(() => null)) as { message?: string } | null;
+        throw new HttpError(res.status, body?.message ?? 'Unable to fetch escalation events.');
+      }
+      throw badGateway('Unable to fetch escalation events — the service returned an error.');
+    }
+
+    const body = (await res.json()) as { data: EscalationEventsResponse };
+    return body.data;
+  }
+}
--- a/apps/approval-service/src/quick-response/escalation.client.ts
+++ b/apps/approval-service/src/quick-response/escalation.client.ts
@@ -20,9 +20,9 @@ interface EscalationEventsResponse {
 
 /**
  * Fetches escalation-sourced Quick Response cards from
- * notification-escalation-service — called directly (not through the
- * gateway), forwarding the caller's own Authorization header, same pattern
- * supervisor-operations-service's SakhiClient uses.
+ * notification-escalation-service — called through the gateway, forwarding
+ * the caller's own Authorization header, same pattern supervisor-operations-
+ * service's SakhiClient uses.
  */
 export class EscalationClient {
   async list(
@@ -53,6 +53,30 @@ export class EscalationClient {
     }
 
     const body = (await res.json()) as { data: EscalationEventsResponse };
+    return body.data;
+  }
+
+  /** Fetches a single escalation-sourced card, or null if it doesn't exist. */
+  async findById(id: string, authorizationHeader: string): Promise<EscalationCard | null> {
+    let res: Response;
+    try {
+      res = await fetch(`${appConfig.API_GATEWAY_BASE_URL}/api/v1/escalation-events/${id}`, {
+        headers: { Authorization: authorizationHeader },
+      });
+    } catch {
+      throw badGateway('Unable to fetch the escalation event — the service is unreachable.');
+    }
+
+    if (res.status === 404) return null;
+    if (!res.ok) {
+      if (res.status >= 400 && res.status < 500) {
+        const body = (await res.json().catch(() => null)) as { message?: string } | null;
+        throw new HttpError(res.status, body?.message ?? 'Unable to fetch the escalation event.');
+      }
+      throw badGateway('Unable to fetch the escalation event — the service returned an error.');
+    }
+
+    const body = (await res.json()) as { data: EscalationCard };
     return body.data;
   }
 }
--- a/apps/approval-service/src/quick-response/escalation.client.ts
+++ b/apps/approval-service/src/quick-response/escalation.client.ts
@@ -37,7 +37,7 @@ export class EscalationClient {
     let res: Response;
     try {
       res = await fetch(
-        `${appConfig.NOTIFICATION_ESCALATION_SERVICE_BASE_URL}/api/v1/escalation-events?${params.toString()}`,
+        `${appConfig.API_GATEWAY_BASE_URL}/api/v1/escalation-events?${params.toString()}`,
         { headers: { Authorization: authorizationHeader } },
       );
     } catch {
--- a/apps/approval-service/src/quick-response/quick-response.service.ts
+++ b/apps/approval-service/src/quick-response/quick-response.service.ts
@@ -3,8 +3,6 @@ import type { QuickResponseRepository } from './quick-response.repository';
 import type { LookupClient } from './lookup.client';
 import type { EscalationClient, EscalationCard } from './escalation.client';
 import type { ReopenRequestClient } from './reopen-request.client';
-import type { NotificationClient } from './notification.client';
-import type { AuditClient } from './audit.client';
 import type { ListQuickResponseInput } from './dto/list-quick-response.dto';
 import type { DecideQuickResponseInput } from './dto/decide-quick-response.dto';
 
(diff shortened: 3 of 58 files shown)


apps/approval-service/src/quick-response/quick-response.service.ts
```diff
@@ -0,0 +1,226 @@
...
+    if (existing.requestedByUserId) {
+      await this.notificationClient.notify(
+        existing.requestedByUserId,
+        'REOPEN_UPDATE',
+        'Reopen request decided',
+        dto.decision === 'APPROVE'
+          ? 'Your reopen request was approved.'
+          : 'Your reopen request was rejected.',
+        authorizationHeader,
+      );
```
palani-ND: This `notificationClient.notify()` call (and `auditClient.log()` above it) isn't wrapped in a try/catch, but by this point `reopenRequestClient.decide()` has already committed the decision in closure-reopen-service — it can't be undone.

Concretely: `GET /quick-response` never filters cards by the caller's own team (no `assignedSupervisorId`/supervisor scoping in `quick-response.repository.ts` or `escalation.repository.ts`), so any SUPERVISOR can view and decide any REOPEN card. But `notification-escalation-service`'s new ownership check (`notification.service.ts`) requires the *deciding* supervisor to equal the *requesting Sakhi's* assigned supervisor. Whenever a Supervisor decides a card for a Sakhi who isn't their own — which the current design allows — this `notify()` call throws a 403, uncaught, and the whole `POST /quick-response/:cardId/decision` request fails even though the reopen request is already `APPROVED`/`REJECTED` and the audit entry already written.

A retry from the client then gets a 409 ("already decided"), so there's no way to recover cleanly — the client is stuck believing the action failed when it actually succeeded.

Suggest: don't let a failed notify (or audit) flip an already-committed decision into an error response — e.g. catch and log/report the notify failure without rethrowing, or resolve the ownership-check semantics so it can't reject a decision that Quick Response itself allowed to be made.

apps/approval-service/src/quick-response/quick-response.service.ts
```diff
@@ -0,0 +1,226 @@
...
+    // The actual PENDING->OPEN->DISMISSED status write happens via a future
+    // escalation-decision endpoint in notification-escalation-service —
+    // out of scope for this phase's EDD_NEARING vertical slice beyond
+    // proving the read+dispatch path; tracked as a follow-up.
+    return {
+      cardId,
+      cardSource: 'escalation_events' as const,
+      decision: 'OKAY',
+      acknowledged: true,
+    };
```
palani-ND, line 161: This never verifies `cardId` corresponds to a real escalation_events row before echoing back `acknowledged: true` — any UUID with `cardSource: 'escalation_events'` and `decision: 'OKAY'` gets a 200. I get that this is deliberately a no-op pending #116, but returning success for a card that may not exist (or was already resolved) seems like it should at least 404 rather than silently succeed.

palani-ND (commented):
Reviewed the Quick Response API surface end-to-end (approval-service, closure-reopen-service, notification-escalation-service, audit-service, api-gateway). The architecture is solid — cursor merge across two sources, the 501-for-unimplemented-card-types scoping, and the audit/notify ownership fixes are all reasonable calls. Left inline comments on a few things worth another look before merge, summarized here:

1. **Likely bug: a successful REOPEN decision can still fail the request.** In `quick-response.service.ts`, `notificationClient.notify()` runs *after* the reopen request is already decided and audited, with no try/catch. `GET /quick-response` doesn't scope cards by the caller's own team (no `assignedSupervisorId` filter anywhere), so any SUPERVISOR can decide any REOPEN card — but `notification-escalation-service`'s new ownership check requires the deciding supervisor to be the *requesting Sakhi's own* assigned supervisor. Whenever those differ, `notify()` throws, and the whole decision request comes back as an error even though the underlying state change already committed. A retry then hits 409 (already decided), so the caller has no clean path forward. Recommend not letting a best-effort side effect (notify, and arguably audit) turn an already-committed decision into a reported failure.

2. **EDD_NEARING acknowledge never checks the card exists.** `decideEscalationCard` returns `{decision: 'OKAY', acknowledged: true}` for literally any `cardId` — it never calls back into notification-escalation-service to confirm the id is a real, still-open escalation event. Understandable given #116 (no persisted state yet), but worth at least a 404 for a nonexistent/bogus id rather than an unconditional 200.

3. **Bypassable audit trail.** `PATCH /reopen-requests/:id/decision` is commented as "internal-use, not part of the public surface," but it's mounted through the gateway with `requireRoles('SUPERVISOR', 'MANAGER', 'ADMIN')` and no other gate — any of those roles can call it directly and decide a reopen request while completely skipping the `QUICK_RESPONSE_*` audit_log entry and the Sakhi notification, since those only happen in approval-service's orchestration layer. If an audit trail is meant to be mandatory for every reopen decision (which the amount of care put into the audit/notify wiring suggests), this is a real gap, not just a comment nit.

4. **Stale docs left over from the gateway-routing fix.** All four Quick Response cross-service clients (`audit.client.ts`, `escalation.client.ts`, `notification.client.ts`, `reopen-request.client.ts`) still say "directly (not through the gateway)" in their doc comments, even though the third commit rerouted every one of them through `API_GATEWAY_BASE_URL`. `.env.example`'s `AUTH_SERVICE_BASE_URL` / `NOTIFICATION_ESCALATION_SERVICE_BASE_URL` / `CLOSURE_REOPEN_SERVICE_BASE_URL` / `AUDIT_SERVICE_BASE_URL` are the same leftover — `app-config.ts` only reads `API_GATEWAY_BASE_URL` now, so those four are dead config describing an approach that was explicitly abandoned. Worth cleaning up so nobody "fixes" these back to direct calls based on the stale comments.

Smaller note: `quick-response.service.spec.ts` mocks `reopenRequestClient`/`auditClient`/`notificationClient` completely, so none of the 71 unit tests would catch #1 — a test asserting the decision's outcome when notify/audit throws *after* a successful `reopenRequestClient.decide()` would have caught this.

apps/closure-reopen-service/src/reopen-requests/reopen-request.controller.ts
```diff
@@ -0,0 +1,93 @@
...
+        200: { description: 'Reopen request decided', schema: envelope(reopenRequestSchema) },
+        400: { description: 'Validation error', schema: apiErrorSchema },
+        401: { description: 'Unauthenticated', schema: apiErrorSchema },
+        403: { description: 'Caller role not permitted', schema: apiErrorSchema },
+        404: { description: 'Reopen request not found', schema: apiErrorSchema },
+        409: { description: 'Already decided', schema: apiErrorSchema },
+      },
+    },
+    trustGatewayIdentity,
+    requireRoles('SUPERVISOR', 'MANAGER', 'ADMIN'),
```
palani-ND, line 87: This endpoint is annotated elsewhere as "internal-use... not part of the public Quick Response surface," but nothing here actually restricts it to being called via approval-service's orchestration — any SUPERVISOR/MANAGER/ADMIN with a valid JWT can hit this directly through the gateway and decide a reopen request while completely bypassing the `QUICK_RESPONSE_*` audit_log entry and Sakhi notification that `quick-response.service.ts` wires up. If the audit trail is meant to be mandatory for reopen decisions, this is a real bypass, not just a naming/documentation concern.

apps/approval-service/src/quick-response/escalation.client.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+interface EscalationEventsResponse {
+  cards: EscalationCard[];
+  nextCursor: string | null;
+}
+
+/**
+ * Fetches escalation-sourced Quick Response cards from
+ * notification-escalation-service — called directly (not through the
```
palani-ND: This comment (and the equivalent one in `audit.client.ts`, `notification.client.ts`, `reopen-request.client.ts`) says "called directly (not through the gateway)" but the implementation below calls `appConfig.API_GATEWAY_BASE_URL` — this is stale from before the third commit ("Route cross-service calls through the gateway, not direct") fixed the actual routing. Worth updating so a future reader doesn't "fix" this back to a direct call based on the comment.

apps/approval-service/.env.example
```diff
@@ -8,3 +8,11 @@ CORS_ORIGINS=http://localhost:5173
 # https://staging-api.example.com,https://api.example.com. Leave empty in
 # local dev to fall back to http://localhost:$PORT.
 PUBLIC_BASE_URLS=
+# Base URLs Quick Response calls directly (service-to-service, forwarding the
+# caller's own Authorization header — same pattern as supervisor-operations-
+# service's SakhiClient): resolving APPROVAL_STATUS lookup values, fetching
+# escalation-sourced cards, deciding REOPEN cards, and writing audit entries.
+AUTH_SERVICE_BASE_URL=http://localhost:3000
```
palani-ND: `AUTH_SERVICE_BASE_URL`, `NOTIFICATION_ESCALATION_SERVICE_BASE_URL`, `CLOSURE_REOPEN_SERVICE_BASE_URL`, and `AUDIT_SERVICE_BASE_URL` aren't read anywhere — `app-config.ts` only defines `API_GATEWAY_BASE_URL` (all Quick Response clients route through the gateway now, per the third commit's fix). These four look like leftovers from the pre-fix direct-call approach; safe to drop along with the comment above them describing that abandoned design.

apps/approval-service/src/quick-response/quick-response.service.ts
Dharanishwaran-ND: Fixed by moving the audit_log write and Sakhi notification into closure-reopen-service's own `ReopenRequestService.decide`, rather than doing them here after the fact. That's the only place a reopen decision is actually persisted, so:

1. The failure mode you flagged is gone — `decide()` now catches a notify/audit failure and logs it (`console.error`) instead of letting it surface as a request error. The reopen decision was already committed by that point, so a downstream notify 403 can no longer flip a successful decision into an error response with no clean retry.
2. This also closes the bypass from comment below (the "internal-use" PATCH endpoint) — the audit trail is written unconditionally now, regardless of whether the caller reached it via Quick Response or hit it directly.

`quick-response.service.ts`'s REOPEN path is simplified to just call `reopenRequestClient.decide()` and return — no more duplicate audit/notify calls here.

apps/approval-service/src/quick-response/quick-response.service.ts
Dharanishwaran-ND: Fixed. Added `GET /escalation-events/:id` in notification-escalation-service (`EscalationRepository.findById` + `EscalationService.findById`, reusing the existing card-type filter) and an `EscalationClient.findById()` in approval-service. `decideEscalationCard` now calls it first and throws a 404 via `notFound()` when the card doesn't exist or isn't one of the 8 supported card types, before checking the decision value.

apps/closure-reopen-service/src/reopen-requests/reopen-request.controller.ts
Dharanishwaran-ND: Agreed this was a real bypass, not just a docs issue. Fixed by moving the audit_log entry and Sakhi notification into `ReopenRequestService.decide` itself (closure-reopen-service), so they're written unconditionally whenever a reopen request is decided — whether the caller is approval-service's Quick Response flow or a direct call through the gateway. The comment above the router now reflects that. approval-service no longer writes its own audit/notify for REOPEN cards, since it would now be redundant.

apps/approval-service/src/quick-response/escalation.client.ts
Dharanishwaran-ND: Fixed — updated the doc comments in escalation.client.ts, reopen-request.client.ts, and escalation.controller.ts (notification-escalation-service side) to say "through the gateway" instead of "not through the gateway / called directly." (audit.client.ts and notification.client.ts no longer live in approval-service — they moved to closure-reopen-service as part of fixing the audit/notify-bypass comment below, with correct gateway wording from the start.)

apps/approval-service/.env.example
Dharanishwaran-ND: Removed all four (AUTH_SERVICE_BASE_URL, NOTIFICATION_ESCALATION_SERVICE_BASE_URL, CLOSURE_REOPEN_SERVICE_BASE_URL, AUDIT_SERVICE_BASE_URL) along with the stale comment describing the abandoned direct-call design. Confirmed nothing in approval-service's app-config.ts or source reads them.

PR #118, opened 2026-08-05
fix(visit-form-service): Land POST /visit-schedules/bulk into develop

Dharanishwaran-ND:
## Summary

`POST /visit-schedules/bulk` (PR #103) was fully built, reviewed, and merged — but into `feat/rules-service-version-lookup` instead of `develop`, due to a branch-basing mixup. Its 2 commits never reached `develop`. This PR cherry-picks them cleanly (no conflicts) onto current `develop`.

Closes the last 3 P0 gaps from the mobile team's `SAKHI-APP-BACKEND-REQUIREMENTS.md`:
- `local_schedule_uuid` migration on `visit_schedules` (device-generated idempotency key, mirrors `VisitInstance.localVisitUuid`)
- `POST /visit-schedules/bulk` itself — idempotent by `localScheduleUuid`, all-or-nothing per batch, resolves HR-visit anchors against both already-stored rows and sibling rows in the same batch, rejects a conflicting re-upload under a new UUID with `409` rather than silently overwriting
- `/visit-schedules` gateway route

Also includes a follow-up fix (already part of the original PR): real ownership enforcement — a SAKHI caller must own the beneficiary directly, a SUPERVISOR caller must have that beneficiary's Sakhi assigned to them (new `sakhi.client.ts`, mirrors `supervisor-operations-service`'s existing `SakhiClient`/`listCallLogsBySakhi` pattern). Both `beneficiary.client.ts` and `ruleVersion.client.ts` route through the gateway (not the target service's own port) so the caller's JWT gets verified and forwarded as trusted headers — same lesson learned independently in the Quick Response API work.

## Test plan

- `nx lint` + `nx test` clean on `visit-form-service` (97 tests), `api-gateway` (19 tests), `service-commons` (47 tests, covers the new `ErrorCode.PAYLOAD_TOO_LARGE`/413 mapping).
- Live-verified end-to-end against local Postgres (migrations applied fresh, real seeded users, a real beneficiary created via `POST /beneficiaries`):
  - Happy path: `201`, 2 schedules created with real `scheduleId`s.
  - Idempotency: identical payload replayed → `created: 0, alreadyExisted: 2`, same `scheduleId`s.
  - Conflict: same `visitCode` under a different `localScheduleUuid` → `409`.
  - Validation: `visitCode`/`sequenceNo` mismatch → `400`.
  - Not found: unknown `beneficiaryId` → `404`.
  - Ownership: SUPERVISOR without the beneficiary's Sakhi assigned to them → `403`.

## Follow-ups (unchanged from the original PR, still out of scope)

- `GET /visit-schedules` (read-back) and `PATCH /visit-schedules/supersede` — not called by the mobile app yet, explicitly marked P2 in the spec.


apps/visit-form-service/src/visit-schedules/visitSchedule.service.ts
```diff
@@ -0,0 +1,213 @@
...
+      const existing = existingByLocalUuid.get(row.localScheduleUuid);
+      if (existing) {
+        alreadyExisted.push({
+          localScheduleUuid: row.localScheduleUuid,
+          scheduleId: existing.id,
+          status: existing.status,
+        });
+      } else {
+        newRows.push(row);
+      }
```
palani-ND, line 181: This only checks whether `row.localScheduleUuid` was already stored — it never compares `row`'s other fields (`visitCode`, `sequenceNo`, dates, `anchorType`, `anchorVisitLocalUuid`) against the matched `existing` row. A replay under the same key with different content silently succeeds and returns the *old* stored data, with no indication of the mismatch. Worth at least a comparison that 409s (or logs loudly) on a same-uuid-different-payload replay, since "idempotent by localScheduleUuid" is the headline guarantee of this endpoint.

apps/visit-form-service/src/visit-schedules/visitSchedule.repository.ts
```diff
@@ -0,0 +1,119 @@
...
+  // so createAllOrNothing patches anchorVisitId in a second pass.
+  batchAnchorLocalUuid: string | null;
+}
+
+/** Data access for visit schedules. Owns only this service's `visit_schedules` table. */
+export class VisitScheduleRepository {
+  constructor(private readonly prisma: PrismaService) {}
+
+  /** Every existing row among the given localScheduleUuids, for the idempotency/conflict checks. */
+  findByLocalScheduleUuids(localScheduleUuids: string[]) {
```
palani-ND: No `beneficiaryId` filter here — this queries by `localScheduleUuid` alone across the whole table. `createBulk` establishes `dto.beneficiaryId` as the authorized scope via the ownership check just above where this is called, but if a `localScheduleUuid` collides with a row belonging to a *different* beneficiary, this returns that row, and the service echoes its `id`/`status` back as "already existed" without ever checking `existing.beneficiaryId === dto.beneficiaryId`. That's a way to read another beneficiary's schedule id/status through the idempotency path even though the main create path's ownership check would reject the same caller for that beneficiary directly.

apps/visit-form-service/src/visit-schedules/dto/create-visit-schedule-bulk.dto.ts
```diff
@@ -0,0 +1,91 @@
...
+
+// schedules has no upper bound here — a >100-row batch must fail with 413
+// (payloadTooLarge), not the 400 Zod would raise via .max(). The size check
+// runs first, in visitSchedule.service.ts, before this schema even sees the
+// body.
+export const createVisitScheduleBulkSchema = z
+  .object({
+    beneficiaryId: z.string().uuid(),
+    generatedByRuleVersionId: z.string().uuid(),
+    generatedAt: z.coerce.date(),
```
palani-ND: `generatedAt` is required here, but grepping the service/repository, it's never read after `createVisitScheduleBulkSchema` parses it — and `VisitSchedule` in `schema.prisma` has no corresponding column. Either wire it into something (stored timestamp, drift check against `scheduledDate`) or drop it from the request contract; as-is every client is forced to send a value that goes nowhere.

palani-ND (commented):
Reviewed the visit-schedules-bulk landing (visit-form-service, api-gateway, service-commons, plus the beneficiary-service Gravida/Para tweak). The bulk-upload design itself is careful — all-or-nothing transaction, batch-sibling anchor resolution, 413 vs 400 distinction, the SAKHI/SUPERVISOR ownership fix from the original PR review. A few things worth a look before merge:

1. **Idempotency replay doesn't check the payload actually matches.** When a `localScheduleUuid` matches an existing row, `createBulk` treats it as `alreadyExisted` and echoes back the stored `id`/`status` — it never compares the replayed row's `visitCode`/`sequenceNo`/dates/`anchorType` against what's actually stored. A client that reuses a `localScheduleUuid` for genuinely different content (retry-logic bug, regenerated schedule under a new rule version reusing an old key, etc.) gets a silent 201 with the *old* stored data and no signal anything diverged. True idempotency-key semantics usually either verify an exact match or reject a same-key-different-payload replay as a conflict.

2. **The idempotency lookup isn't scoped to the caller's beneficiary.** `findByLocalScheduleUuids` queries by `localScheduleUuid` alone, with no `beneficiaryId` filter — even though `createBulk` just established `dto.beneficiaryId` as the authorized scope via the SAKHI/SUPERVISOR ownership check a few lines above. If a `localScheduleUuid` happens to collide with a row belonging to a *different* beneficiary (device bug, weak client-side UUID generation, or a caller replaying/guessing an id), the response reports that other beneficiary's `scheduleId`/`status` as "already existed" — a side door around the exact ownership boundary this PR's second commit was written to close on the main create path.

3. **`generatedAt` is required but never used.** It's a mandatory field in `createVisitScheduleBulkSchema`, but neither `visitSchedule.service.ts` nor the repository ever reads it after parsing, and `VisitSchedule` has no matching column in `schema.prisma`. Either it should be wired into something (an audit timestamp? drift detection vs `scheduledDate`?) or dropped from the contract — right now every client is forced to send a value the server silently discards.

4. **Unrelated third commit.** The PR is scoped to landing the visit-schedules-bulk cherry-pick, but the third commit (`10ca24f`) fixes a Para/Gravida validation gap in beneficiary-service — different service, different feature, not mentioned in the PR description. Worth confirming that's meant to ride along here rather than land as its own PR — bundling it makes rollback/bisect for "the visit-schedules-bulk landing" less clean.

palani-ND:
Follow-up on my review — a correction to finding #1 (idempotency replay doesn't check the payload matches).

On a closer look, that's not specific to this PR: `form.service.ts`'s `createSubmission` (`findSubmissionByLocalUuid`) and `visitInstance.service.ts`'s `create` (`findByLocalVisitUuid`) both do the exact same thing — look up by a bare local-UUID and return the existing row on any match, with no comparison against the replayed payload. `visitSchedule.service.ts` is consistent with that established convention, not introducing a new gap. So I'd drop #1 as a blocker for this PR specifically — if it's worth fixing, it's a cross-cutting concern across all three idempotent-by-local-uuid endpoints, not something to scope to this one.

Finding #2 (the idempotency lookup not scoped to the caller's beneficiary) still stands as-is — that one's a closer inconsistency within this same PR, since the second commit here explicitly added an ownership check for the create path but the idempotency lookup a few lines over doesn't carry the same scoping. Findings #3 (`generatedAt` unused) and #4 (unrelated third commit) are unchanged.

apps/visit-form-service/src/visit-schedules/visitSchedule.service.ts
Dharanishwaran-ND: Fixed in 107c8af — added `matchesStoredRow()`, which compares the replayed row's `visitCode`/`sequenceNo`/dates/`anchorType`/anchor-presence against the stored row before treating it as `alreadyExisted`. A same-`localScheduleUuid`-different-content replay now throws a 409 instead of silently returning the old stored data. Covered by the new "rejects with 409 when a replayed localScheduleUuid carries different content than what is stored" test.

apps/visit-form-service/src/visit-schedules/visitSchedule.repository.ts
Dharanishwaran-ND: Fixed in 107c8af — `findByLocalScheduleUuids` now takes `beneficiaryId` and scopes the query to it, so a colliding `localScheduleUuid` belonging to a different beneficiary is no longer visible to this lookup. It's then treated as a new row, and if it collides with the DB's own unique constraint on insert, that's caught and surfaced as a 409 rather than leaking the other beneficiary's data or 500ing. Covered by the new "rejects with 409 when a localScheduleUuid collides with another beneficiary's row" test.

apps/visit-form-service/src/visit-schedules/dto/create-visit-schedule-bulk.dto.ts
Dharanishwaran-ND: Fixed in 107c8af — dropped generatedAt from createVisitScheduleBulkSchema entirely, since it was never read after parsing and has no matching column in schema.prisma.

PR #119, opened 2026-08-06
Fix Mother Registration form schema

Dharanishwaran-ND:
## Summary
- Fix Mother Registration form schema per source doc
- Split age_of_the_beneficiary into DOB and age fields
- Correct enter_the_beneficiary_address input_type
- Add Td dose date fields and contains visibility operator
- Replace firstName/middleName/lastName with fullName
- Add numericRange to mother/child age fields

## Test plan
- [ ] Verify Mother Registration form renders with corrected schema
- [ ] Verify age/DOB fields validate against numericRange bounds

### DIFF
--- a/apps/visit-form-service/prisma/seed-data/child-registration.json
+++ b/apps/visit-form-service/prisma/seed-data/child-registration.json
@@ -162,19 +162,18 @@
       "question_code": "caregiver_name_first_name_middle_name_last_name"
     },
     {
-      "label": "Date of birth of the mother",
+      "label": "Mother date of birth",
       "section": "Personal Info",
       "required": false,
       "input_type": "date",
-      "question_code": "date_of_birth_of_mother"
+      "question_code": "mother_date_of_birth"
     },
     {
-      "label": "Age or DOB of the mother",
+      "label": "Age of the mother",
       "section": "Personal Info",
       "required": false,
       "input_type": "number",
-      "question_code": "age_or_dob_of_the_mother",
-      "computedFrom": "AGE_FROM_DOB",
+      "question_code": "mother_age",
       "numericRange": {
         "min": 10,
         "max": 50
@@ -954,7 +953,7 @@
   "validationJson": [
     {
       "rule": "ANY_OF_REQUIRED",
-      "fields": ["date_of_birth_of_mother", "age_or_dob_of_the_mother"]
+      "fields": ["mother_date_of_birth", "mother_age"]
     }
   ]
 }
--- a/apps/visit-form-service/prisma/seed-data/child-registration.json
+++ b/apps/visit-form-service/prisma/seed-data/child-registration.json
@@ -162,11 +162,19 @@
       "question_code": "caregiver_name_first_name_middle_name_last_name"
     },
     {
+      "label": "Date of birth of the mother",
+      "section": "Personal Info",
+      "required": false,
+      "input_type": "date",
+      "question_code": "date_of_birth_of_mother"
+    },
+    {
       "label": "Age or DOB of the mother",
       "section": "Personal Info",
-      "required": true,
-      "input_type": "date,integer",
+      "required": false,
+      "input_type": "number",
       "question_code": "age_or_dob_of_the_mother",
+      "computedFrom": "AGE_FROM_DOB",
       "numericRange": {
         "min": 10,
         "max": 50
@@ -943,5 +951,10 @@
       "question_code": "remarks"
     }
   ],
-  "validationJson": []
+  "validationJson": [
+    {
+      "rule": "ANY_OF_REQUIRED",
+      "fields": ["date_of_birth_of_mother", "age_or_dob_of_the_mother"]
+    }
+  ]
 }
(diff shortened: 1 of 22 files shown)


apps/visit-form-service/prisma/seed-data/child-registration.json
```diff
@@ -157,8 +165,12 @@
       "label": "Age or DOB of the mother",
       "section": "Personal Info",
       "required": true,
-      "input_type": "date",
-      "question_code": "age_or_dob_of_the_mother"
+      "input_type": "date,integer",
```
palani-ND: This is now `"date,integer"` — the same non-existent/unrecognized `input_type` value that commit `dc773e4` in this very PR identified as making a field "unusable" on the mobile client (per CR037), and fixed elsewhere by splitting into separate `date`/`number` fields with an `ANY_OF_REQUIRED` rule. This field is `required: true` here with no such split and no fallback, so as written it looks like it'll block child registration on-device rather than just being cosmetically wrong. Looks unintentional given the commit's stated scope was only "add numericRange" — probably meant to stay `"date"` (or get the same DOB/age split treatment as `age_of_the_beneficiary`).

palani-ND (commented):
Reviewed the schema/validation code (form-validation.ts, form-field.dto.ts, beneficiary.service.ts/dto.ts, beneficiary-duplicated-fields.ts) plus a full pass over both seed-data JSON files. The `contains` operator, `SUM_EQUALS` offset, `ANY_OF_REQUIRED`, and `exactLength` additions are all correctly wired and well-tested, and the `visibleWhen` field-code fix (`is_there_a_family_history_of_acute_or_chronic_illnesses_note`) does match the field's actual `question_code`, so that part of the diff checks out. Two things worth a look:

1. **The final commit reintroduces the exact bug commit #2 of this same PR just fixed, in a different form.** `child-registration.json`'s `age_or_dob_of_the_mother` field had its `input_type` changed from `"date"` to `"date,integer"` while adding its `numericRange`. Commit #2 (`dc773e4`) diagnosed this precise pattern — "the previous 'date,integer' input_type... is not a recognized client type and left the field unusable" — and fixed it for MOTHER_REGISTRATION's `age_of_the_beneficiary` by splitting it into two real fields. The identical string landed back in CHILD_REGISTRATION's required `age_or_dob_of_the_mother` field one commit later. Since `input_type` is validated server-side as a bare non-empty string (`form-field.dto.ts`), this passes seeding/validation cleanly and will only surface as a broken, unfillable required field on-device — this looks like it should get the same DOB/age split (or at least revert to a valid single type) rather than ship as-is.

2. **Four consent questions were dropped from MOTHER_REGISTRATION, not just reordered — worth confirming that's intentional.** The old Consent section had 7 fields including `willing_to_share_personal_info`, `willing_to_share_health_history`, `willing_to_participate_in_diagnostic_tests`, and `understands_the_logic_for_referral`. None of those 4 appear anywhere in the new schema (only as `-` lines in the diff, no matching `+`) — they were replaced by `edd`/`gestational_age_at_the_time_of_registration` (computed fields) plus the existing `did_we_receive_consent`/`consent_form_photo`/`consent_audio`. The new Consent section does land at exactly 7 fields, matching the PR's "Consent 1-7" framing, so this is plausibly a deliberate doc-driven redesign rather than an accidental loss — but given these are informed-consent tracking questions, it's worth an explicit confirmation that the source doc really does retire per-topic consent granularity in favor of the single `did_we_receive_consent` field, rather than this being a byproduct of the reorder.

Dharanishwaran-ND:
Addressed both points from the review — thanks for the thorough pass.

**1. `age_or_dob_of_the_mother` in child-registration.json** — confirmed and fixed in b1c12ab. This reintroduced the exact `"date,integer"` bug that dc773e4 fixed for MOTHER_REGISTRATION's `age_of_the_beneficiary`. Applied the same fix here: split into `date_of_birth_of_mother` (date) + `age_or_dob_of_the_mother` (number, `computedFrom: AGE_FROM_DOB`), enforced by the existing `ANY_OF_REQUIRED` cross-field rule (no engine changes needed — that rule and computedFrom type already exist from dc773e4). This also matches the source doc exactly: `docs/Revised_App_Form_Final_20.3.26.xlsx.md` line 412 says "Mandatory / Either date of birth or age should be filled" for this field.

**2. Dropped consent fields in MOTHER_REGISTRATION** — confirmed intentional, no change needed. Checked `docs/Revised_App_Form_Final_20.3.26.xlsx.md` lines 24-31 (MOTHER_REGISTRATION's CONSENT SECTION): the source doc only ever specifies 4 consent-related fields — Arogya Sakhi video, consent audio, "Did we receive consent?", and consent form photo. `willing_to_share_personal_info`, `willing_to_share_health_history`, `willing_to_participate_in_diagnostic_tests`, and `understands_the_logic_for_referral` never existed in the source doc, so 5f14808 removing them was correct — the doc retires per-topic consent granularity in favor of the single `did_we_receive_consent` field, as you suspected.

Tests and lint pass for visit-form-service.

PR #121, opened 2026-08-07
fix(supervisor-operations-service): Add validation gaps for events, transactions and call logs

Dharanishwaran-ND:
## Summary
Fixes 7 backend defects reported in API testing (`Arogya Sakhi API Observation.docx.md`), scoped to `supervisor-operations-service`:

- **Supervisor events**: reject a past `eventDate` on create/reschedule; reject `remarks` shorter than 3 characters; reject a duplicate event for the same supervisor/project at the exact same `eventDate` (409)
- **Inventory transactions**: reject a future `transactionDate` on both create and update
- **Call logs**: `GET /call-logs` now scopes results to the caller's own logs (MANAGER/ADMIN remain unrestricted) — previously returned every supervisor's logs
- **Call logs**: reject a future `callStartAt`/`callEndAt`, and reject `callEndAt` earlier than `callStartAt` on both create and update — this is the root cause of the reported `502`/`INTERNAL_ERROR` on `POST /call-logs`, which now returns a clean `400 VALIDATION_ERROR` for that payload

Items reported as `403 FORBIDDEN` on events/transactions/call-logs the tester "should have access to" were investigated and confirmed **working as intended** (SUPERVISOR-scoped ownership checks by design) — not changed here.

## Test plan
- [x] `nx test supervisor-operations-service` — 182/182 passing (30 new tests added across 6 new DTO spec files + service-layer additions)
- [x] `nx lint supervisor-operations-service` — clean
- [x] `nx build supervisor-operations-service` — compiles with no type errors
- [x] Regression-tested the exact reported `POST /call-logs` payload (`callEndAt` before `callStartAt`) — confirmed it now returns `400`, not `502`

### DIFF
--- a/apps/supervisor-operations-service/src/operations/dto/create-supervisorEvent.dto.ts
+++ b/apps/supervisor-operations-service/src/operations/dto/create-supervisorEvent.dto.ts
@@ -1,4 +1,5 @@
 import { z } from 'zod';
+import { diffInDays, startOfUTCDay } from '@armman/core';
 
 /**
  * Recursive JSON value type usable inside the `topicsJson` payload (nulls allowed
@@ -41,7 +42,11 @@ export const createSupervisorEventSchema = z
     projectId: z.string().uuid(),
     supervisorId: z.string().uuid(),
     eventType: z.enum(['MEETING', 'TRAINING']),
-    eventDate: z.coerce.date().refine((date) => date.getTime() >= Date.now(), {
+    // eventDate is @db.Date (calendar date, no time-of-day) — comparing its
+    // coerced midnight value against the exact current instant would reject
+    // today itself for all but the first millisecond after UTC midnight, so
+    // "now" is normalized to midnight UTC before comparing whole calendar days.
+    eventDate: z.coerce.date().refine((date) => diffInDays(startOfUTCDay(new Date()), date) >= 0, {
       message: 'eventDate must not be in the past.',
     }),
     topicsJson: jsonValueSchema,
--- a/apps/supervisor-operations-service/src/operations/dto/create-supervisorEvent.dto.ts
+++ b/apps/supervisor-operations-service/src/operations/dto/create-supervisorEvent.dto.ts
@@ -41,9 +41,11 @@ export const createSupervisorEventSchema = z
     projectId: z.string().uuid(),
     supervisorId: z.string().uuid(),
     eventType: z.enum(['MEETING', 'TRAINING']),
-    eventDate: z.coerce.date(),
+    eventDate: z.coerce.date().refine((date) => date.getTime() >= Date.now(), {
+      message: 'eventDate must not be in the past.',
+    }),
     topicsJson: jsonValueSchema,
-    remarks: z.string().trim().min(1).optional(),
+    remarks: z.string().trim().min(3).optional(),
     status: z.enum(['SCHEDULED', 'COMPLETED', 'CANCELLED']),
     photoMediaId: z.string().uuid().optional(),
   })
--- a/apps/supervisor-operations-service/src/operations/dto/reschedule-event.dto.ts
+++ b/apps/supervisor-operations-service/src/operations/dto/reschedule-event.dto.ts
@@ -8,8 +8,10 @@ import { z } from 'zod';
  */
 export const rescheduleEventSchema = z
   .object({
-    eventDate: z.coerce.date(),
-    remarks: z.string().trim().min(1).optional(),
+    eventDate: z.coerce.date().refine((date) => date.getTime() >= Date.now(), {
+      message: 'eventDate must not be in the past.',
+    }),
+    remarks: z.string().trim().min(3).optional(),
   })
   .strict();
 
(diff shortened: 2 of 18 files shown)


apps/supervisor-operations-service/src/operations/dto/create-supervisorEvent.dto.ts
```diff
@@ -41,9 +41,11 @@ export const createSupervisorEventSchema = z
     projectId: z.string().uuid(),
     supervisorId: z.string().uuid(),
     eventType: z.enum(['MEETING', 'TRAINING']),
-    eventDate: z.coerce.date(),
+    eventDate: z.coerce.date().refine((date) => date.getTime() >= Date.now(), {
```
palani-ND: `eventDate` is `@db.Date` (date-only) in `schema.prisma`, but this compares against `Date.now()` (the exact current instant). A date-only value for "today", coerced to today's UTC midnight, will be less than `Date.now()` at any point after midnight UTC — so this rejects a same-day event as "in the past" essentially all day, every day. Looks like this needs a calendar-date comparison (compare `date`'s day against today's day) rather than an instant comparison. Same issue in `reschedule-event.dto.ts`.

apps/supervisor-operations-service/src/operations/dto/reschedule-event.dto.ts
```diff
@@ -8,8 +8,10 @@ import { z } from 'zod';
  */
 export const rescheduleEventSchema = z
   .object({
-    eventDate: z.coerce.date(),
-    remarks: z.string().trim().min(1).optional(),
+    eventDate: z.coerce.date().refine((date) => date.getTime() >= Date.now(), {
```
palani-ND: Same instant-vs-date-only mismatch as `create-supervisorEvent.dto.ts`'s `eventDate` refine — see that comment. `eventDate` is a `@db.Date` column, so comparing its coerced midnight value against `Date.now()` rejects same-day reschedules almost all day, every day.

palani-ND (commented):
Reviewed the DTO/service changes across events, inventory transactions, and call logs. The call-log timestamp validation, the `GET /call-logs` ownership scoping, and the `transactionDate`-not-in-future check all look correct — I checked `transactionDate`'s comparison direction specifically against its `@db.Date` column type and it holds up (a date-only midnight value is never "in the future" once today has begun, so that check is safe). One thing looks like it'll actively break a common case, though:

**`eventDate` "must not be in the past" compares a date-only value against the exact current instant, so it rejects "today" almost all day, every day.** Both `create-supervisorEvent.dto.ts` and `reschedule-event.dto.ts` add `date.getTime() >= Date.now()`. `SupervisorEvent.eventDate` is `@db.Date` in `schema.prisma` — date-only, any time-of-day component gets truncated at storage regardless of what's sent. If a client submits `eventDate` for *today* (e.g. a date-only string, coerced by `z.coerce.date()` to today at UTC midnight — the natural representation for a date-only field, and the only one that survives DB storage unchanged), then `date.getTime()` is today's midnight, which is less than `Date.now()` at literally any point after midnight UTC. So this refine fails for "schedule/reschedule for today" essentially all day, every day — only a request landing in the first moment after UTC midnight would pass. Even in the more generous case where the client sends a full current timestamp instead of midnight, normal request latency means the captured client-side "now" is always a little behind the server's `Date.now()` at validation time, making even a genuine "right now" submission a coin-flip.

The new tests only exercise `±24h` (`Date.now() - 24h` / `Date.now() + 24h`), so they never hit the same-day boundary case that trips this up — worth adding a same-day case to the suite once this is fixed. The fix is presumably a calendar-date comparison (e.g. truncate both sides to a date before comparing) rather than an instant comparison, matching how `dateOnlySchema`-style date-only fields are handled elsewhere in this codebase (e.g. visit-form-service's `visit_schedules` date fields).

Smaller note: the new duplicate-event conflict check (`findConflictingEvent`, 409) is only wired into `createEvent`, not `rescheduleEvent` — `rescheduleEvent` doesn't call it, so rescheduling Event A onto the exact same `eventDate` as an already-existing Event B for the same supervisor/project wouldn't be caught, even though the reported issue and this PR's `eventDate`-in-the-past validation were both applied to reschedule as well as create. Might be intentional scope (the reported bug was specifically about creation), but worth confirming.

apps/supervisor-operations-service/src/operations/dto/create-supervisorEvent.dto.ts
Dharanishwaran-ND: Fixed in e5be464 — added `startOfUTCDay` to `@armman/core` and normalized `Date.now()` to midnight UTC before the `diffInDays` comparison, so this now compares calendar days instead of instants. Added a regression test ("accepts today as eventDate regardless of the current time of day") that constructs `eventDate` as today's date-only value and asserts it passes no matter what time the test runs — this would have caught the original bug.

apps/supervisor-operations-service/src/operations/dto/reschedule-event.dto.ts
Dharanishwaran-ND: Fixed in e5be464 — same fix as create-supervisorEvent.dto.ts: now compares calendar days via startOfUTCDay + diffInDays instead of instants. Added the matching regression test for same-day reschedules.

PR #122, opened 2026-08-07
Add bulk-upsert endpoint for lookup category values

Dharanishwaran-ND:
## Summary
- Adds `PATCH /lookups/:categoryCode/values/bulk-upsert` (ADMIN-only), so an environment whose lookup master data has drifted from the current form schema can be reconciled in one call instead of N manual `POST`/`PATCH` calls.
- Creates any `valueCode` missing from the category, updates `valueLabel`/`sortOrder`/`parentLookupValueId` on any that differ, and leaves every value not mentioned in the payload completely untouched — additive/updating only, never deletes or deactivates.
- Runs as a single DB transaction, so a failure partway through can't leave a category half-reconciled.
- Motivated by a real bug found on dev: `EDUCATION_LEVEL` had an outdated 6-value set instead of the current form's 8 values, silently dropping 5 of 8 possible education answers to `null`. Fixed by hand via the existing single-value endpoints during investigation — this endpoint makes that reconciliation a single scripted call going forward.

## Test plan
- [x] New `LookupService.bulkUpsertValues` unit tests (21 cases): create-only, no-op/unchanged, update-only, mixed create+update+unchanged in one call, `parentLookupValueId` validation (same-category success, cross-category 422 on both create and update paths), unknown category 404, never touches values absent from the payload, idempotency (same payload run twice → second run reports all `unchanged`).
- [x] `nx run auth-service:lint` — clean.
- [x] `nx test auth-service --testPathPattern=lookup.service.spec` — 21/21 passing.
- [x] Confirmed the 12 other failing tests in the full `auth-service` suite are pre-existing on `develop` (reproduced identically with this change stashed out) — not a regression from this PR.
- [ ] Live endpoint call against a running service — not exercised in this session (local dev server was running older code and wasn't restarted to avoid disrupting other in-progress work); recommend a smoke test after merge/deploy.

### DIFF
--- a/apps/auth-service/src/lookups/dto/bulk-upsert-lookup-values.dto.ts
+++ b/apps/auth-service/src/lookups/dto/bulk-upsert-lookup-values.dto.ts
@@ -1,0 +1,30 @@
+import { z } from 'zod';
+
+const bulkUpsertLookupValueItemSchema = z
+  .object({
+    valueCode: z.string().trim().min(1).max(80),
+    valueLabel: z.string().trim().min(1).max(160),
+    sortOrder: z.number().int().min(0).optional(),
+    parentLookupValueId: z.string().uuid().optional(),
+  })
+  .strict();
+
+/**
+ * Validation schema for reconciling a whole category's values in one call —
+ * e.g. an environment whose lookup master data has drifted from the current
+ * form schema. Creates any valueCode not already in the category, updates
+ * valueLabel/sortOrder/parentLookupValueId on any that exist and differ, and
+ * leaves every value not mentioned in the payload untouched (additive/
+ * updating only — this endpoint never deletes or deactivates a value).
+ */
+export const bulkUpsertLookupValuesSchema = z
+  .object({
+    values: z.array(bulkUpsertLookupValueItemSchema).min(1),
+  })
+  .strict()
+  .refine(
+    (data) => new Set(data.values.map((v) => v.valueCode)).size === data.values.length,
+    { message: 'valueCode must be unique within the payload.', path: ['values'] },
+  );
+
+export type BulkUpsertLookupValuesInput = z.infer<typeof bulkUpsertLookupValuesSchema>;
--- a/apps/auth-service/src/lookups/dto/bulk-upsert-lookup-values.dto.ts
+++ b/apps/auth-service/src/lookups/dto/bulk-upsert-lookup-values.dto.ts
@@ -5,7 +5,12 @@ const bulkUpsertLookupValueItemSchema = z
     valueCode: z.string().trim().min(1).max(80),
     valueLabel: z.string().trim().min(1).max(160),
     sortOrder: z.number().int().min(0).optional(),
-    parentLookupValueId: z.string().uuid().optional(),
+    // Omitted = leave the existing parent alone (create path: no parent);
+    // null = explicitly clear an existing value's parent — same
+    // omitted-vs-null distinction as update-lookup-value.dto.ts, needed here
+    // too since "this value should no longer be a child of anything" is a
+    // legitimate reconciliation case.
+    parentLookupValueId: z.string().uuid().nullable().optional(),
   })
   .strict();
 
--- a/apps/auth-service/src/lookups/lookup.service.ts
+++ b/apps/auth-service/src/lookups/lookup.service.ts
@@ -2,6 +2,7 @@ import { conflict, notFound, unprocessable } from '@armman/service-commons';
 import type { LookupRepository } from './lookup.repository';
 import type { CreateLookupValueInput } from './dto/create-lookup-value.dto';
 import type { UpdateLookupValueInput } from './dto/update-lookup-value.dto';
+import type { BulkUpsertLookupValuesInput } from './dto/bulk-upsert-lookup-values.dto';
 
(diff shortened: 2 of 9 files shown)


apps/auth-service/src/lookups/lookup.service.ts
```diff
@@ -87,6 +88,71 @@ export class LookupService {
...
+
+      if (isUnchanged) {
+        unchanged.push(item.valueCode);
+      } else {
+        toUpdate.push({ id: existing.id, data: item });
+      }
+    }
+
+    if (toCreate.length > 0 || toUpdate.length > 0) {
+      await this.repository.bulkUpsertValues(category.id, toCreate, toUpdate);
```
palani-ND: Unlike `createValue` just above (which wraps its `repository.createValue` call in try/catch and turns a P2002 into a `409`), this has no equivalent handling. If a concurrent request creates one of these `valueCode`s in the window between the `findValuesByCategoryId` read above and this transaction, the `@@unique([lookupCategoryId, valueCode])` constraint would throw a P2002 here that surfaces as an unhandled 500 rather than a clean error, for the same class of race the single-value path already guards against.

apps/auth-service/src/lookups/dto/bulk-upsert-lookup-values.dto.ts
```diff
@@ -0,0 +1,30 @@
+import { z } from 'zod';
+
+const bulkUpsertLookupValueItemSchema = z
+  .object({
+    valueCode: z.string().trim().min(1).max(80),
+    valueLabel: z.string().trim().min(1).max(160),
+    sortOrder: z.number().int().min(0).optional(),
+    parentLookupValueId: z.string().uuid().optional(),
```
palani-ND: This is `.optional()` only — `update-lookup-value.dto.ts`'s equivalent field is `.nullable()` too, specifically so a caller can explicitly clear an existing value's parent. As written there's no way to unset a `parentLookupValueId` through this bulk endpoint (only set it, or leave it alone by omitting it) — worth confirming that's intentional for a reconciliation endpoint, since "this value should no longer be a child of anything" seems like a plausible reconciliation case.

palani-ND (commented):
Small, well-tested PR — the create/update/unchanged split, the same-category `parentLookupValueId` validation, the all-or-nothing transaction, and the idempotency test all check out. Two things worth a look:

1. **The bulk transaction can raise an unhandled 500 on a unique-constraint race, unlike the single-value path in this same file.** `createValue` wraps `repository.createValue` in a try/catch and turns a P2002 into a clean `409 conflict(...)` — the exact scenario guarded against is a concurrent create landing on the same `(lookupCategoryId, valueCode)` pair, which is a real `@@unique` constraint in `schema.prisma`. `bulkUpsertValues` computes `toCreate`/`toUpdate` from a `findValuesByCategoryId` snapshot and then calls `repository.bulkUpsertValues(...)` with no equivalent try/catch — if another request creates one of the same `valueCode`s in the window between that snapshot read and this transaction's write, the whole batch throws an unhandled P2002 instead of a clean error. Given this is ADMIN-only and presumably low-concurrency in practice, the odds of hitting it are low, but since the exact handling already exists one method up in this same file, it seems worth reusing here too rather than leaving this one path inconsistent.

2. **This endpoint can't clear an existing value's `parentLookupValueId`, unlike the single-value update endpoint.** `bulk-upsert-lookup-values.dto.ts` declares `parentLookupValueId: z.string().uuid().optional()`, but `update-lookup-value.dto.ts` (the existing single-value PATCH) declares it `.nullable()` as well, specifically to support unsetting a value's parent. Since the bulk schema has no `.nullable()`, there's no way to send an explicit "clear this parent" through this endpoint — omitting the field just leaves the existing parent untouched (correct for "don't mention it, don't change it"), but there's no way to say "unset it" short of falling back to the single-value endpoint. Probably fine given the endpoint's stated additive/reconciliation purpose, but worth confirming that's an acceptable gap rather than an oversight, since a hierarchy restructuring (a value that should no longer be a child of anything) is a plausible reconciliation scenario too.

apps/auth-service/src/lookups/lookup.service.ts
Dharanishwaran-ND: Fixed in a4d81a6 — wrapped the `repository.bulkUpsertValues` transaction call in try/catch, same pattern as `createValue`, so a P2002 from a concurrent create winning that race now maps to a 409 instead of an unhandled 500. Added a regression test ("throws 409 when a concurrent create wins the race on a duplicate valueCode").

apps/auth-service/src/lookups/dto/bulk-upsert-lookup-values.dto.ts
Dharanishwaran-ND: Good catch, and confirmed intentional to support — fixed in a4d81a6. Made `parentLookupValueId` nullable (not just optional) on the item schema, same distinction as `update-lookup-value.dto.ts`: omitted leaves an existing value's parent untouched, explicit `null` clears it. The service's unchanged-detection and the repository's update payload already handled `null` correctly once the DTO allowed it through — added tests for both the clear-parent case and the already-parentless-plus-null-is-unchanged case.

PR #123, opened 2026-08-07
fix(visit-form-service): Close validation gaps on CHILD_REGISTRATION/ANC_VISIT, fix met-beneficiary skip logic

Dharanishwaran-ND:
## Summary
- Adds `REQUIRED_IF_SELECTED` cross-field rule and closes CHILD_REGISTRATION vaccination/date validation gaps (dateRule, pattern, exactLength) per the Infant Registration form spec.
- Closes ANC_VISIT validation gaps: LMP-edit/latest-visit/USG date rules, urine-test exclusivity.
- Fixes a bug where "Have you been able to meet the beneficiary?" = No still demanded every other field on the form, blocking submission for a visit that never happened.
- **Hotfix**: reverts an array-form `visibleWhen` shape that broke ANC_VISIT/INFANT_VISIT form loading on mobile (`FormVisibilityEvaluator` only parses a single condition object) — collapses back to single-condition, keeping the met-beneficiary gate.

## Commits
- `83c9892` Close CHILD_REGISTRATION validation gaps
- `4bc4404` Enforce date-mandatory-if-selected on vaccine/Td-dose checkboxes
- `f087b95` Close ANC_VISIT validation gaps
- `7d99520` Remove unconfirmed Td-dose fields from ANC_VISIT
- `0f7aed6` Stop demanding rest of form when beneficiary not met
- `431a15f` Revert array-form visibleWhen (mobile crash hotfix)

## Test plan
- [x] `nx test visit-form-service` — 220 tests pass
- [x] Live-verified via real API against local dev: published new CHILD_REGISTRATION/ANC_VISIT/INFANT_VISIT/MOTHER_REGISTRATION form versions and submitted deliberately-broken and fully-valid payloads to confirm each rule fires/clears correctly
- [x] Confirmed no form (ANC_VISIT/INFANT_VISIT/MOTHER_REGISTRATION/CHILD_REGISTRATION) has an array-form `visibleWhen` in its currently active version

## Follow-ups
- Mobile needs to update `FormVisibilityEvaluator` to accept `visibleWhen` as either a single object or an array of ANDed conditions before the dropped secondary conditions (gestational-age gating, sonography-confirmed, USG-done, per-vaccine gating on ~39 fields) can be restored.
- Q26 (Haemoglobin ±2 g/dl vs last visit) and the LMP-edit registration-date window remain unimplemented — both need data outside a single submission's own formData (prior visit / beneficiary case), which `validateSubmission()` doesn't currently have access to.

---

## Quick Response: implement remaining card decisions + security fix

Quick Response previously only implemented the REOPEN card decision (and a
`supervisorId` trust bug in supervisor-operations-service); every other
card type fell through to a generic 501. This adds the rest and fixes a
privilege-escalation bug found by automated review along the way.

### Summary
- **service-commons**: 501 error messages now pass through verbatim instead of being masked as a generic "Something went wrong" — 501s are a deliberate, safe API-contract statement, not leaked internals.
- **auth-service**: adds `PATCH /users/:id/reactivate`, a narrow status-only endpoint (SUPERVISOR/MANAGER/ADMIN), separate from the ADMIN-only general `PATCH /users/:id`.
- **beneficiary-service**: adds `PATCH /beneficiaries/:id/lmp` (applies an approved LMP change, recomputing eddDate server-side) and `PATCH /beneficiaries/:id/reactivate` (reopens a CLOSED case, recording the transition in status history).
- **closure-reopen-service**: an APPROVED reopen decision now actually reactivates the beneficiary case via the new beneficiary-service endpoint — previously it updated `reopen_requests` but had no real-world effect. Not tolerated like audit/notification: a failed reactivation surfaces as an error.
- **risk-referral-service**: adds `PATCH /referrals/:id/decision` (LAPSE/REFILL/COMPLETE), backing REFERRAL_INCOMPLETE.
- **incentive-wages-service**:
  - adds `GET /incentive-rates/active` so callers can discover the rate to reference.
  - **security fix**: `POST /incentives` was widened to SUPERVISOR without re-deriving the payout amount server-side, letting any Supervisor mint an incentive event with an arbitrary `amountInr` (privilege escalation, flagged by automated review). Removed `amountInr` from the request schema entirely; the amount is now always looked up from the referenced `rateId`.
- **approval-service**: wires up the remaining Quick Response dispatch branches (LMP_CHANGE, CLOSURE_REVIEW, REFERRAL_INCOMPLETE, ACCOMPANIED_REFERRAL, DATA_RESTORE), each via a new gateway HTTP client.
  - **DATA_RESTORE** deliberately does NOT implement the SRS's literal FR-SV-4.6 wording (device-level data restore) — that was explicitly flagged in the SRS as pending confirmation with ARMMAN. Implements a narrower, explicitly client-approved behaviour instead: approving the card reactivates the requesting Sakhi's own user account.

### Commits
- `feea442` fix(service-commons): pass 501 messages through instead of masking as generic errors
- `4d3ed50` feat(auth-service): add narrow user-reactivation endpoint for DATA_RESTORE decisions
- `382a019` feat(beneficiary-service): add LMP-change apply and case-reactivation endpoints
- `a2956a1` feat(closure-reopen-service): reactivate the beneficiary case when a reopen request is approved
- `ee404f3` feat(risk-referral-service): add referral decision endpoint (LAPSE/REFILL/COMPLETE)
- `dafd1a8` feat(incentive-wages-service): add GET /incentive-rates/active lookup endpoint
- `8382ad9` fix(incentive-wages-service): re-derive incentive amount server-side instead of trusting the caller
- `693ae19` feat(approval-service): wire up remaining Quick Response card decisions

### Test plan
- [x] `npx nx run-many -t test -p approval-service closure-reopen-service risk-referral-service incentive-wages-service auth-service beneficiary-service` — all 6 projects pass
- [x] Live end-to-end verified for DATA_RESTORE: deactivate user → login blocked (401) → create card → approve → 200 → login succeeds again
- [x] Live end-to-end verified for LMP_CHANGE against local dev DB (LMP changed, eddDate recomputed correctly)
- [ ] CLOSURE_REVIEW / REFERRAL_INCOMPLETE / ACCOMPANIED_REFERRAL are unit-tested but not yet live-verified — the closure-reopen-service and risk-referral-service Prisma migrations need to be re-applied against the correct per-service DB schema (a prior `migrate deploy` run picked up the root `.env`'s `DATABASE_URL`, which lacks a `?schema=` param and landed tables in `public` instead of each service's own schema)

### Follow-ups
- Re-run the closure-reopen-service and risk-referral-service Prisma migrations with the correct per-service `DATABASE_URL` (with `?schema=<service-name>`), then live-verify CLOSURE_REVIEW/REFERRAL_INCOMPLETE/ACCOMPANIED_REFERRAL end-to-end.

### DIFF
--- a/apps/approval-service/src/quick-response/quick-response.service.ts
+++ b/apps/approval-service/src/quick-response/quick-response.service.ts
@@ -1,4 +1,4 @@
-import { badRequest, notFound, unprocessable, HttpError } from '@armman/service-commons';
+import { badRequest, conflict, notFound, unprocessable, HttpError } from '@armman/service-commons';
 import type { QuickResponseRepository } from './quick-response.repository';
 import type { LookupClient } from './lookup.client';
 import type { EscalationClient, EscalationCard } from './escalation.client';
@@ -150,11 +150,16 @@ export class QuickResponseService {
    * write endpoint in a service that doesn't have one yet, so they respond
    * 501 rather than silently no-op or guess at undefined behavior.
    */
-  async decide(cardId: string, dto: DecideQuickResponseInput, authorizationHeader: string) {
+  async decide(
+    cardId: string,
+    dto: DecideQuickResponseInput,
+    decidedByUserId: string,
+    authorizationHeader: string,
+  ) {
     if (dto.cardSource === 'escalation_events') {
       return this.decideEscalationCard(cardId, dto, authorizationHeader);
     }
-    return this.decideApprovalRequestCard(cardId, dto, authorizationHeader);
+    return this.decideApprovalRequestCard(cardId, dto, decidedByUserId, authorizationHeader);
   }
 
   private async decideEscalationCard(
@@ -186,65 +191,109 @@ export class QuickResponseService {
   private async decideApprovalRequestCard(
     cardId: string,
     dto: DecideQuickResponseInput,
+    decidedByUserId: string,
     authorizationHeader: string,
   ) {
     const existing = await this.repository.findById(cardId);
     if (!existing) throw notFound('Quick Response card not found.');
 
+    // Only LMP_CHANGE has no downstream state of its own to lean on for
+    // idempotency (unlike CLOSURE_REVIEW/REFERRAL_INCOMPLETE/ACCOMPANIED_
+    // REFERRAL/REOPEN, each guarded by the target service's own
+    // PENDING-only status check) — so re-approving an already-decided
+    // LMP_CHANGE card silently re-applied the LMP/EDD write and re-notified
+    // the Sakhi, contradicting this endpoint's documented "409: Card
+    // already decided" contract. Checked here, once, for every card type
+    // rather than duplicated per handler.
+    if (existing.decidedAt) {
+      throw conflict('This Quick Response card has already been decided.');
+    }
+
+    let result: Record<string, unknown>;
     if (existing.requestType === 'LMP_CHANGE') {
-      return this.decideLmpChangeCard(cardId, existing, dto, authorizationHeader);
-    }
+      result = await this.decideLmpChangeCard(cardId, existing, dto, authorizationHeader);
+    } else if (existing.requestType === 'CLOSURE_REVIEW') {
+      result = await this.decideClosureReviewCard(cardId, existing, dto, authorizationHeader);
+    } else if (existing.requestType === 'REFERRAL_INCOMPLETE') {
+      result = await this.decideReferralIncompleteCard(cardId, existing, dto, authorizationHeader);
+    } else if (existing.requestType === 'ACCOMPANIED_REFERRAL') {
+      result = await this.decideAccompaniedReferralCard(cardId, existing, dto, authorizationHeader);
+    } else if (existing.requestType === 'DATA_RESTORE') {
+      result = await this.decideDataRestoreCard(cardId, existing, dto, authorizationHeader);
+    } else if (existing.requestType === 'REOPEN') {
+      if (dto.decision !== 'APPROVE' && dto.decision !== 'REJECT') {
+        throw badRequest('decision: Must be APPROVE or REJECT for a REOPEN card.');
+      }
+      if (!existing.reopenRequestId) {
+        // Data integrity issue, not a client error — a REOPEN
+        // approval_requests row must always carry the reopen_requests id it
+        // originated from.
+        throw new HttpError(500, 'This REOPEN card has no linked reopen request.');
+      }
 
-    if (existing.requestType === 'CLOSURE_REVIEW') {
-      return this.decideClosureReviewCard(cardId, existing, dto, authorizationHeader);
-    }
+      // The audit_log entry and Sakhi notification are written by
+      // closure-reopen-service's own decide flow, not here — that's the only
+      // place a reopen decision is actually persisted, so the audit trail
+      // can't be bypassed by calling that endpoint directly instead of through
+      // Quick Response.
+      const decided = await this.reopenRequestClient.decide(
+        existing.reopenRequestId,
+        dto.decision === 'APPROVE' ? 'APPROVED' : 'REJECTED',
+        dto.decisionReasonCodeLookupId,
+        dto.decisionNotes,
+        authorizationHeader,
+      );
 
-    if (existing.requestType === 'REFERRAL_INCOMPLETE') {
-      return this.decideReferralIncompleteCard(cardId, existing, dto, authorizationHeader);
-    }
-
-    if (existing.requestType === 'ACCOMPANIED_REFERRAL') {
-      return this.decideAccompaniedReferralCard(cardId, existing, dto, authorizationHeader);
-    }
-
-    if (existing.requestType === 'DATA_RESTORE') {
-      return this.decideDataRestoreCard(cardId, existing, dto, authorizationHeader);
-    }
-
-    if (existing.requestType !== 'REOPEN') {
+      result = {
+        cardId,
+        cardSource: 'approval_requests' as const,
+        decision: dto.decision,
+        reopenRequest: decided,
+      };
+    } else {
       throw new HttpError(
         501,
         `Decisions on "${existing.requestType}" cards are not yet implemented.`,
       );
     }
-    if (dto.decision !== 'APPROVE' && dto.decision !== 'REJECT') {
-      throw badRequest('decision: Must be APPROVE or REJECT for a REOPEN card.');
-    }
-    if (!existing.reopenRequestId) {
-      // Data integrity issue, not a client error — a REOPEN approval_requests
-      // row must always carry the reopen_requests id it originated from.
-      throw new HttpError(500, 'This REOPEN card has no linked reopen request.');
+
+    // Marked decided only after the real side effect above has already
+    // succeeded — a failure to persist this marker must not be reported as
+    // if the decision itself failed (log-and-continue, like every other
+    // post-commit call in this method), but it also must never run before
+    // the side effect, or a card could be marked decided while its actual
+    // effect never happened.
+    try {
+      // Every approval_requests decision reaching this point is APPROVE or
+      // REJECT — the OKAY decision only exists on the escalation_events
+      // branch, which returns earlier in decide() and never reaches here.
+      const decisionStatusLookupId = await this.lookupClient.resolveApprovalStatusId(
+        dto.decision === 'APPROVE' ? 'APPROVED' : 'REJECTED',
+        authorizationHeader,
+      );
+      if (decisionStatusLookupId) {
+        await this.repository.markDecided(
+          cardId,
+          decisionStatusLookupId,
+          decidedByUserId,
+          dto.decisionNotes,
+          dto.decisionReasonCodeLookupId,
+        );
+      } else {
+        console.error(
+          `Quick Response card ${cardId} was decided (${dto.decision}) but no ` +
+            `${dto.decision === 'APPROVE' ? 'APPROVED' : 'REJECTED'} APPROVAL_STATUS lookup ` +
+            'value was found — the card cannot be marked decided and remains re-decidable.',
+        );
+      }
+    } catch (err) {
+      console.error(
+        `Quick Response card ${cardId} was decided (${dto.decision}) but marking it decided failed:`,
+        err,
+      );
     }
 
-    // The audit_log entry and Sakhi notification are written by
-    // closure-reopen-service's own decide flow, not here — that's the only
-    // place a reopen decision is actually persisted, so the audit trail
-    // can't be bypassed by calling that endpoint directly instead of through
-    // Quick Response.
-    const decided = await this.reopenRequestClient.decide(
-      existing.reopenRequestId,
-      dto.decision === 'APPROVE' ? 'APPROVED' : 'REJECTED',
-      dto.decisionReasonCodeLookupId,
-      dto.decisionNotes,
-      authorizationHeader,
-    );
-
-    return {
-      cardId,
-      cardSource: 'approval_requests' as const,
-      decision: dto.decision,
-      reopenRequest: decided,
-    };
+    return result;
   }
 
   /**
@@ -411,11 +460,22 @@ export class QuickResponseService {
    * Decides an ACCOMPANIED_REFERRAL card (FR-SV-4.9). Approve marks the
    * referral Completed, then resolves the assigned Sakhi (via the referral's
    * beneficiary — neither approval_requests nor referrals carries a sakhiId
-   * column) and triggers the incentive. The referral decision and incentive
-   * trigger are NOT tolerated — a Supervisor approving this card needs both
-   * to actually happen, not silently fail while looking successful. Reject
-   * makes no referral-side call at all (the referral stays Pending, per
-   * spec) and grants no incentive. Either way the Sakhi is notified,
+   * column) and triggers the incentive. The referral decision itself is NOT
+   * tolerated — a Supervisor approving this card needs the referral to
+   * actually complete, not silently fail while looking successful.
+   *
+   * The incentive trigger, unlike the referral decision, IS best-effort: by
+   * the time it runs, risk-referral-service has already committed the
+   * referral to COMPLETED — a one-shot, PENDING_FOLLOWUP-only transition
+   * with no way back. If the incentive call were allowed to fail the whole
+   * request, there would be no way to retry just this step (any retry
+   * immediately 409s on the already-terminal referral), permanently
+   * dropping a payout instead of just needing a manual follow-up. Logged,
+   * not thrown — same log-and-continue shape as the Sakhi notification
+   * below, not a "this must succeed" call like the referral decision above.
+   *
+   * Reject makes no referral-side call at all (the referral stays Pending,
+   * per spec) and grants no incentive. Either way the Sakhi is notified,
    * best-effort.
    */
   private async decideAccompaniedReferralCard(
@@ -451,11 +511,20 @@ export class QuickResponseService {
       if (!beneficiary) {
         throw notFound('The beneficiary linked to this referral was not found.');
       }
-      await this.incentiveClient.triggerAccompaniedReferral(
-        beneficiary.sakhiId,
-        existing.referralId,
-        authorizationHeader,
-      );
+      try {
+        await this.incentiveClient.triggerAccompaniedReferral(
+          beneficiary.sakhiId,
+          existing.referralId,
+          authorizationHeader,
+        );
+      } catch (err) {
+        console.error(
+          `ACCOMPANIED_REFERRAL card ${cardId} was approved and the referral marked COMPLETED, ` +
+            `but the incentive trigger failed (referral cannot be re-decided to retry — ` +
+            `needs manual follow-up):`,
+          err,
+        );
+      }
     }
 
     try {
--- a/apps/approval-service/src/quick-response/quick-response.service.ts
+++ b/apps/approval-service/src/quick-response/quick-response.service.ts
@@ -1,12 +1,28 @@
-import { badRequest, notFound, HttpError } from '@armman/service-commons';
+import { badRequest, notFound, unprocessable, HttpError } from '@armman/service-commons';
 import type { QuickResponseRepository } from './quick-response.repository';
 import type { LookupClient } from './lookup.client';
 import type { EscalationClient, EscalationCard } from './escalation.client';
 import type { ReopenRequestClient } from './reopen-request.client';
+import type { BeneficiaryClient } from './beneficiary.client';
+import type { NotificationClient } from './notification.client';
+import type { ClosureClient } from './closure.client';
+import type { ReferralClient } from './referral.client';
+import type { IncentiveClient } from './incentive.client';
+import type { UserClient } from './user.client';
 import type { ListQuickResponseInput } from './dto/list-quick-response.dto';
 import type { DecideQuickResponseInput } from './dto/decide-quick-response.dto';
 
-/** Every ApprovalRequestType maps 1:1 to a Quick Response card type — same name. */
+/**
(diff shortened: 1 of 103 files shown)


palani-ND (commented):
This PR bundles two large, unrelated efforts (visit-form-service validation gaps, and finishing the Quick Response feature across 6 services) — reviewed both. Given the size I split the review across parallel passes and then independently re-verified the highest-severity findings myself by reading the actual code. Grouped by area, most severe first:

### incentive-wages-service — the flagged security fix is real but incomplete

The literal bug (`amountInr` trusted from the client) is genuinely closed — `createIncentiveEventSchema` has no `amountInr` field, and `IncentiveEventService.create` always re-derives it from `rateId` server-side. But `IncentiveRateRepository.findById` — the only check applied before trusting `rate.amountInr` — filters on nothing but `isDeleted: false`. It never checks that `rate.rateType` matches `dto.sourceEntityType`, never checks `effectiveFrom`/`effectiveTo`, never checks `geographyUnitId` scope. Since `GET /incentive-rates/active` is open to SUPERVISOR and lets a caller enumerate any active rate's id across all types/geographies, a Supervisor can submit `sourceEntityType: 'VISIT'` with a `rateId` that actually belongs to a `RETAINER` (or any other, possibly much higher) rate and get paid that amount instead. The fix narrows "arbitrary amount" down to "any real rate's amount regardless of whether it's the correct one for the claimed event" — still an unauthorized-payout bug, just one extra self-service lookup away. `incentiveEvent.service.spec.ts` only tests with a rate that already matches by construction, so this gap isn't covered by the new tests either.

### risk-referral-service — `PATCH /referrals/:id/decision` has no ownership scoping at all

The route handler is `const updated = await service.decide(req.params.id, req.body);` — the caller's identity (`req.user`) is never passed to the service at all, and `referral.service.ts`'s `decide()` takes no caller parameter. Only `requireRoles('SUPERVISOR', 'MANAGER', 'ADMIN')` gates the route. Compare this to the two other new mutations in this same PR — `beneficiary-service`'s `applyLmpChange`/`reactivateCase`, both of which call `assertCallerCanTouchCase` against the caller's own Sakhi roster, and `auth-service`'s `reactivateUser`, which does the equivalent roster check. Any authenticated Supervisor/Manager/Admin — regardless of project or whether they supervise the Sakhi who logged the referral — can decide any referral system-wide. Same gap applies to `closure-reopen-service`'s closure-decision route (also `requireRoles` only, no roster check in `closure.service.ts`'s `decide()`), so REFERRAL_INCOMPLETE, ACCOMPANIED_REFERRAL, and CLOSURE_REVIEW are all cross-team IDOR-reachable, while LMP_CHANGE and DATA_RESTORE (which do check the roster downstream) aren't. That's a real, live authorization gap now that these 3 card types have real side effects.

### ACCOMPANIED_REFERRAL — a new instance of the exact bug class this feature has already been burned by once

`decideAccompaniedReferralCard` calls `await this.referralClient.decide(existing.referralId, 'COMPLETE', authorizationHeader)` (commits the referral to `COMPLETED` — a one-shot transition, `risk-referral-service` only allows deciding from `PENDING_FOLLOWUP`), then, with no try/catch, `await this.incentiveClient.triggerAccompaniedReferral(...)`. If the incentive trigger fails for any reason (no active rate configured, the beneficiary lookup 404s, incentive-wages-service errors), the whole request fails — but the referral is already `COMPLETED` and permanently un-decidable again (any retry immediately 409s before ever reaching the incentive trigger). Unlike the original audit/notify bug this PR's REOPEN flow already fixed (where a failed side effect at least left the door open for a clean retry), here there's no recovery path at all: a real payout is silently and permanently dropped, with the workflow state claiming success. Worth wrapping the incentive trigger the same way the earlier REOPEN fix wrapped audit/notify — log-and-continue rather than fail the whole request after the referral is already committed — or, if the incentive absolutely must happen, doing it before the referral's terminal-state transition so a failure can still be retried.

### closure-reopen-service — reopen-request reactivation can leave a stuck, contradictory state

`reopen-request.service.ts`'s `decide()` calls `repository.decide()` (commits `supervisorStatus = 'APPROVED'`) and then, unguarded, `await this.beneficiaryClient.reactivateCase(existing.beneficiaryId, authorizationHeader)`. If that call fails, the request errors out correctly (per the PR description, this one actually isn't tolerated like audit/notify) — but there's no way to retry just the reactivation: any subsequent call to `decide()` hits the `existing.supervisorStatus !== 'PENDING'` guard and 409s immediately. Net effect: `reopen_requests.supervisor_status = APPROVED` forever, while the beneficiary case stays `CLOSED`, with no self-service recovery path — a transient beneficiary-service outage during this window permanently stalls the beneficiary in a closed state despite an approved reopen. Same shape as the ACCOMPANIED_REFERRAL issue above: a terminal-state transition committed before an un-retriable follow-on call.

### LMP_CHANGE has no idempotency protection

`beneficiary-service`'s `applyLmpChange` recomputes and writes `lmpDate`/`eddDate` unconditionally on every call — there's no "already applied" check anywhere in the chain (the `approval_requests` row itself is never marked decided; nothing in `QuickResponseRepository` writes `decisionStatusLookupId`/`decidedAt` after creation). Re-submitting `POST /quick-response/:cardId/decision` with `APPROVE` on an already-approved LMP_CHANGE card silently 200s and re-applies the LMP write (and re-notifies the Sakhi) again, contradicting the controller's own documented `409: Card already decided`. Relatedly, since the underlying approval_requests row is never marked decided, replaying with the *opposite* decision (REJECT after an already-applied APPROVE, or vice versa for DATA_RESTORE) hits a decision branch that does nothing and returns a clean 200, silently misrepresenting a card that was already actioned differently.

### visit-form-service — the array-`visibleWhen` "revert" didn't actually revert the schema

Independently confirmed the PR's specific data claims are true: no seed-data file (mother/child-registration, anc-visit, infant-visit) currently has an array-shaped `visibleWhen`, and the "beneficiary not met" gating fix is broad and correctly implemented (verified against ANC_VISIT/INFANT_VISIT's own field lists, not just spot-checked). But `form-field.dto.ts`'s schema — `visibleWhen: z.union([visibleWhenConditionSchema, z.array(visibleWhenConditionSchema).min(1)]).optional()` — still accepts the array shape that caused the mobile crash this PR is hot-fixing, and this schema is live-wired into the admin form-authoring `PATCH` endpoint that publishes new form versions in production. The only things preventing a repeat of this exact incident are two data-file-specific unit test assertions scoped to `anc-visit.json`/`infant-visit.json` only — there's no equivalent test for mother/child-registration, and zero protection on the live admin-authoring path itself. Given the PR's own commit message for the hotfix apparently already acknowledges this ("that capability isn't removed, only unused in published content"), this seems like a known, accepted gap rather than an oversight — but it's worth being explicit that this is a live landmine for the next form edit, not a closed loop, and the Follow-ups section's framing ("mobile needs to update FormVisibilityEvaluator... before the dropped conditions can be restored") undersells that the backend itself would happily accept and publish the breaking shape again in the meantime.

### What's solid

- The original PR #113 audit/notify uncaught-failure bug is genuinely fixed globally for REOPEN, LMP_CHANGE, REFERRAL_INCOMPLETE, and DATA_RESTORE's own notify calls — all wrapped in try/catch, verified by reading each branch.
- `auth-service`'s `PATCH /users/:id/reactivate` and `beneficiary-service`'s `applyLmpChange`/`reactivateCase` are both correctly roster-scoped, correctly guard against re-triggering on an already-active/already-open record, and use atomic conditional updates against races. DATA_RESTORE's reactivation target is derived from the persisted `approval_requests.requestedByUserId`, never from the client-supplied decide body — confirmed not client-manipulable.
- `REQUIRED_IF_SELECTED` and the `SUM_EQUALS`/`LTE`/`ANY_OF_REQUIRED` cross-field rules are correctly implemented, and every rule's field references across all 4 forms resolve to real question_codes (no silently-dead rules).
- The `service-commons` 501-passthrough change is narrowly and correctly scoped — 500/502/503/504 remain masked and logged as before, the classification is type-based (`instanceof HttpError`, not a coerced status code), the only two 501 call sites in the whole monorepo use static enum-constrained strings, and there's an explicit regression test asserting 500s stay masked.
- All 6 new/modified Quick Response HTTP clients correctly route through the gateway with the caller's forwarded Authorization header — no direct-service-port regressions.

apps/incentive-wages-service/src/rates/incentiveRate.repository.ts
```diff
@@ -0,0 +1,45 @@
...
+/** Data access for incentive rates. Owns only this service's `incentive_rates` table. */
+export class IncentiveRateRepository {
+  constructor(private readonly prisma: PrismaService) {}
+
+  /**
+   * Looks up a rate by id — used by IncentiveEventService.create to
+   * re-derive amountInr server-side rather than trust a client-supplied
+   * value (see createIncentiveEventSchema's doc comment).
+   */
+  findById(id: string) {
```
palani-ND, line 12: This is the only check standing between `rateId` and a trusted payout amount (see `incentiveEvent.service.ts`'s `create()`), and it only filters on `isDeleted: false`. It doesn't check that the rate's `rateType` matches the incentive event's `sourceEntityType`, doesn't check `effectiveFrom`/`effectiveTo`, and doesn't check `geographyUnitId` scope. Since `GET /incentive-rates/active` lets a SUPERVISOR discover any active rate's id, a caller can pair a mismatched-but-real `rateId` with an unrelated `sourceEntityType` and get that rate's amount — the fix closes "arbitrary amount" but not "any real rate's amount, correct or not."

apps/risk-referral-service/src/referrals/referral.controller.ts
```diff
@@ -113,5 +123,30 @@ export function createReferralRouter(service: ReferralService) {
...
+        404: { description: 'Referral not found', schema: apiErrorSchema },
+        409: { description: 'Referral is not in PENDING_FOLLOWUP status', schema: apiErrorSchema },
+      },
+    },
+    trustGatewayIdentity,
+    requireRoles('SUPERVISOR', 'MANAGER', 'ADMIN'),
+    validate(referralIdParamsSchema, 'params'),
+    validateBody(decideReferralRequestSchema),
+    asyncHandler(async (req, res) => {
+      const updated = await service.decide(req.params.id, req.body);
```
palani-ND: `req.user` is never passed into `service.decide()` here, and `referral.service.ts`'s `decide()` takes no caller parameter at all — only the route's `requireRoles('SUPERVISOR', 'MANAGER', 'ADMIN')` gates this. Unlike the sibling endpoints added in this same PR (`beneficiary-service`'s `applyLmpChange`/`reactivateCase`, `auth-service`'s `reactivateUser`, all of which check the caller's own Sakhi roster), any authenticated Supervisor/Manager/Admin can decide any referral system-wide regardless of project or whether they supervise the Sakhi who raised it.

apps/approval-service/src/quick-response/quick-response.service.ts
```diff
@@ -204,4 +246,291 @@ export class QuickResponseService {
...
+        throw new HttpError(500, 'This ACCOMPANIED_REFERRAL card has no linked beneficiary.');
+      }
+      const beneficiary = await this.beneficiaryClient.getById(
+        existing.beneficiaryId,
+        authorizationHeader,
+      );
+      if (!beneficiary) {
+        throw notFound('The beneficiary linked to this referral was not found.');
+      }
+      await this.incentiveClient.triggerAccompaniedReferral(
```
palani-ND: By this point `referralClient.decide(existing.referralId, 'COMPLETE', ...)` (line 442) has already committed the referral to `COMPLETED` — a one-shot transition per `risk-referral-service`'s `PENDING_FOLLOWUP`-only guard. If this incentive trigger throws (no active rate, beneficiary lookup 404, incentive-wages-service error), the request fails but the referral can never be decided again to retry just this step — any retry immediately 409s on the already-terminal referral status. That's a real payout silently and permanently dropped, worse than the original audit/notify bug this feature already fixed elsewhere (which at least left room for a clean retry). Worth wrapping this the same way REOPEN's audit/notify calls were wrapped — log-and-continue rather than fail the whole request after the referral is already committed.

apps/closure-reopen-service/src/reopen-requests/reopen-request.service.ts
```diff
@@ -43,6 +106,10 @@ export class ReopenRequestService {
       throw conflict('This reopen request has already been decided.');
     }
 
+    if (dto.decision === 'APPROVED') {
+      await this.beneficiaryClient.reactivateCase(existing.beneficiaryId, authorizationHeader);
```
palani-ND: `repository.decide()` just above already committed `supervisorStatus = 'APPROVED'`. If this reactivation call fails, there's no way to retry just this step — any subsequent `decide()` call hits the `existing.supervisorStatus !== 'PENDING'` guard at the top of this method and 409s before ever reaching this line again. A transient `beneficiary-service` outage here permanently strands the beneficiary in `CLOSED` with the reopen request forever showing `APPROVED` — same failure shape as the ACCOMPANIED_REFERRAL issue elsewhere in this PR.

apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
```diff
@@ -213,6 +349,75 @@ export class BeneficiaryService {
...
+   * schedule-generation logic to trigger. Re-syncing/regenerating the
+   * schedule after an LMP change is the Sakhi app's responsibility, not
+   * this endpoint's — a known, accepted gap, not an oversight.
+   *
+   * A SUPERVISOR caller may only apply this to a case in their own Sakhi
+   * roster (same scoping as list()) — this endpoint is reachable by a human
+   * Supervisor role, not just server-to-server, so it needs the same IDOR
+   * guard as any other single-case mutation.
+   */
+  async applyLmpChange(
```
palani-ND, line 370: No check here (or anywhere upstream — the `approval_requests` row backing an LMP_CHANGE card is never marked decided after creation) that this LMP change hasn't already been applied. Re-approving an already-approved LMP_CHANGE Quick Response card silently 200s and re-applies the LMP/EDD write and re-notifies the Sakhi, contradicting the Quick Response controller's own documented `409: Card already decided` contract.

apps/visit-form-service/src/forms/dto/form-field.dto.ts
```diff
@@ -56,18 +74,13 @@ export const formFieldSchema = z
     // Fixed digit-count fields (e.g. a 10-digit mobile number) — distinct from
     // numericRange, which bounds a value rather than its digit length.
     exactLength: z.number().int().positive().optional(),
+    // A single condition, or an array of conditions ANDed together (all must
+    // pass for the field to be visible) — e.g. a visit-form field that is
+    // both gated behind "met the beneficiary = yes" AND its own condition
+    // (gestational age, USG done, etc). Most fields only ever need one
+    // condition, so the plain single-object shape is still accepted as-is.
     visibleWhen: z
```
palani-ND: This still accepts the array shape that caused the mobile crash this PR is hot-fixing around, and it's live-wired into the admin form-authoring PATCH endpoint that publishes form versions in production — not just this PR's own seed data. The only guardrails against a repeat are the seed-data tests scoped to `anc-visit.json`/`infant-visit.json`; there's no equivalent test for mother/child-registration, and no protection at all on the live authoring path. If this is intentionally left permissive for a future multi-condition feature (as the neighboring comment suggests), it's probably worth saying so explicitly in the Follow-ups section, since right now it reads like mobile is the only thing standing between this and a repeat incident.

apps/incentive-wages-service/src/rates/incentiveRate.repository.ts
Dharanishwaran-ND: Fixed in 418dff7 + b61dbef.

418dff7 adds the rateType/sourceEntityType check and an effective-window check. b61dbef then corrects the window check after our automated security reviewer caught that I'd anchored it to `dto.calculatedAt` — entirely caller-supplied, so a client could submit any `calculatedAt` to make an expired/not-yet-effective rate pass. Now anchored to the server's own clock instead; the one real caller (approval-service's `incentive.client.ts`) already sends the actual current time, so this changes nothing for it.

geographyUnitId scope is still open — flagged as a known gap in the new doc comment, not fixed here since it needs a caller-geography param that doesn't exist on this endpoint yet (a larger change than this validation tightening).

apps/risk-referral-service/src/referrals/referral.controller.ts
Dharanishwaran-ND: Fixed in 54cab85.

decide() now takes the caller and authorization header. Added a BeneficiaryClient (referrals carries no sakhiId — resolved via the referral's beneficiaryId) and a duplicated listSakhiIdsForSupervisor (this service has no shared client to import it from, per the forklift rule). A SUPERVISOR caller's roster is checked against the resolved beneficiary's sakhiId before any status transition; MANAGER/ADMIN remain unscoped. Same pattern beneficiary-service's applyLmpChange/reactivateCase and auth-service's reactivateUser already use.

apps/approval-service/src/quick-response/quick-response.service.ts
Dharanishwaran-ND: Fixed in 57991b9.

Agreed on the risk — by the time the incentive trigger runs, the referral is already committed to COMPLETED (one-shot, PENDING_FOLLOWUP-only, no way back), so a hard failure there permanently drops the payout with no retry path, which is strictly worse than the notify-failure case this same file already tolerates. Wrapped it the same way: log-and-continue instead of throw.

Same commit also closes the LMP_CHANGE re-decision gap you flagged separately below — both are decision-robustness issues in the same dispatch method, so I landed them together.

apps/closure-reopen-service/src/reopen-requests/reopen-request.service.ts
Dharanishwaran-ND: Fixed in 793e241.

You're right that the doc comment's "fixable via retry" claim was wrong — decide()'s own PENDING-only guard means a second call always 409s once supervisorStatus is APPROVED, so there was never an actual retry path. Wrapped the reactivation call in try/catch, log-and-continue like the audit/notification calls beside it, with a distinct log line calling out that this one needs manual follow-up (not just a shrug) since it's a real data-integrity gap, not a cosmetic one. Corrected the doc comment too.

apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
Dharanishwaran-ND: Fixed in 57991b9.

Added approval_requests.decidedAt as the guard, checked once in decideApprovalRequestCard for every card type (not just LMP_CHANGE — CLOSURE_REVIEW/REFERRAL_INCOMPLETE/ACCOMPANIED_REFERRAL/REOPEN all had the same gap in principle, they just happened to be covered incidentally by their downstream service's own PENDING-only status guard; LMP_CHANGE was the one with nothing downstream to lean on). Added a repository.markDecided call, invoked only after each handler's real side effect has already succeeded, log-and-continue if the marker write itself fails. Threaded decidedByUserId through from the controller's req.user.id, which wasn't reaching the service before either.

apps/visit-form-service/src/forms/dto/form-field.dto.ts
Dharanishwaran-ND: Fixed in 40de159 — tightened rather than documented, since nothing currently depends on the array shape being accepted (checked every seed file, including MOTHER_REGISTRATION/CHILD_REGISTRATION as you noted — zero usage anywhere).

visibleWhen is now a single {field,operator,value} condition only, at the schema level, so the live admin form-authoring PATCH endpoint can't republish the crash on any form either. Added a form-field.dto.spec.ts asserting the array shape (including single-element arrays) is rejected. Also removed the now-unreachable array-handling branch from form-validation.ts's isVisible and the describe block that tested it, since a FormField satisfying the type can no longer carry an array — keeping tested-but-dead code around didn't seem worth it. Re-allowing the array shape is still possible later, just needs to happen deliberately once mobile ships a parser for it, not silently.

PR #124, opened 2026-08-08
fix(media-service): Allow SUPERVISOR to register a media asset via POST /media

Dharanishwaran-ND:
## Summary
- Allow `SUPERVISOR` (in addition to `SAKHI`) to call `POST /media`, closing the last gap in the event-photo flow: a supervisor could already get a presigned upload URL and PUT the photo to S3, but could never register it against their event.
- Fix a `noImplicitAny` build failure in `OperationsService.upsertTopicMark` by explicitly typing the `gt` callback parameter.

## Test plan
- [ ] `npm run lint && npm run affected:test`
- [ ] Manually verify a SUPERVISOR token can call `POST /media` end-to-end (presigned URL -> S3 PUT -> POST /media) and get a 201
- [ ] Confirm SAKHI behavior is unchanged

### DIFF
--- a/apps/media-service/src/media/mediaAsset.controller.ts
+++ b/apps/media-service/src/media/mediaAsset.controller.ts
@@ -136,7 +136,7 @@ export function createMediaAssetRouter(service: MediaAssetService) {
       },
     },
     trustGatewayIdentity,
-    requireRoles('SAKHI'),
+    requireRoles('SAKHI', 'SUPERVISOR'),
     validateBody(createMediaAssetSchema),
     asyncHandler(async (req, res) => {
       const created = await service.create(req.body);
--- a/apps/supervisor-operations-service/src/operations/operations.service.ts
+++ b/apps/supervisor-operations-service/src/operations/operations.service.ts
@@ -530,7 +530,9 @@ export class OperationsService {
     }
 
     const gatheringTopics = await this.repository.findGatheringTopics(dto.gatheringId);
-    const belongsToGathering = gatheringTopics.some((gt) => gt.topicId === topicId);
+    const belongsToGathering = gatheringTopics.some(
+      (gt: (typeof gatheringTopics)[number]) => gt.topicId === topicId,
+    );
     if (!belongsToGathering) {
       throw unprocessable('topicId: this topic is not part of the referenced gathering.');
     }


palani-ND (commented):
Small PR, both changes checked out:

- The `noImplicitAny` fix is correct and trivial — `(typeof gatheringTopics)[number]` correctly extracts `findGatheringTopics`'s element type (a Prisma `findMany` result array), and doesn't change behavior.
- The `SUPERVISOR` role addition to `POST /media` is a reasonable, narrowly-described fix for the stated gap (a Supervisor completing an event's presigned upload had no way to register it).

One thing worth a second look before merging, though not something this diff introduces — it just now applies to twice as many callers: `mediaAsset.service.ts`'s `create()` is a pure passthrough (`return this.repository.create(dto)`), and `createMediaAssetSchema` has no `.strict()`-enforced ownership binding — `uploadedByUserId` is client-suppliable and optional (not derived from the caller's own JWT identity), and `beneficiaryId`/`visitId`/`eventId`/`referralId`/`followupId`/`linkedEntityType`+`linkedEntityId`/`storageUri` are all accepted with no check that the calling SAKHI or (now) SUPERVISOR actually owns or is otherwise authorized to attach media to the referenced entity. So today, any authenticated caller with either role can register a metadata row claiming an arbitrary `storageUri` belongs to an arbitrary beneficiary/visit/event/referral, or attribute the upload to a different `uploadedByUserId` entirely. This predates this PR (SAKHI already had the same unrestricted access), but since this PR is specifically about widening who can call the endpoint, it seemed worth flagging now rather than assuming it's already been considered — deriving `uploadedByUserId` server-side from the caller's identity and validating the linked-entity ownership (mirroring the roster/ownership checks this session has seen added elsewhere in the codebase, e.g. `assertCallerCanTouchCase` in beneficiary-service) would close it, but that's clearly bigger than this PR's scope.

Dharanishwaran-ND:
Thanks for the review, @palani-ND — agreed on all counts.

Confirming the ownership-validation gap you flagged: it's pre-existing (SAKHI already had the same unrestricted access before this PR), and you're right that it's a materially bigger change than this one-line role addition — `createMediaAssetSchema` would need to derive `uploadedByUserId` server-side from the caller's JWT rather than accepting it from the body, and each linked-entity field (`beneficiaryId`/`visitId`/`eventId`/`referralId`/`followupId`) would need its own ownership/roster check against the caller, similar to `assertCallerCanTouchCase` in beneficiary-service. That's a cross-cutting change to `mediaAsset.service.ts` and its schema, not something to fold into this PR silently.

I'll file a follow-up issue to track closing this gap properly rather than let it sit unrecorded. Given this PR itself is scoped to the SUPERVISOR role addition (not introducing the ownership gap), proposing we merge this as-is and track the fix separately — let me know if you'd rather block on it here.

Dharanishwaran-ND:
Filed as #126 to track separately.

PR #125, opened 2026-08-08
fix: Gateway CORS preflight headers and SAKHI 502 on GET /beneficiaries

Dharanishwaran-ND:
## Summary
- **api-gateway**: CORS preflight responses only ever sent a bare 204 with no `Access-Control-Allow-Methods`/`-Headers`, so browsers rejected any non-simple authenticated request (JSON body, `Authorization`, `x-request-id`) from an otherwise-allowed origin before it ever reached the API. `buildCorsMiddleware` now echoes back the browser's requested method/headers (or a sane default) and adds `Access-Control-Max-Age`.
- **beneficiary-service**: `GET /beneficiaries` 502'd for every SAKHI caller. `enrichListPage`'s per-id name fallback called `GET /sakhis/:sakhiId` (SUPERVISOR/MANAGER/ADMIN-only) for any id not already in the supervisor roster — which for a SAKHI's own list is always their own id, since `list()` forces `filters.sakhiId = caller.id`. The 403 from that scoped route surfaced as a 502 to the SAKHI. `list()` now skips the fallback lookup entirely when the caller is a SAKHI.

## Test plan
- [x] `npx nx affected --target=lint --base=origin/develop`
- [x] `npx nx affected --target=test --base=origin/develop` (191 tests passed across api-gateway and beneficiary-service)
- [ ] Manually verify a SAKHI token can call `GET /beneficiaries` without a 502
- [ ] Manually verify an authenticated JSON request (Authorization + Content-Type) from an allowed web origin now completes without a CORS preflight failure

### DIFF
--- a/apps/api-gateway/src/app.module.ts
+++ b/apps/api-gateway/src/app.module.ts
@@ -13,6 +13,7 @@ import { createHealthRouter } from './health/health.controller';
 import { createInfoRouter } from './info/info.controller';
 import { createDocsRouter } from './docs/docs.controller';
 import { registerProxies } from './proxy/register-proxies';
+import { buildCorsMiddleware } from './cors/cors-middleware';
 
 export {
   asyncHandler,
@@ -39,18 +40,7 @@ export function createApp(signer: Pick<TokenSigner, 'verify'>): Application {
 
   app.use(pinoHttp(buildLoggerOptions(appConfig.LOG_LEVEL)));
   app.use(helmet());
-  app.use((req, res, next) => {
-    const origin = req.header('origin');
-    if (origin && appConfig.CORS_ORIGINS.includes(origin)) {
-      res.setHeader('Access-Control-Allow-Origin', origin);
-      res.setHeader('Access-Control-Allow-Credentials', 'true');
-    }
-    if (req.method === 'OPTIONS') {
-      res.sendStatus(204);
-      return;
-    }
-    next();
-  });
+  app.use(buildCorsMiddleware(appConfig.CORS_ORIGINS));
   app.use(requestId);
 
   registerProxies(app, signer);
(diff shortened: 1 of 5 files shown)


palani-ND (commented):
Both fixes are correct and well-targeted:

- **beneficiary-service**: the `skipSakhiNameLookup` fix is sound. `req.user`/`caller` (populated by `trustGatewayIdentity`) only ever carries `id`/`roles`/`projectId`/`geographyUnitId` — no display name — so `sakhiName: null` for a SAKHI's own rows is the only option without adding a new lookup, and it correctly avoids the 403-turned-502. Reasonable trade-off given the client already knows its own logged-in user's name.
- **api-gateway**: echoing the specific allowed origin (never `*`) alongside `Access-Control-Allow-Credentials: true`, and reflecting the browser's actual `Access-Control-Request-Method`/`-Headers` on preflight rather than a fixed list, is the correct fix for the described failure and matches standard CORS practice. Good test coverage, including the disallowed-origin cases.

One small thing worth considering, not blocking: there's no `Vary: Origin` header on the response. Since `Access-Control-Allow-Origin` is echoed conditionally per request (varies by the `Origin` header), the standard recommendation is to add `Vary: Origin` (and for the preflight response specifically, `Access-Control-Request-Method`/`Access-Control-Request-Headers` too) so any intermediary cache in front of this gateway can't serve a response computed for one origin to a different one. This matters a bit more now than before, since this PR also introduces a 24-hour `Access-Control-Max-Age` — if anything between the browser and this service ever caches at the HTTP layer (a CDN, a caching reverse proxy), a missing `Vary` widens the window for a stale cross-origin response to be served. Cheap to add if there's any caching layer in the deployment path; skip it if there genuinely isn't one.

Dharanishwaran-ND:
Thanks @palani-ND — good catch, especially given the 24h Access-Control-Max-Age this PR adds. Added in b5ab141:

- \`Vary: Origin\` on every response (Allow-Origin is echoed per-request).
- On preflight specifically, \`Vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers\`, since Allow-Methods/-Headers are echoed from those.

Added two test cases covering both. Lint + full \`cors-middleware.spec.ts\` suite (9 tests) pass.

PR #127, opened 2026-08-08
feat: Risk-grading write pipeline, visit summary/update, and beneficiary summary widgets

Dharanishwaran-ND:
## Summary

Closes the gap identified in the ngrok endpoint audit — 5 previously-missing endpoints built, 1 existing endpoint extended, plus one unrelated bugfix picked up along the way:

- **rules-service**: `POST /rules/:setId/evaluate` — first real implementation of "Central GoRules execution" (`@gorules/zen-engine`). Only ever evaluates a PUBLISHED rule pack, never DRAFT.
- **risk-referral-service**: `POST /risk-assessments` — the missing risk-grading write pipeline. Resolves the rule set, calls rules-service to evaluate, persists the `RiskAssessment` row (unique per `submissionId`), and updates the beneficiary's risk-condition-summary rollup.
- **api-gateway**: new `/risk-assessments` proxy route (internal-use, mirrors `/admin/forms`/`/admin/rules`).
- **beneficiary-service**:
  - `GET /beneficiaries/registration-summary` and `GET /beneficiaries/risk-summary` dashboard widgets.
  - `PATCH /beneficiaries/:id/risk-condition-summary` — SAKHI-only rollup upsert, called from the visit-form-service → risk-referral-service → here chain.
  - Extends `GET /beneficiaries` with `registeredFrom`/`registeredTo` as aliases for the existing `fromDate`/`toDate` filter.
- **visit-form-service**:
  - `PATCH /visits/:id` — update visit status, role-scoped (SAKHI own visit, SUPERVISOR own roster, MANAGER/ADMIN unrestricted); auto-sets `completedAt` on COMPLETED.
  - `GET /visits/visit-summary` — status-count widget, same role scoping.
  - Wires `form.service.ts` to trigger the risk-grading pipeline after a visit-linked submission is persisted, forwarding the originating SAKHI's own token (no machine identity exists in this codebase).
- **auth-service** (unrelated bugfix found and fixed alongside this work): `PATCH /users/:id` assigning a SAKHI's `projectId` and creating their Sakhi profile in the same call previously failed with "SAKHI role has no project assigned" — the profile-creation check read the pre-update row instead of this same request's `input.projectId`.

## Test plan
- [x] `npx nx affected --target=lint --base=origin/develop` — 19 projects, all clean
- [x] `npx nx test auth-service` — 66/66 passed
- [x] `npx nx test beneficiary-service` — 79/79 passed
- [x] `npx nx test risk-referral-service` — 35/35 passed
- [x] `npx nx test rules-service` — 26/26 passed
- [x] `npx nx test visit-form-service` — 243/243 passed
- [ ] Manual end-to-end: submit a visit-linked form → confirm risk-assessment is recorded and the beneficiary's risk-condition-summary rollup updates
- [ ] Manual: `PATCH /visits/:id` and both new summary widgets against a live SAKHI/SUPERVISOR/MANAGER token

## Migrations
- `beneficiary-service`: `20260808000000_add_risk_condition_summary_grade_rank`
- `risk-referral-service`: `20260808000000_add_risk_assessment_submission_unique`
- `visit-form-service`: `20260808000001_add_form_definition_risk_rule_set_id`

### DIFF
--- a/apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
+++ b/apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
@@ -239,6 +239,7 @@ export class BeneficiaryRepository {
         sourceRuleVersionId: data.ruleVersionId,
       },
       update: {
+        phase: data.phase as never,
         latestGrade: data.grade,
         latestGradeRank: data.gradeRank,
         latestObservedValueJson: data.observedValueJson ?? undefined,
--- a/apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
+++ b/apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
@@ -1,8 +1,10 @@
 import type { PrismaService } from '../prisma/prisma.service';
 import type {
   BeneficiaryListFilters,
+  BeneficiarySummaryFilters,
   CreateEnrollmentInput,
   DuplicateSearchTokens,
+  UpsertRiskConditionSummaryData,
 } from './beneficiary.repository.types';
 
 // Re-exported so existing importers of `./beneficiary.repository` keep working;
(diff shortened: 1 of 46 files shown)


apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
```diff
@@ -113,6 +115,154 @@ export class BeneficiaryRepository {
...
+        everHighestObservedValueJson: data.observedValueJson ?? undefined,
+        everHighestVisitId: data.visitId,
+        everHighestSubmissionId: data.submissionId,
+        everHighestAssessedAt: data.assessedAt,
+        everAtRiskFlag: data.grade !== 'NORMAL',
+        currentReferralTriggerFlag: data.isReferralTrigger,
+        currentHrVisitTriggerFlag: data.isHrVisitTrigger,
+        sourceRuleVersionId: data.ruleVersionId,
+      },
+      update: {
```
palani-ND, line 241: `phase` is set in the `create:` block above (`phase: data.phase as never`) but missing here — every other field is written on both branches. Since a beneficiary can legitimately be reassessed for the same `riskConditionId` in a later phase (the unique key is `(beneficiaryId, riskConditionId)` only, no phase component), this permanently freezes `phase` at whatever it was on the very first insert, even though the caller sends the correct current phase on every subsequent PATCH.

apps/visit-form-service/src/visits/visitInstance.controller.ts
```diff
@@ -97,6 +108,35 @@ export function createVisitInstanceRouter(service: VisitInstanceService) {
...
+      responses: {
+        200: { description: 'Visit status counts', schema: envelope(visitSummarySchema) },
+        400: errorResponse(400, { message: 'fromDate must be on or before toDate.' }),
+        401: errorResponse(401),
+        403: errorResponse(403, { message: "sakhiId is not in this Supervisor's roster." }),
+        500: errorResponse(500),
+      },
+    },
+    trustGatewayIdentity,
+    requireRoles('SAKHI', 'SUPERVISOR', 'MANAGER'),
```
palani-ND: This route's own summary text says "MANAGER/ADMIN unscoped," and `isPrivileged()` in `visitInstance.service.ts` treats ADMIN as unrestricted — but `'ADMIN'` isn't in this role list, so an ADMIN-only token 403s here before ever reaching that logic. Same issue at line 186 (`PATCH /visits/:id`).

apps/rules-service/src/rules/ruleVersion.service.ts
```diff
@@ -61,6 +63,29 @@ export class RuleVersionService {
...
+   * describes, previously unimplemented (this service only stored rule
+   * packs opaquely; nothing interpreted rulesJson until now).
+   *
+   * 404 if the set has no published version — never silently evaluates
+   * against a DRAFT (an unreviewed rule pack must not affect real risk
+   * grading). Returns the published version's own id as `ruleVersionId`
+   * alongside the results, so the caller (risk-referral-service) can record
+   * which rule version produced a given RiskAssessment.
+   */
+  async evaluate(ruleSetId: string, dto: EvaluateRuleSetInput) {
```
palani-ND, line 82: This method's own doc comment says "404 if the set has no published version," but it throws `unprocessable(...)` (422) instead — and unlike `getPublished()` just above, it never checks whether the rule set exists at all first, so a nonexistent/deleted `ruleSetId` and one that just has nothing published yet get the identical 422 response.

apps/risk-referral-service/src/risk-assessments/riskAssessment.controller.ts
```diff
@@ -0,0 +1,130 @@
...
+          schema: apiErrorSchema,
+        },
+        502: {
+          description: 'rules-service or beneficiary-service unreachable',
+          schema: apiErrorSchema,
+        },
+      },
+    },
+    trustGatewayIdentity,
+    requireRoles('SAKHI'),
```
palani-ND: The service this calls into (`riskAssessment.service.ts`) has full, tested SUPERVISOR-roster and MANAGER/ADMIN-unrestricted branches, none of which are reachable through this route today since only SAKHI passes this gate. If that broader access was intended (e.g. an admin tool re-triggering grading), this needs updating; if SAKHI-only really is the intended final state, the unreachable branches in the service are worth trimming so the code doesn't imply a capability that doesn't exist.

palani-ND (commented):
This builds out the whole risk-grading pipeline end to end (rules-service evaluation, risk-referral-service's write path, visit-form-service's trigger, beneficiary-service's rollup/widgets) plus an unrelated auth-service fix. Split the review across all five areas and personally verified the findings below against the actual code. Most severe first:

### beneficiary-service — a beneficiary's `phase` freezes at whatever it was on first insert

`upsertRiskConditionSummary`'s `create:` block sets `phase: data.phase as never`, but the `update:` block has no `phase` field at all — every other field (`grade`, `gradeRank`, `observedValueJson`, `visitId`, `submissionId`, `assessedAt`, both trigger flags) is written on both branches, `phase` is the sole outlier. `phase` is a required field on every `PATCH /beneficiaries/:id/risk-condition-summary` request, and `SUMMARY_PHASES` mirrors the beneficiary's own journey (REGISTRATION → ANC → DELIVERY → PP → NN → INFANT_FOLLOWUP → CLOSURE) — the same `riskConditionId` plausibly gets reassessed in a later phase as she progresses. Since the row's unique key is `(beneficiaryId, riskConditionId)` only, a second call for the same condition in a new phase correctly updates the grade/assessed-at fields but leaves `phase` stuck at the original value forever, so `GET /beneficiaries/:id`'s `riskConditionSummaries[].phase` silently reports a phase the beneficiary is no longer in. No repository-level test exists to catch this — `beneficiary.service.spec.ts`'s "passes through all fields" test only asserts against a mocked repository call, never the real Prisma `upsert` object.

### visit-form-service — ADMIN is locked out of both new endpoints, contradicting their own documented behavior

Both `GET /visits/visit-summary` and `PATCH /visits/:id` gate with `requireRoles('SAKHI', 'SUPERVISOR', 'MANAGER')` — `'ADMIN'` is missing from both. But `isPrivileged()` in `visitInstance.service.ts` treats MANAGER *and* ADMIN as unrestricted, and both routes' own OpenAPI summary text explicitly says "MANAGER/ADMIN unrestricted"/"MANAGER/ADMIN unscoped." An ADMIN-only token gets a bare 403 at the gateway/role-check layer before ever reaching the service logic that was written to allow it. Every other place in this codebase that documents MANAGER/ADMIN access includes both in `requireRoles` (`beneficiary.controller.ts`, `incentiveRate.controller.ts`, `closure.controller.ts`, even auth-service's own sakhi endpoint this PR's `sakhi.client.ts` calls) — this looks like a plain typo/omission rather than a deliberate scoping decision. No controller-level test exists for either route (the specs call the service directly, bypassing `requireRoles` entirely), so this is invisible to the suite.

### rules-service — `evaluate()`'s error contract contradicts both its own doc comment and its sibling endpoint

`evaluate()`'s doc comment says "404 if the set has no published version," but the code throws `unprocessable(...)` — a 422, not a 404. Compare `getPublished()` two methods up in the same file, which correctly checks `findSetById` first (404 "Rule set not found" if the set itself doesn't exist) *before* separately checking for a published version (404 "No published rule pack version" if it exists but nothing's published) — two genuinely different conditions, both reported at least as 404s. `evaluate()` skips the set-existence check entirely and always returns the same 422 for both "this rule set doesn't exist" and "this rule set exists but has nothing published," and does so at a different status code than its sibling uses for the equivalent case. Downstream, `risk-referral-service`'s `ruleSet.client.ts` re-throws this 422 as "needs publishing," so a stale/typo'd `FormDefinition.riskRuleSetId` pointing at a rule set that was deleted or never existed gets misreported as a publishing gap rather than a configuration error.

### risk-referral-service — the route's role gate doesn't match what the service was built for

`POST /risk-assessments` is `requireRoles('SAKHI')` only, with a comment explaining this codebase has no machine identity so the caller is always a forwarded SAKHI token from the visit-form-service chain — a reasonable rationale for *this specific* internal caller. But `riskAssessment.service.ts`'s own doc comment and code spell out a full four-role authorization matrix ("a SAKHI may only act on her own beneficiary, a SUPERVISOR only on a beneficiary whose Sakhi is in her roster, MANAGER/ADMIN unrestricted"), with real branches for all three non-SAKHI cases — none of which are reachable through the actual HTTP route today. The existing "allows a MANAGER/ADMIN caller" test calls `service.create()` directly, bypassing the route, so it doesn't actually prove the real API accepts those callers — it doesn't. Worth clarifying whether SUPERVISOR/MANAGER/ADMIN direct access (e.g. for an admin tool to manually re-trigger grading) was intended and the route just hasn't caught up, or whether the service-layer branches are speculative and should be trimmed to match the SAKHI-only reality.

### Smaller items

- **`ruleSet.evaluator.ts`'s trigger-flag coercion** (`isReferralTrigger`/`isEducationTrigger`/`isHrVisitTrigger`) uses a bare `Boolean(e.field)`, while `riskConditionId`/`grade`/`gradeRank` just above it get explicit `typeof` guards that throw on a type mismatch. Since a GoRules decision graph is authored visually (not in code), an output column that happens to emit the literal string `"false"` would have `Boolean("false")` evaluate to `true`, silently flipping a trigger on. Given the asymmetry with the stricter checks right next to it, this doesn't look deliberate — worth the same explicit validation the other fields get.
- **`ruleSet.evaluator.spec.ts`** only exercises an "identity graph" that reflects its input straight back as output — it verifies the zen-engine wiring and the shape-validation logic, but zero real GoRules decision-table/branching behavior, and never touches an actual risk-grading `rulesJson` artifact. Since this is the first real use of zen-engine in the codebase, a genuine mismatch between what `answers` risk-referral-service sends and what a real decision graph expects wouldn't be caught by anything in the current suite.
- **Minor OpenAPI-doc/behavior mismatches** in `riskAssessment.controller.ts`: it documents a `502` for beneficiary-service unreachability that can never actually happen (that call's failures are deliberately swallowed and logged, not thrown — a documented, tested tradeoff, not a bug), and documents a `200` for the idempotent-replay case that the handler never returns (it always does `res.status(201)` regardless of whether the row was just created or already existed).

Everything else held up well: the SAKHI/SUPERVISOR/MANAGER/ADMIN scoping on the new beneficiary-service summary endpoints is consistently applied and matches the existing list endpoint; the `registeredFrom`/`registeredTo` aliasing has sensible, documented precedence; the risk-condition-summary rollup upsert is a genuine composite-key upsert with no duplicate-row risk; the auth-service SAKHI-profile fix genuinely reads the current request's `projectId` and correctly guards against an unrelated role's `projectId` leaking in, with both the happy path and the safeguard tested; the risk-grading trigger in `form.service.ts` fires strictly after the submission is durably persisted and is correctly fire-and-forget (though untested); the `PATCH /visits/:id` SUPERVISOR-roster check re-derives the visit's own Sakhi server-side rather than trusting anything client-supplied; `completedAt` is recomputed correctly on every status change; and the visit-summary widget correctly filters on `VisitSchedule.scheduledDate` rather than repeating this codebase's earlier completedAt/scheduledDate mix-up.

apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
Dharanishwaran-ND: Confirmed and fixed in bd5db8e — added `phase: data.phase as never` to the `update:` block, matching `create:` exactly. Verified the beneficiary-service test suite (179 tests) still passes.

apps/visit-form-service/src/visits/visitInstance.controller.ts
Dharanishwaran-ND: Confirmed and fixed in 9f513a3 — added 'ADMIN' to requireRoles on both PATCH /visits/:id and GET /visits/visit-summary. Verified visit-form-service's test suite (243 tests) still passes.

apps/rules-service/src/rules/ruleVersion.service.ts
Dharanishwaran-ND: Confirmed and fixed in 4baa061 — evaluate() now mirrors getPublished()'s two-step check: 404 "Rule set not found" if the set itself doesn't exist (checked first), then a separate 404 "No published rule pack version" if it exists but has nothing published. Updated the doc comment and the controller's OpenAPI response to 404 as well, and updated the downstream risk-referral-service ruleSet.client.ts (the only real caller) to handle 404 instead of 422. Added two new service-level tests for the two distinct 404 cases; existing tests updated to mock findSetById.

apps/risk-referral-service/src/risk-assessments/riskAssessment.controller.ts
Dharanishwaran-ND: Went with widening the route in c69233e — added SUPERVISOR/MANAGER/ADMIN to requireRoles on POST /risk-assessments, since the service logic is already written and tested for all four roles, and it matches how this codebase handles the equivalent case elsewhere (e.g. beneficiary.controller.ts, incentiveRate.controller.ts). Also fixed the doc/behavior mismatches you flagged in the same commit: dropped the impossible 502-from-beneficiary-service case (that push is deliberately swallowed/logged, never thrown) and the 200 the handler never returns (always 201).

Dharanishwaran-ND:
Thanks for the thorough review, @palani-ND — replied inline on each of the 4 main findings with the fix commit. Summary of what changed:

- **beneficiary-service** (`bd5db8e`): fixed the `phase` freeze — `update:` now sets `phase` alongside every other field, matching `create:`.
- **visit-form-service** (`9f513a3`): added `'ADMIN'` to `requireRoles` on both `PATCH /visits/:id` and `GET /visits/visit-summary`, matching `isPrivileged()` and both routes' own documented behavior.
- **rules-service** (`4baa061`): `evaluate()` now checks rule-set existence first (404 "Rule set not found") before checking for a published version (404 "No published..."), mirroring `getPublished()`. Also fixed the `Boolean(e.field)` trigger-flag coercion you flagged in the smaller items — now uses the same explicit `typeof` guard as `riskConditionId`/`grade`/`gradeRank`, so a literal `"false"` string from a decision graph is rejected instead of silently coercing to `true`. Added a regression test for that specific case.
- **risk-referral-service** (`c69233e`): widened `POST /risk-assessments` to `SAKHI`/`SUPERVISOR`/`MANAGER`/`ADMIN` to match the service's own already-implemented, already-tested authorization matrix, rather than trimming the service — went with widening since it's less churn and matches this codebase's convention elsewhere. Also updated `ruleSet.client.ts` for the 404 change above, and fixed the controller's OpenAPI doc mismatches (dropped the impossible 502-from-beneficiary-service case and the never-returned 200).

On the remaining smaller item — `ruleSet.evaluator.spec.ts` only exercising an identity graph, never a real risk-grading `rulesJson` — agreed this is a real gap, but didn't attempt a fix here since it'd need a hand-authored decision-table artifact from whoever owns the actual rule-pack content, which is out of scope for me to fabricate correctly. Flagging it as a known follow-up rather than leaving it silently unaddressed.

All 4 touched services' test suites pass (beneficiary-service 179, rules-service 29, risk-referral-service 35, visit-form-service 243), lint clean.

PR #128, opened 2026-08-09
feat(beneficiary-service): Populate riskConditionSummaries from enrollment health-history answers

Dharanishwaran-ND:
## Summary
- `GET /beneficiaries/:id` always returned an empty `riskConditionSummaries` for a mother's Q58 (diagnosed conditions) and Q60 (sickle cell status) enrollment answers — the app only showed them correctly when reading the same device's local cache. This never populated for a supervisor, a different device, or after a reinstall.
- `beneficiary-service`'s `create()` has no access to these answers (they live only in `visit-form-service`'s MOTHER_REGISTRATION submission), and `BeneficiaryRiskConditionSummary` is a graded clinical-risk rollup, not a flat conditions list — so instead of wiring `create()`, this pushes the resolved conditions right after MOTHER_REGISTRATION is submitted, mirroring the existing socio-demographics sync pattern.
- Adds `GET /risk-conditions?conditionCode=...` on risk-referral-service to resolve condition codes to `riskConditionId`s, seeds 17 ungraded `RiskCondition` rows (15 Q58 conditions + SCD/SCT), relaxes `grade`/`gradeRank` to nullable on beneficiary-service's upsert DTO for these self-reported entries, and adds `health-history.client.ts` in visit-form-service to extract + sync them.

## Test plan
- [x] `npx nx affected --target=lint,test --base=origin/develop` — 3 affected projects (beneficiary-service, risk-referral-service, visit-form-service), all lint clean, 494 tests passing, 0 failures
- [x] Verified the existing graded risk-assessment pipeline (risk-referral-service → beneficiary-service) is behaviorally unchanged for non-null grade/gradeRank — the new null-handling logic in `upsertRiskConditionSummary` algebraically reduces to the old formula for any non-null input
- [x] Confirmed no other consumer reads `latestGrade`/`everHighestGrade`/`everAtRiskFlag` assuming always-non-null (the one aggregation that reads them, `countByRiskGrade`, already handles null via `?? 'UNGRADED'`)
- [ ] Seed must be run in each target environment (`npm run seed` at repo root) before self-reported conditions can resolve via the new lookup endpoint

## Follow-ups / known limitations
- `everAtRiskFlag=true` is set unconditionally for these self-reported rows (a reported condition is treated as inherently a risk signal) — worth confirming downstream UI/reports that filter on this flag are fine mixing self-reported entries alongside rule-graded ones.
- No new Prisma migration needed — only new seed data and DTO relaxation.

### DIFF
--- a/apps/beneficiary-service/src/beneficiary/beneficiary.controller.ts
+++ b/apps/beneficiary-service/src/beneficiary/beneficiary.controller.ts
@@ -496,7 +496,7 @@ export function createBeneficiaryRouter(service: BeneficiaryService) {
           description: 'Risk condition summary upserted',
           schema: envelope(riskConditionSummarySchema),
         },
-        400: errorResponse(400, { message: 'gradeRank: Required' }),
+        400: errorResponse(400, { message: 'riskConditionId: Invalid uuid' }),
         401: errorResponse(401),
         403: errorResponse(403, { message: 'This beneficiary case is outside your own roster.' }),
         404: errorResponse(404, { message: 'Beneficiary case not found.' }),
--- a/apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
+++ b/apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
@@ -209,8 +209,7 @@ export class BeneficiaryRepository {
     // rule-engine evaluation — always at-risk (a reported condition is
     // inherently a risk signal) and never grade-ranked against the existing
     // row, since there is no rank to compare.
-    const isEverAtRisk =
-      data.grade === null || data.grade !== 'NORMAL' || (existing?.everAtRiskFlag ?? false);
+    const isEverAtRisk = data.grade !== 'NORMAL' || (existing?.everAtRiskFlag ?? false);
     const outranksEverHighest =
       data.gradeRank !== null &&
       (existing?.everHighestGradeRank == null || data.gradeRank > existing.everHighestGradeRank);
--- a/apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
+++ b/apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
@@ -204,9 +204,16 @@ export class BeneficiaryRepository {
       },
     });
 
-    const isEverAtRisk = data.grade !== 'NORMAL' || (existing?.everAtRiskFlag ?? false);
+    // A null grade means this is a self-reported, ungraded entry (e.g.
+    // enrollment-time diagnosed conditions/sickle cell status) rather than a
+    // rule-engine evaluation — always at-risk (a reported condition is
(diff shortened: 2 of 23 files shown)


palani-ND (commented):
Went through this carefully, including cross-checking the parts most prone to silent breakage:

- **`DIAGNOSED_CONDITION_CODE_TO_RISK_CONDITION_CODE`/`SICKLE_CELL_STATUS_CODE_TO_RISK_CONDITION_CODE`** — pulled the live Q58/Q60 options straight out of `mother-registration.json` and diffed them against these two maps by hand. Every one of the 15 Q58 value_codes and both Q60 positive value_codes matches exactly, and the excluded codes (`no_known_medical_condition`, `don_t_know`, and Q60's negative/pending codes) are correctly absent. This is exactly the kind of hardcoded cross-reference that silently goes stale (a typo'd key just never resolves, with no error), so glad this one's clean.
- **The null-handling algebra in `upsertRiskConditionSummary`** — traced both `isEverAtRisk` and `outranksEverHighest` through the create and update paths for all four orderings (self-reported first then graded, graded first then self-reported, self-reported only, graded only). A self-reported entry can never clobber an existing graded "ever highest" record (guarded by `data.gradeRank !== null`), and a later real graded assessment correctly overrides a self-reported entry's null baseline. No corruption path found. (Minor, non-functional nit: `data.grade === null || data.grade !== 'NORMAL'` — the first clause is redundant, since `null !== 'NORMAL'` is already true in JS, so `data.grade !== 'NORMAL'` alone covers it. Doesn't affect correctness, just a bit more than it needs to be.)
- **DTO defaulting** in `beneficiary.service.ts` correctly coerces every now-optional field (`grade`, `gradeRank`, `observedValueJson`, `visitId`, `submissionId`, `ruleVersionId`) from `undefined` to `null` before it reaches the repository, matching the repository's `| null` (not `| undefined`) types.
- **Seed script idempotency**: `findUnique` before `create`, keyed on the unique `conditionCode` column — safe to re-run, and there's no pre-existing seed data for this table that a self-reported condition code could collide with (this is genuinely the first seed script for `RiskCondition`).
- The `everAtRiskFlag=true`-unconditionally-for-self-reported-entries tradeoff is already called out in the PR's own Follow-ups section, so no need to re-flag it here beyond agreeing it's worth a follow-up check against any UI/report that filters on that flag.

Nothing else stood out. This is a well-scoped, well-tested fix.

Dharanishwaran-ND:
Fixed the redundant null check nit in `5e3592f` — `data.grade !== 'NORMAL'` already covers a null grade in JS (`null !== 'NORMAL'` is `true`), so the `data.grade === null ||` clause was dropped. Re-ran the full `upsertRiskConditionSummary` suite (184 tests) and lint after the change — both still pass, and CI's `verify` job is green on the latest commit. Thanks for the thorough review!

PR #129, opened 2026-08-10
fix(beneficiary-service): Enforce ownership check on GET /beneficiaries/:id

Dharanishwaran-ND:
## Summary

**Bug:** a beneficiary created under one SAKHI account was visible to another SAKHI. Root cause: `GET /beneficiaries/:id` had no ownership check at all — `BeneficiaryService.getById()` never received the caller's identity, unlike `list()`/`create()`/`applyLmpChange()`/`reactivateCase()`, which all correctly scope by caller. Any authenticated SAKHI/SUPERVISOR/MANAGER could fetch **any** beneficiary case by id (full PII included) just by knowing or guessing it.

**Fix:** `getById()` now applies the same ownership pattern already used elsewhere in this file — SAKHI may only fetch their own case, SUPERVISOR only a case whose Sakhi is in their roster, MANAGER unrestricted (ADMIN was never in this route's allowed roles to begin with, unrelated to this bug). Extracted a `projectCase()` private helper (decrypt + socio-demographics resolution, no ownership check) so the three internal post-mutation re-fetches (`upsertSocioDemographics`/`applyLmpChange`/`reactivateCase`) — which already verify ownership earlier in their own flow — don't need to thread a caller through `getById` redundantly.

## Root cause investigation

Followed systematic debugging before touching any code:
- Code-inspected `list()`, `create()`, the repository's Prisma query, and the JWT identity chain — all correct, ruled out.
- Reproduced live against a real environment with two fresh SAKHI accounts: `GET /beneficiaries` correctly isolated (0 results for the non-owner), `POST /beneficiaries` correctly ignored a spoofed `sakhiId` in the body — but **`GET /beneficiaries/:id` returned the other Sakhi's full record with 200 OK**. That pinpointed the actual broken endpoint.
- Confirmed via every existing `getById` unit test: all of them called it as `service.getById('x', AUTH_HEADER)` with no caller argument, proving the method genuinely never checked identity in any test either.

## Known related gap (not in this PR)

While investigating, found the identical missing-ownership-check pattern on `PATCH /beneficiaries/:id/socio-demographics` — same route-level role gate, same absence of any check. Deliberately **not included here** to keep this PR scoped to the reported bug; will be a separate follow-up PR.

## Test plan
- [x] Added 5 new unit tests: SAKHI-not-owner → 403, SAKHI-owner → 200, SUPERVISOR-out-of-roster → 403, SUPERVISOR-in-roster → 200, MANAGER-unrestricted → 200
- [x] Updated 6 pre-existing `getById` tests to pass a caller
- [x] `npx nx test beneficiary-service` — 189/189 passing
- [x] `npx nx lint beneficiary-service` — clean
- [x] Reproduced the bug live pre-fix (200 OK returning another Sakhi's beneficiary)
- [x] Re-verified the fix live on local after restarting services with current code — same request now returns `403 "This beneficiary case is outside your own roster."`; confirmed no regression for the owning SAKHI (200) and MANAGER (200, unrestricted)

### DIFF
--- a/apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
+++ b/apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
@@ -422,8 +422,47 @@ export class BeneficiaryService {
     });
   }
 
-  async getById(id: string, authorizationHeader: string) {
+  /**
+   * Gated by requireRoles('SAKHI', 'SUPERVISOR', 'MANAGER') at the route —
+   * same as upsertRiskConditionSummary, assertCallerCanTouchCase alone
+   * doesn't cover this route since it only scopes SUPERVISOR callers,
+   * early-returning for everyone else. A SAKHI reaching this method must be
+   * checked explicitly against the case's own sakhiId, or any authenticated
+   * Sakhi could fetch any beneficiary's full profile by id regardless of
+   * whose case it is (ADMIN is not in this route's role list, so MANAGER is
+   * the only unrestricted caller here).
+   */
+  async getById(id: string, caller: AuthenticatedUser, authorizationHeader: string) {
     const found = await this.repository.findById(id);
+    if (!found) throw notFound('Beneficiary case not found.');
+
+    if (caller.roles.includes('SAKHI')) {
+      if (found.sakhiId !== caller.id) {
+        throw forbidden('This beneficiary case is outside your own roster.');
+      }
+    } else {
+      await assertCallerCanTouchCase(found.sakhiId, caller, authorizationHeader);
+    }
+
+    return this.projectCase(id, authorizationHeader, found);
+  }
+
+  /**
+   * Decrypts/enriches a case row for the response envelope, without any
+   * ownership check — only for use by callers that have already verified
+   * (or, per their own route's role gate, can't reach a case outside their
+   * own scope) that the requester may see this record: getById (after its
+   * own check above) and the post-mutation re-fetches in
+   * upsertSocioDemographics/applyLmpChange/reactivateCase (each of which
+   * already ran assertCallerCanTouchCase/the SAKHI-ownership check before
+   * reaching this point). Never call this directly from a route handler.
+   */
+  private async projectCase(
+    id: string,
+    authorizationHeader: string,
+    prefetched?: Awaited<ReturnType<BeneficiaryRepository['findById']>>,
+  ) {
+    const found = prefetched ?? (await this.repository.findById(id));
     if (!found) throw notFound('Beneficiary case not found.');
     const projected = withDecryptedName(found);
     return withResolvedSocioDemographics(projected, authorizationHeader);
@@ -493,7 +532,7 @@ export class BeneficiaryService {
     if (dto.childrenUnder5Count !== undefined) data.childrenUnder5Count = dto.childrenUnder5Count;
 
     await this.repository.upsertSocioDemographics(beneficiaryId, data);
-    return this.getById(beneficiaryId, authorizationHeader);
+    return this.projectCase(beneficiaryId, authorizationHeader);
   }
 
   /**
@@ -527,7 +566,7 @@ export class BeneficiaryService {
     const eddDate = addDays(lmpDate, GESTATION_DAYS);
     const updated = await this.repository.updateMotherLmp(beneficiaryId, lmpDate, eddDate);
     if (!updated) throw notFound('Beneficiary case not found.');
-    return this.getById(beneficiaryId, authorizationHeader);
+    return this.projectCase(beneficiaryId, authorizationHeader);
   }
 
   /**
@@ -562,7 +601,7 @@ export class BeneficiaryService {
       throw conflict(`Cannot reactivate a case with status ${existing.currentStatus}.`);
     }
 
-    return this.getById(beneficiaryId, authorizationHeader);
+    return this.projectCase(beneficiaryId, authorizationHeader);
   }
 
   /**
--- a/apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
+++ b/apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
@@ -452,10 +452,13 @@ export class BeneficiaryService {
    * ownership check — only for use by callers that have already verified
    * (or, per their own route's role gate, can't reach a case outside their
    * own scope) that the requester may see this record: getById (after its
-   * own check above) and the post-mutation re-fetches in
-   * upsertSocioDemographics/applyLmpChange/reactivateCase (each of which
-   * already ran assertCallerCanTouchCase/the SAKHI-ownership check before
-   * reaching this point). Never call this directly from a route handler.
+   * own check above) and the post-mutation re-fetches in applyLmpChange/
+   * reactivateCase (each of which already ran assertCallerCanTouchCase
+   * before reaching this point). upsertSocioDemographics also calls this,
+   * but takes no caller and runs no ownership check of its own — a known,
+   * separately-tracked gap (see PATCH /beneficiaries/:id/socio-demographics
+   * in the PR description), not something this helper covers for it. Never
+   * call this directly from a route handler.
    */
   private async projectCase(
     id: string,
--- a/apps/rules-service/src/rules/scheduling/anc.rulesJson.ts
+++ b/apps/rules-service/src/rules/scheduling/anc.rulesJson.ts
@@ -10,6 +10,16 @@
  * ANC visits (FR-S-3.7/BR-04/G.3) — that is a DB-state mutation the calling
  * service performs; this pack only reports `deliveryFormFiledByEddPlus7`
  * so the caller knows whether the Post-EDD visit applies.
+ *
+ * This pack takes no "today" input, so `deliveryFormFiledByEddPlus7: false`
+ * (and the accompanying `postEddVisit`) is returned identically whether the
+ * delivery form is unfiled because EDD+7 has genuinely passed, or because
+ * the woman is still mid-pregnancy, months from her EDD, and simply hasn't
+ * delivered yet. Both cases have `deliveryFormFiledDate: null` and are
+ * indistinguishable from this function's inputs alone. The caller MUST
+ * independently check that the real calendar date has passed EDD+7 before
+ * treating `postEddVisit` as due — this pack's output is a candidate visit,
+ * not a decision that it applies yet.
  */
 export const ancRulesJson = {
   contentType: 'application/vnd.gorules.decision',
--- a/apps/rules-service/src/rules/scheduling/anc.rulesJson.ts
+++ b/apps/rules-service/src/rules/scheduling/anc.rulesJson.ts
@@ -1,0 +1,82 @@
+/**
+ * ANC visit-schedule decision graph (SRS §3A.2.3 "ANC Visit Schedule",
+ * Appendix A.1, Appendix B, BR-01/BR-08). A single gorules function node
+ * computing the full formula-driven, uncapped ANC schedule from
+ * registrationDate/edd — stateless and pure, so re-running it with a new
+ * `edd` after a Supervisor-approved LMP/EDD change (BR-01) IS the
+ * regeneration; no separate "regenerate" rule exists.
+ *
+ * Does not decide whether the delivery form has actually lapsed the open
+ * ANC visits (FR-S-3.7/BR-04/G.3) — that is a DB-state mutation the calling
+ * service performs; this pack only reports `deliveryFormFiledByEddPlus7`
+ * so the caller knows whether the Post-EDD visit applies.
+ */
+export const ancRulesJson = {
+  contentType: 'application/vnd.gorules.decision',
+  nodes: [
+    { id: 'input1', type: 'inputNode', name: 'request', position: { x: 0, y: 0 } },
+    {
+      id: 'fn1',
+      type: 'functionNode',
+      name: 'computeAncSchedule',
+      position: { x: 200, y: 0 },
+      content: `
+const handler = (input, { dayjs }) => {
+  const { registrationDate, edd, deliveryFormFiledDate } = input;
+  const reg = dayjs(registrationDate);
+  const eddDate = dayjs(edd);
+
+  // FR-S-3.1 / Appendix A.1: ((EDD - Registration date) / 30) + 1, uncapped.
+  const totalVisits = Math.round(eddDate.diff(reg, 'day') / 30) + 1;
+
+  const visits = [];
+  // FR-S-3.2: ANC1 on Day 0 (registration date). Window Day 0 to +5.
+  visits.push({
+    visitName: 'ANC1',
+    scheduledDate: reg.format('YYYY-MM-DD'),
+    windowOpen: reg.format('YYYY-MM-DD'),
+    windowClose: reg.add(5, 'day').format('YYYY-MM-DD'),
+  });
+
+  // FR-S-3.3: ANC2..ANCn every 30 days from the previous scheduled date, +/-5 day window.
+  for (let i = 2; i <= totalVisits; i++) {
+    const scheduled = reg.add((i - 1) * 30, 'day');
+    visits.push({
+      visitName: 'ANC' + i,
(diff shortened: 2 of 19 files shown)


Dharanishwaran-ND:
Pushed 2 more commits to this branch — note these are unrelated to this PR's original scope (the `GET /beneficiaries/:id` ownership fix), added here per explicit instruction rather than opened as separate PRs:

- `9aaab5c` — real GoRules scheduling packs (ANC/PP/NN/INC/CCV/HR/DELIVERY) via a new `POST /rules/:setId/evaluate-schedule`. This was in-progress work already staged in the working tree from elsewhere; committing it here rather than discarding it. Verified: 66/66 rules-service tests passing, lint clean.
- `b7f2aea` — suppresses a Node `DEP0060` deprecation warning (`util._extend`) that shows up when running `npm run serve:all` locally. Traced to `@nx/webpack` → `webpack-dev-server` → `spdy`, three layers below our own code — no usages of `util._extend` anywhere in `apps/`/`libs/`. Scoped only to the `serve:all` script; `build`/`test`/`lint` are untouched.

Happy to split these into separate PRs if preferred — flagging here for visibility since they landed on this branch.

apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
```diff
@@ -422,9 +422,48 @@ export class BeneficiaryService {
...
+  }
+
+  /**
+   * Decrypts/enriches a case row for the response envelope, without any
+   * ownership check — only for use by callers that have already verified
+   * (or, per their own route's role gate, can't reach a case outside their
+   * own scope) that the requester may see this record: getById (after its
+   * own check above) and the post-mutation re-fetches in
+   * upsertSocioDemographics/applyLmpChange/reactivateCase (each of which
+   * already ran assertCallerCanTouchCase/the SAKHI-ownership check before
```
palani-ND: This claims `upsertSocioDemographics` "already ran assertCallerCanTouchCase/the SAKHI-ownership check" — but that method takes no `caller` argument (see its signature below), so it can't have. This contradicts the PR description's own disclosure that this exact gap exists and is being deferred to a follow-up — worth narrowing this comment to just `applyLmpChange`/`reactivateCase`, which do genuinely check first.

apps/rules-service/src/rules/scheduling/ccv.rulesJson.ts
```diff
@@ -0,0 +1,134 @@
...
+    cadence19to24MonthsEveryNMonths = 1;
+  } else if (mostRecentIncVisitHrType === 'OTHER') {
+    riskState = 'CURRENTLY_HR_OTHER';
+    cadence13to18MonthsEveryNMonths = 1;
+    cadence19to24MonthsEveryNMonths = 1;
+  } else if (!last3IncVisitsNormal) {
+    // HR was detected historically, and the last 3 visits are still not
+    // all-normal, but the most recent visit itself is not HR - treat as
+    // Recently Recovered per Appendix A.5's "Last 3 INC visits were at risk".
+    riskState = 'RECENTLY_RECOVERED';
```
palani-ND, line 77: This branch's comment ("last 3 INC visits were at risk... per Appendix A.5") describes the opposite condition from this file's own top-of-file docstring for `RECENTLY_RECOVERED` ("HR detected earlier, but last 3 INC visits normal"). Worth resolving which description actually matches Appendix A.5 — as written, the file contradicts itself on what makes a case "recently recovered" vs "stable low risk."

apps/rules-service/src/rules/scheduling/anc.rulesJson.ts
```diff
@@ -0,0 +1,82 @@
...
+      visitName: 'ANC' + i,
+      scheduledDate: scheduled.format('YYYY-MM-DD'),
+      windowOpen: scheduled.subtract(5, 'day').format('YYYY-MM-DD'),
+      windowClose: scheduled.add(5, 'day').format('YYYY-MM-DD'),
+    });
+  }
+
+  // SR-ANC-01 / BR-08: if delivery form not filed by EDD+7, generate a dynamically
+  // named ANC(n+1) visit on EDD+8, window EDD+8 to EDD+13.
+  const deliveryFiledByEddPlus7 =
```
palani-ND, line 64: This is `false` (triggering a computed `postEddVisit`) any time `deliveryFormFiledDate` is null — including for a woman who's still mid-pregnancy, months from her EDD, not just one who's genuinely passed EDD+7 without filing. The function has no "today" input, so it can't distinguish "not filed, and it's not even close to EDD" from "not filed, and EDD+7 has actually passed" — both produce the same `deliveryFiledByEddPlus7: false` and the same populated `postEddVisit`. Whatever calls this needs to independently gate on the real calendar date before scheduling `postEddVisit`, since this flag alone doesn't tell them that.

palani-ND (commented):
The core fix (ownership check on `getById`, `projectCase()` extraction, the 5 new + 6 updated tests) is solid — matches the existing SAKHI/SUPERVISOR/MANAGER pattern used elsewhere in this file, and the live-reproduction-before-fix approach in the PR description is exactly the right way to validate a bug like this.

A few things worth a look, split by the two halves of this diff:

### beneficiary-service (the stated fix)

**The new `projectCase()` doc comment misstates one of the three call sites it lists.** It says `upsertSocioDemographics`/`applyLmpChange`/`reactivateCase` all "already ran `assertCallerCanTouchCase`/the SAKHI-ownership check before reaching this point" — true for the latter two, but `upsertSocioDemographics` doesn't take a `caller` parameter at all, so it structurally can't have run any such check. This lines up with what the PR description itself already discloses ("found the identical missing-ownership-check pattern on `PATCH /beneficiaries/:id/socio-demographics`... deliberately not included here") — so the underlying gap is known and intentionally deferred, but this new comment asserts the opposite of that disclosure for the one function it doesn't apply to. Worth fixing the comment now so it doesn't contradict the PR's own stated context, especially since whoever picks up the deferred follow-up will read this comment as evidence it's already handled.

### rules-service (the bundled scheduling-packs commit)

Appreciate the transparency in the commit message about this being unrelated in-progress work kept separate rather than folded in or discarded — agreed it'd be cleaner to land as its own PR before merge, if that's still feasible, given it's a completely different feature (visit-schedule generation) from the stated bug fix. Since it's here, a few things in the actual scheduling logic worth checking against the SRS before this ships, given the stakes (maternal/child visit cadences):

1. **CCV pack's `RECENTLY_RECOVERED` vs `STABLE_LOW_RISK` split may not match its own file-level docstring.** The top-of-file comment describes both states as requiring "last 3 INC visits normal" (differing only by immunization/growth status, which isn't even one of this function's inputs) — but the actual branching uses `last3IncVisitsNormal` as the *sole* differentiator: `!last3IncVisitsNormal` → `RECENTLY_RECOVERED`, else → `STABLE_LOW_RISK`. The inline comment at the `RECENTLY_RECOVERED` branch ("last 3 INC visits were at risk... per Appendix A.5's 'Last 3 INC visits were at risk'") directly contradicts the file's own top-of-file description of that same state ("last 3 INC visits normal"). Worth resolving which of these two in-file descriptions actually matches Appendix A.5 before trusting either — right now the file disagrees with itself.

2. **ANC pack can flag a Post-EDD visit as due for a woman who's nowhere near her EDD yet.** `deliveryFiledByEddPlus7` (and the `postEddVisit` this pack always computes when that flag is false) is derived purely from whether `deliveryFormFiledDate` is set — there's no "has EDD+7 actually passed" check, and the function doesn't even take today's date as an input, so it has no way to express that distinction. For any currently-pregnant woman mid-ANC (`deliveryFormFiledDate: null`, months before EDD), `deliveryFiledByEddPlus7` comes back `false` and a dated `postEddVisit` is returned as if BR-08's trigger condition were already met. The doc comment says the caller uses this flag to decide whether the Post-EDD visit applies — if any future caller trusts it at face value without an independent "is it actually past EDD+7 yet" check, every ANC evaluation before delivery would surface a spurious Post-EDD visit.

A couple of smaller, lower-confidence items also worth a look while this is fresh: `evaluateSchedule()`/`evaluate()` don't check the resolved rule set's category before feeding its `rulesJson` into the engine, so a `:setId` for a RISK/other-category pack passed to `evaluate-schedule` (or vice versa) would likely hit an engine-level error instead of a clean 404; and `scheduleEvaluator.ts`'s `engine.dispose()` only runs after a successful `decision.evaluate()`, so a throwing evaluation (malformed graph, wrong-category mismatch per the above) skips disposal.

None of this blocks the actual stated fix — just flagging before the scheduling packs go live, since (per a repo-wide grep) nothing currently calls the new `evaluate-schedule` endpoint yet, so there's still room to get these right before a real caller depends on them.

apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
Dharanishwaran-ND: Fixed — the comment now cites only `applyLmpChange`/`reactivateCase` and explicitly calls out that `upsertSocioDemographics` takes no `caller` and runs no ownership check, per the PR description's own disclosure. See beneficiary.service.ts's updated `projectCase()` doc comment.

apps/rules-service/src/rules/scheduling/ccv.rulesJson.ts
Dharanishwaran-ND: Traced the actual branching logic, not just the wording — this was a real inversion, not only a stale comment. Per SRS Appendix A.5 (lines 1557-1559), RECENTLY_RECOVERED = 'HR detected earlier but last 3 INC visits normal' (matches this file's own top-of-file docstring), which means `last3IncVisitsNormal === true` maps to RECENTLY_RECOVERED, not STABLE_LOW_RISK — the code had this backwards. Fixed the branching (now `else if (last3IncVisitsNormal)` → RECENTLY_RECOVERED) and updated the inline comments to stop citing the summary table at line 355, which contradicts its own line 1557-1559 table.

Also surfaced a second, previously undocumented gap while fixing this: the branch reached when HR was detected earlier, the last 3 visits are *not* all normal, and the most recent visit itself isn't HR has no state defined for it anywhere in Appendix A.5 — both named states require last3IncVisitsNormal=true. Documented this as a known gap (falls through to STABLE_LOW_RISK's cadence provisionally) and flagged it for SRS/product clarification before this pack goes live, per your note that nothing calls evaluate-schedule yet.

Updated the existing unit tests in scheduleEvaluator.spec.ts to match the corrected mapping — they had encoded the same inverted expectation. All rules-service tests pass (71/71).

apps/rules-service/src/rules/scheduling/anc.rulesJson.ts
Dharanishwaran-ND: Confirmed — this pack has no 'today' input so it genuinely can't distinguish those two cases. Strengthened the top-of-file doc comment to state this explicitly: `deliveryFormFiledByEddPlus7: false` and the accompanying `postEddVisit` are returned identically whether EDD+7 has actually passed or the woman is still mid-pregnancy, and the caller MUST independently gate on the real calendar date before treating `postEddVisit` as due — it's a candidate, not a decision. Left the actual computation unchanged since the doc already disclosed the caller owns the 'has it lapsed' decision; this just makes the specific failure mode explicit so a future caller doesn't trust the flag at face value.

Dharanishwaran-ND:
Addressed all review feedback in 1bc848a:

**Inline comments** (replied individually in their threads):
- `projectCase()` doc comment fixed — no longer misattributes an ownership check to `upsertSocioDemographics`.
- CCV `RECENTLY_RECOVERED`/`STABLE_LOW_RISK` — this turned out to be a real logic inversion, not just a stale comment. Traced the branching against SRS Appendix A.5 (lines 1557-1559) and this file's own docstring: `last3IncVisitsNormal === true` should map to `RECENTLY_RECOVERED`, and the code had it backwards. Fixed the branch condition, updated the existing unit tests (which had encoded the same inverted expectation), and documented a separate SRS gap this surfaced: the combination "HR earlier, last 3 not normal, most recent visit not itself HR" has no defined state in Appendix A.5 at all. Flagged for SRS/product clarification before this pack goes live.
- ANC `deliveryFiledByEddPlus7` — confirmed the function has no "today" input and genuinely can't distinguish "not filed, EDD+7 not yet reached" from "not filed, EDD+7 passed." Strengthened the doc comment so a future caller can't miss that `postEddVisit` is a candidate, not an authoritative "due now" signal.

**Review-body items** (lower-confidence, not inline, but verified as real gaps):
- `evaluate()`/`evaluateSchedule()` now check the resolved rule set's `ruleCategory` (RISK/SCHEDULE respectively) and 400 on mismatch instead of feeding a wrong-category pack into the engine. Added test coverage for both.
- `engine.dispose()` in both `ruleSet.evaluator.ts` and `scheduleEvaluator.ts` now runs in a `finally` block so a throwing `decision.evaluate()` no longer skips disposal.

All rules-service (71/71) and beneficiary-service (189/189) tests pass; lint clean on both.

apps/rules-service/src/rules/scheduling/ccv.rulesJson.ts
```diff
@@ -0,0 +1,147 @@
...
+  } else {
+    // GAP: HR was detected earlier, the most recent visit itself is not HR,
+    // but the last 3 INC visits are still not all normal. Appendix A.5's
+    // five-state model (SRS lines 350-356 / 1557-1559) does not define a
+    // distinct condition for this combination - both named non-HR states
+    // (Recently Recovered, Stable Low Risk) require last3IncVisitsNormal to
+    // be true. Falling through to Stable Low Risk's cadence here is
+    // provisional pending SRS/product clarification (flagged in PR #129
+    // review) - do not treat this as validated clinical behaviour.
+    riskState = 'STABLE_LOW_RISK';
```
palani-ND, line 89: This is the provisional fallback for the SRS-undefined combination (HR earlier, most recent visit not HR, last 3 visits still not normal). Given the last-3-not-normal signal is the same kind of active-risk indicator that drives the monthly cadence elsewhere in this function, defaulting an undefined case to the *least* frequent cadence (every 2 months) seems like erring in the riskier direction rather than the safer one. Would `RECENTLY_RECOVERED`'s monthly 13-18m cadence be a more conservative provisional choice until Appendix A.5 is clarified, given the data doesn't actually look low-risk here?

palani-ND (commented):
Re-reviewed after `1bc848a`. All four points from the previous review are correctly addressed:

- `projectCase()`'s doc comment now correctly singles out `upsertSocioDemographics` as not covered, matching the PR description's own disclosure instead of contradicting it.
- The CCV `RECENTLY_RECOVERED`/`STABLE_LOW_RISK` branch is flipped to match Appendix A.5 lines 1557-1559 (and now matches this file's own top-of-file docstring), with the summary-table-vs-detailed-table SRS contradiction called out explicitly rather than silently picked one way.
- The ANC pack's doc comment now spells out exactly the limitation flagged — that `deliveryFormFiledByEddPlus7` can't tell "not filed, EDD+7 not reached" from "not filed, EDD+7 passed" — and states plainly that the caller must gate on the real calendar date. Good, honest documentation of a real constraint rather than a change in behavior.
- Both `evaluate()`/`evaluateSchedule()` now reject a category-mismatched rule set with a clean 400 before ever touching the engine, and `engine.dispose()` is now in a `finally` in both evaluators. Both are well covered by new tests, including the not-found-vs-wrong-category distinction.

One follow-up worth a look, on the CCV fix specifically: the newly-labeled `GAP` branch (HR detected earlier, most recent visit not HR, but last 3 INC visits *not* all normal — the combination Appendix A.5 doesn't define) falls through to `STABLE_LOW_RISK`'s cadence, which is the *least* frequent monitoring option (every 2 months, both sub-periods). Given this combination's actual signal — last 3 visits still showing risk — falling back to the lowest-monitoring-frequency state seems like erring in the wrong direction for an undefined case; `RECENTLY_RECOVERED`'s cadence (monthly for 13-18m) seems like the safer provisional default until Appendix A.5 is clarified, precisely because the underlying data here doesn't actually look "stable" or "low risk." The comment already flags this as provisional and not validated clinical behavior, so this is really a question of which direction to provisionally err in, not a correctness bug in the code as written — but worth deciding deliberately rather than inheriting whichever state happened to be the `else` branch.

PR #146, opened 2026-08-11
chore(repo): Pin Node.js to v24.18.0 across CI, Dockerfiles, and local dev config

Dharanishwaran-ND:
## Summary

Pins Node.js to **v24.18.0** everywhere, closing a known gap: the dev box already runs Node 24 while CI pinned 20.11.0, which is exactly why `cd-dev.yml` had to skip `npm ci` on deploy (stale warning comment removed here now that the mismatch it described no longer exists).

## Changes

- `.nvmrc` and all 17 `apps/*/Dockerfile` (34 `FROM` lines, builder + runner) — **hard-pinned to `24.18.0`** exactly, since these drive actual CI/deploy reproducibility.
- `package.json` `engines.node` — **`>=24.18.0 <25.0.0`** (a range, not an exact pin), so a contributor on a later 24.x patch isn't blocked by `npm install` on their local machine. `.nvmrc`/Dockerfiles remain the source of truth for the exact version that actually builds/deploys.
- `@types/node` bumped `^20.14.0` → `^24.13.0` to match the Node 24 line.
- `.github/workflows/ci.yml`: `setup-node` now uses `24.18.0`.
- `README.md`: fixed stale "Node 20.11.0" references in Prerequisites/Getting-started that were missed in the initial pass. Also corrected — Node 24 ("Krypton") is **Active LTS**, not a non-LTS current release as I'd initially and incorrectly assumed in review discussion.
- Removed the now-stale Node 20/24 mismatch warning comment in `cd-dev.yml`.

## Verification

- Confirmed `node:24.18.0-alpine` exists on Docker Hub (`docker manifest inspect` succeeds) — this exact-patch tag is real, not just assumed.
- `npm install` succeeds under the new `engines` range; `package-lock.json` diff stays small and contained.
- `npx nx run-many -t lint` — all 19 projects clean (re-verified after the `engines` range change too).
- `npx nx run-many -t build` — all 19 projects build successfully.
- `npx nx run-many -t test` — every project with test files passes (rules-service 71/71, beneficiary-service 189/189, api-gateway 28/28, audit-service 6/6, media-service 12/12, etc.). `wrapper-api-service`/`cms-content-service` report "No tests found" — confirmed pre-existing on `develop`, unrelated to this change.

## Notes

- `npm audit` reports pre-existing vulnerabilities (2 critical, 30 high) in the lockfile, unrelated to this change but worth a separate look/ticket.


apps/visit-form-service/src/visit-schedules/visitSchedule.service.ts
```diff
@@ -219,6 +205,125 @@ export class VisitScheduleService {
...
+  async regenerateAncSchedule(
+    beneficiaryId: string,
+    registrationDate: string,
+    edd: string,
+    caller: CallerIdentity,
+    authorizationHeader: string,
+  ): Promise<{ supersededCount: number; created: number; schedules: CreatedScheduleResult[] }> {
+    await this.assertBeneficiaryAccess(beneficiaryId, caller, authorizationHeader);
+
+    const superseded = await this.repository.supersedeAnc(beneficiaryId, caller.id);
```
palani-ND: This commits immediately (plain `updateMany`), but the `evaluateAncScheduleFull` call two lines below isn't wrapped in try/catch and can throw `badGateway` on any rules-service hiccup — at which point this supersede has already committed and nothing creates the replacement rows. Since the caller (`beneficiary-service`'s `scheduleRegeneration.client.ts`) treats this whole endpoint as best-effort and just logs failures, a transient rules-service error here leaves the beneficiary with zero open ANC visits and no automatic retry. Wrapping supersede+evaluate+insert in one flow where the evaluate happens before any write commits (or in a single `$transaction`) would close this window.

apps/rules-service/src/rules/ruleSet.evaluator.ts
```diff
@@ -4,6 +4,31 @@ import { badRequest } from '@armman/service-commons';
...
+ */
+export async function runDecisionGraph(
+  rulesJson: unknown,
+  input: Record<string, unknown>,
+): Promise<unknown> {
+  const engine = new ZenEngine();
+  const content = new ZenDecisionContent(rulesJson as object);
+  const decision = engine.createDecision(content);
+  const response = await decision.evaluate(input);
+  engine.dispose();
```
palani-ND: `engine.dispose()` only runs if `decision.evaluate()` resolves — a throwing evaluation (malformed rulesJson, category mismatch, etc.) skips disposal and leaks the native zen-engine instance. This was flagged on the single `evaluateRulePack` call site in PR #127; now that `evaluateScheduleRulePack` and `evaluateEscalationRulePack` also route through this shared helper, the same gap applies to all three. Worth a `try { ... } finally { engine.dispose(); }` here.

palani-ND (commented):
This is a big one — Node pin, the visit-form-service schedule sync/lapse/regeneration APIs, and a full GoRules-graph rewrite of the scheduling/escalation rule packs in rules-service. Traced the LMP-change-triggers-regeneration chain end to end (beneficiary-service → gateway → visit-form-service → rules-service) and personally verified the code below directly rather than relying on the earlier read-through alone. Also spot-checked whether the two issues flagged on PR #129's rulesJson-based CCV/ANC packs recur in the new GoRules `graphs/*.graph.ts` versions.

### Critical: an LMP/EDD correction can leave a beneficiary with zero open ANC visits, silently, with no retry

`VisitScheduleService.regenerateAncSchedule` (`visitSchedule.service.ts:255-289`) does, in order:
1. `this.repository.supersedeAnc(beneficiaryId, caller.id)` — a plain `updateMany`, commits immediately, marking every OPEN/GENERATED ANC-family row SUPERSEDED.
2. `evaluateAncScheduleFull(...)` — an HTTP call to rules-service, **not wrapped in try/catch**.
3. `this.repository.createAllOrNothing(...)` — inserts the new rows.

`evaluateAncScheduleFull` (`ruleVersion.client.ts:64-84`) throws `badGateway(...)` on any network failure or non-2xx from rules-service. That throw propagates straight through this method with nothing to compensate for step 1 having already committed — step 3 never runs. On the caller side, `beneficiary-service`'s `scheduleRegeneration.client.ts` treats this whole call as best-effort by design (its own doc comment says so explicitly: "the LMP/EDD write itself must succeed regardless of whether regeneration does... no retry/reconciliation mechanism exists anywhere in this codebase yet") and just logs the failure. So a single transient rules-service blip during an LMP-change approval leaves a real beneficiary with **no open/generated ANC visit schedule at all** — old rows superseded, new rows never created — and the only trace is a log line. Nothing polls for or retries this state.

Given `supersedeAnc` and `createAllOrNothing` are both single-purpose repository methods already, wrapping steps 1–3 in one Prisma `$transaction` (evaluate first, then supersede+insert together) would remove the window entirely — evaluate-then-write means a rules-service failure never touches the DB. `visitSchedule.service.spec.ts` only covers the happy path and a pre-write auth rejection; there's no test for "evaluate throws after supersede already committed," so this wouldn't be caught by the suite either.

### Medium: `engine.dispose()` still isn't in a `finally`, and now three evaluators share the leak instead of one

`ruleSet.evaluator.ts`'s new `runDecisionGraph()` helper (lines 20-30) is `evaluateRulePack`'s pre-existing engine-lifecycle pattern extracted into a shared function — `engine.dispose()` still runs only after `decision.evaluate()` resolves, with no try/finally. This was already flagged as a minor item on PR #127 for the single `evaluateRulePack` call site; this PR routes `evaluateScheduleRulePack` and `evaluateEscalationRulePack` through the same helper, so a throwing evaluation (malformed rulesJson, a category-mismatched rule set slipping through, etc.) now leaks a native zen-engine instance from three call sites instead of one. Worth fixing once here since it's now shared: `try { ... } finally { engine.dispose(); }`.

### CCV/ANC graph re-check (against the PR #129 findings)

The scheduling packs have been substantially redesigned since PR #129's `*.rulesJson.ts` versions (now real GoRules decision tables driven by `scheduleOrchestrator.ts`, one call per candidate visit rather than one big function per pack), so the two specific issues flagged there don't carry over as-is:
- The ANC graph no longer takes a `deliveryFormFiledDate`-derived flag at all — post-EDD scheduling is now offset/window arithmetic off `edd` directly (`anc.graph.ts`), so the old "can't distinguish not-yet-due from overdue" ambiguity doesn't apply to this shape of the logic (I didn't trace where/whether `scheduleOrchestrator.ts` gates on today's date before exposing a POST_EDD row — worth a quick look there specifically if that matters to whoever picks this up next, but the mechanism that produced the original bug is gone).
- The CCV risk-state table (`ccv.graph.ts:69-111`) now maps `last3AllAtRisk: true` to `RECENTLY_RECOVERED`'s faster cadence (`o2: '1'`) rather than falling through to `STABLE_LOW_RISK`'s slowest cadence — which is actually the direction I'd suggested erring in during the PR #129 follow-up for this exact ambiguous case. The `RECENTLY_RECOVERED` label for an "at risk" input is still a bit of a semantic mismatch on first read (data says at-risk, label says recovered), but functionally this looks like the safer choice was made.

I did not do a line-by-line audit of the full graph set (7 `*.graph.ts` files plus `scheduleOrchestrator.ts` and `escalation.graph.ts` — this is a lot of new decision-table surface) — this was a targeted check against the two specific prior findings, not a clean bill of health for the whole rewrite.

Everything else held up on inspection: `visitInstance.controller.ts`'s `GET /visits` role list now correctly includes ADMIN (fixing the earlier omission pattern); `resolveSakhiScope`'s SAKHI/SUPERVISOR/MANAGER/ADMIN scoping is applied consistently and the keyset pagination is correctly tie-broken; the `ruleVersion.controller.ts` role widening on `GET /admin/rules/:setId` (ADMIN-only → SAKHI/SUPERVISOR/MANAGER/ADMIN) reads as a deliberate, documented change for offline rulesJson caching rather than an accidental broadening.

Dharanishwaran-ND:
Thanks for the deep trace on this, @palani-ND.

Heads up: this review was submitted against commit `02309be`, which is now stale — since your review, this branch was completely rebuilt to scope it strictly to the Node 24.18.0 pin (per its title/description). That rebuild dropped 4 commits that had drifted from `develop` with an older, since-superseded implementation of the visit-form-service schedule sync/lapse/regeneration APIs and the rules-service GoRules graph rewrite.

Current `chore/node-24-18-0` diff vs. `develop` is now just: `.nvmrc`, all 17 `apps/*/Dockerfile`, `package.json`/`package-lock.json`, `.github/workflows/ci.yml`, and `README.md`. Neither `apps/visit-form-service/src/visit-schedules/visitSchedule.service.ts` (regenerateAncSchedule) nor `apps/rules-service/src/rules/ruleSet.evaluator.ts`'s `runDecisionGraph` helper is part of this PR anymore.

Checked both findings against current `develop` directly, since that's where the reviewed code actually lives now:

- **Critical (supersede-before-evaluate race)**: `develop`'s `visitSchedule.service.ts` has no `regenerateAncSchedule`/`supersedeAnc`/`evaluateAncScheduleFull` at all — that ANC-regeneration feature was never merged to `develop`. Not applicable as filed; if/when that feature is re-proposed in a future PR, this is a real design issue worth carrying forward into that review.
- **Medium (`engine.dispose()` leak)**: `develop`'s `ruleSet.evaluator.ts` has no `runDecisionGraph` helper — just `evaluateRulePack`, and it already wraps `decision.evaluate()` in `try { ... } finally { engine.dispose(); }` (the PR #127 fix is already in place there). Not applicable to current `develop`.

No code changes needed on this PR as a result — both findings are on code this branch no longer contains. Let me know if you'd like a fresh review pass against the current diff.

PR #151, opened 2026-08-11
feat: GoRules scheduling/escalation, schedule sync APIs, and LMP-change regeneration

### LINKED ISSUE #147: visit_schedules / visit_instances read, sync, and lapse APIs
## Objective
Close the read/sync/lapse gaps in visit_schedules and visit_instances per SRS v3.0 §3A.2.3/§3A.2.8, FR-S-3.7, FR-S-10.1/10.2.

## Sub-tasks
- [x] **1.1** — Add `LAPSED` to `VisitScheduleStatus` enum (Prisma migration)
- [x] **1.2** — `GET /visit-schedules` — new list/sync-pull endpoint (query: `beneficiaryId`, `status?`, `updatedAfter?`, `cursor?`, `limit`), scoped via beneficiary→sakhi ownership check
- [x] **1.3** — Fix `GET /visits` in place (breaking change, confirmed) — same scoping/filter/pagination shape as 1.2, replacing the current unscoped `take: 50`
- [x] **1.4** — Bulk-lapse logic: `lapseOpenSchedules(beneficiaryId, reason)`, built as a standalone service/repository method (not yet wired to a form-submission trigger — DELIVERY/CLOSURE form codes don't exist in this repo yet)
- [x] **1.5** — Schedule regeneration on approved LMP/EDD change — wired end-to-end: `beneficiary-service`'s `applyLmpChange` now triggers `visit-form-service`'s new `POST /visit-schedules/regenerate-anc`, which supersedes old ANC rows and calls `rules-service`'s new `POST /rules/:setId/evaluate-schedule/anc-full` (wrapping the existing `generateAncSchedule` orchestrator) to generate the replacement schedule. Best-effort, matching the codebase's existing cross-service side-effect convention.

## Status: Complete

All 5 sub-tasks implemented, tested, and verified (`nx affected --target=lint,test` clean across `rules-service`, `visit-form-service`, `beneficiary-service`).

### Follow-ups spun out of this work (not blocking, tracked separately if picked up later)
- 1.4's trigger wiring to real DELIVERY/CLOSURE form submissions is deferred until those form codes exist as seeded `FormDefinition`s.
- 1.5 also gave `rules-service`'s `evaluate-schedule` family its first real production caller — closing part of Phase 4 (#150)'s "disconnected from live flow" gap as a side effect.

## Context
See prior completion-status audit in this session's conversation history for full evidence (file:line citations) backing each gap.

### LINKED ISSUE #150: GoRules — CCV orchestrator and form-to-rule-set convenience lookup
## Objective
Close the in-repo GoRules gaps per SRS v3.0 §3A.2.3. Production wiring of evaluate-schedule into a live mobile-app caller is explicitly out of scope (requires a mobile-app change outside this repo) — though 1.5 (#147) incidentally gave evaluate-schedule its first real production caller (visit-form-service's ANC regeneration flow).

## Sub-tasks
- [x] **4.1** — CCV schedule orchestrator: `generateCcvSchedule()` in `scheduleOrchestrator.ts` — RISK_STATE once, then CADENCE chained from the INC-to-CCV transition point, switching cadence at the 18-month boundary, stopping at the 24-month (DOB+730) EXIT point. The 25m CCV-HR extension is NOT generated here (whether the last visit had HR isn't known until it's conducted) — handled on-demand via the existing `generateCcvHrVisit`, same pattern as every other HR visit in this codebase.
- [x] **4.2** — Convenience endpoint: `GET /forms/:formCode/active-version/risk-rules` in visit-form-service resolves formCode → riskRuleSetId → rulesJson in one call, calling rules-service's `GET /admin/rules/:setId`. Required widening that route's role gate from ADMIN-only to SAKHI/SUPERVISOR/MANAGER/ADMIN (confirmed as a deliberate access-boundary change — rulesJson for a RISK rule set is business configuration a form already depends on, not admin-sensitive content).
- [x] **4.3** — `riskRuleSetId` now exposed on `GET /forms/:formCode/active-version`'s response (via a join on `formDefinition` in `findActiveVersion`), prerequisite for 4.2.

## Status: Complete

All 3 sub-tasks implemented, tested, and verified (`nx test/lint/build` clean on both `rules-service` and `visit-form-service`).

## Context
See prior completion-status audit in this session's conversation history for full evidence. Builds on the GoRules scheduling/escalation work merged in commit 03b1e8f, and on #147's schedule regeneration work.

Dharanishwaran-ND:
## Summary
- Implements GoRules decision graphs for ANC/PP/NN/INC/CCV scheduling and escalation (SRS v3.0 §3A.2.3), including the CCV schedule orchestrator and a full-ANC-schedule convenience endpoint.
- Closes the visit_schedules/visit_instances read/sync/lapse gaps: `GET /visit-schedules`, a properly scoped/paginated `GET /visits` (breaking change), `LAPSED` status, and bulk-lapse logic.
- Wires an approved LMP/EDD change (FR-SV-4.2) end-to-end to server-side ANC schedule regeneration via rules-service, beneficiary-service, and visit-form-service.
- Adds a form → risk-rule-set convenience endpoint for offline HR evaluation (FR-S-5.1), widening one rules-service route's role gate from ADMIN-only to all authenticated roles.
- Also includes the pre-existing Node 24.18.0 version pin this branch started from.

Closes #147
Closes #150

## Test plan
- [x] `nx test/lint/build` clean on `rules-service`, `visit-form-service`, `beneficiary-service`
- [x] `nx affected --target=lint,test` clean across the repo after each commit
- [ ] Manual smoke test of `POST /visit-schedules/regenerate-anc` end-to-end via the LMP-change approval flow (approval-service → beneficiary-service → visit-form-service → rules-service)


apps/visit-form-service/src/visit-schedules/visitSchedule.service.ts
```diff
@@ -0,0 +1,376 @@
...
+  async regenerateAncSchedule(
+    beneficiaryId: string,
+    registrationDate: string,
+    edd: string,
+    caller: CallerIdentity,
+    authorizationHeader: string,
+  ): Promise<{ supersededCount: number; created: number; schedules: CreatedScheduleResult[] }> {
+    await this.assertBeneficiaryAccess(beneficiaryId, caller, authorizationHeader);
+
+    const superseded = await this.repository.supersedeAnc(beneficiaryId, caller.id);
```
palani-ND, line 264: This commits immediately (plain `updateMany`), but the `evaluateAncScheduleFull` call two lines below isn't wrapped in try/catch and can throw `badGateway` on any rules-service hiccup — at which point this supersede has already committed and nothing creates the replacement rows. Since the caller (`beneficiary-service`'s `scheduleRegeneration.client.ts`) treats this whole endpoint as best-effort and just logs failures, a transient rules-service error here leaves the beneficiary with zero open ANC visits and no automatic retry. Wrapping supersede+evaluate+insert in one flow where the evaluate happens before any write commits (or in a single `$transaction`) would close this window.

apps/rules-service/src/rules/ruleSet.evaluator.ts
```diff
@@ -0,0 +1,291 @@
...
+ */
+export async function runDecisionGraph(
+  rulesJson: unknown,
+  input: Record<string, unknown>,
+): Promise<unknown> {
+  const engine = new ZenEngine();
+  const content = new ZenDecisionContent(rulesJson as object);
+  const decision = engine.createDecision(content);
+  const response = await decision.evaluate(input);
+  engine.dispose();
```
palani-ND, line 28: `engine.dispose()` only runs if `decision.evaluate()` resolves — a throwing evaluation (malformed rulesJson, category mismatch, etc.) skips disposal and leaks the native zen-engine instance. This was flagged on the single `evaluateRulePack` call site in PR #127; now that `evaluateScheduleRulePack` and `evaluateEscalationRulePack` also route through this shared helper, the same gap applies to all three. Worth a `try { ... } finally { engine.dispose(); }` here.

palani-ND (commented):
This is a big one — Node pin, the visit-form-service schedule sync/lapse/regeneration APIs, and a full GoRules-graph rewrite of the scheduling/escalation rule packs in rules-service. Traced the LMP-change-triggers-regeneration chain end to end (beneficiary-service → gateway → visit-form-service → rules-service) and personally verified the code below directly rather than relying on the earlier read-through alone. Also spot-checked whether the two issues flagged on PR #129's rulesJson-based CCV/ANC packs recur in the new GoRules `graphs/*.graph.ts` versions.

### Critical: an LMP/EDD correction can leave a beneficiary with zero open ANC visits, silently, with no retry

`VisitScheduleService.regenerateAncSchedule` (`visitSchedule.service.ts:255-289`) does, in order:
1. `this.repository.supersedeAnc(beneficiaryId, caller.id)` — a plain `updateMany`, commits immediately, marking every OPEN/GENERATED ANC-family row SUPERSEDED.
2. `evaluateAncScheduleFull(...)` — an HTTP call to rules-service, **not wrapped in try/catch**.
3. `this.repository.createAllOrNothing(...)` — inserts the new rows.

`evaluateAncScheduleFull` (`ruleVersion.client.ts:64-84`) throws `badGateway(...)` on any network failure or non-2xx from rules-service. That throw propagates straight through this method with nothing to compensate for step 1 having already committed — step 3 never runs. On the caller side, `beneficiary-service`'s `scheduleRegeneration.client.ts` treats this whole call as best-effort by design (its own doc comment says so explicitly: "the LMP/EDD write itself must succeed regardless of whether regeneration does... no retry/reconciliation mechanism exists anywhere in this codebase yet") and just logs the failure. So a single transient rules-service blip during an LMP-change approval leaves a real beneficiary with **no open/generated ANC visit schedule at all** — old rows superseded, new rows never created — and the only trace is a log line. Nothing polls for or retries this state.

Given `supersedeAnc` and `createAllOrNothing` are both single-purpose repository methods already, wrapping steps 1–3 in one Prisma `$transaction` (evaluate first, then supersede+insert together) would remove the window entirely — evaluate-then-write means a rules-service failure never touches the DB. `visitSchedule.service.spec.ts` only covers the happy path and a pre-write auth rejection; there's no test for "evaluate throws after supersede already committed," so this wouldn't be caught by the suite either.

### Medium: `engine.dispose()` still isn't in a `finally`, and now three evaluators share the leak instead of one

`ruleSet.evaluator.ts`'s new `runDecisionGraph()` helper (lines 20-30) is `evaluateRulePack`'s pre-existing engine-lifecycle pattern extracted into a shared function — `engine.dispose()` still runs only after `decision.evaluate()` resolves, with no try/finally. This was already flagged as a minor item on PR #127 for the single `evaluateRulePack` call site; this PR routes `evaluateScheduleRulePack` and `evaluateEscalationRulePack` through the same helper, so a throwing evaluation (malformed rulesJson, a category-mismatched rule set slipping through, etc.) now leaks a native zen-engine instance from three call sites instead of one. Worth fixing once here since it's now shared: `try { ... } finally { engine.dispose(); }`.

### CCV/ANC graph re-check (against the PR #129 findings)

The scheduling packs have been substantially redesigned since PR #129's `*.rulesJson.ts` versions (now real GoRules decision tables driven by `scheduleOrchestrator.ts`, one call per candidate visit rather than one big function per pack), so the two specific issues flagged there don't carry over as-is:
- The ANC graph no longer takes a `deliveryFormFiledDate`-derived flag at all — post-EDD scheduling is now offset/window arithmetic off `edd` directly (`anc.graph.ts`), so the old "can't distinguish not-yet-due from overdue" ambiguity doesn't apply to this shape of the logic (I didn't trace where/whether `scheduleOrchestrator.ts` gates on today's date before exposing a POST_EDD row — worth a quick look there specifically if that matters to whoever picks this up next, but the mechanism that produced the original bug is gone).
- The CCV risk-state table (`ccv.graph.ts:69-111`) now maps `last3AllAtRisk: true` to `RECENTLY_RECOVERED`'s faster cadence (`o2: '1'`) rather than falling through to `STABLE_LOW_RISK`'s slowest cadence — which is actually the direction I'd suggested erring in during the PR #129 follow-up for this exact ambiguous case. The `RECENTLY_RECOVERED` label for an "at risk" input is still a bit of a semantic mismatch on first read (data says at-risk, label says recovered), but functionally this looks like the safer choice was made.

I did not do a line-by-line audit of the full graph set (7 `*.graph.ts` files plus `scheduleOrchestrator.ts` and `escalation.graph.ts` — this is a lot of new decision-table surface) — this was a targeted check against the two specific prior findings, not a clean bill of health for the whole rewrite.

Everything else held up on inspection: `visitInstance.controller.ts`'s `GET /visits` role list now correctly includes ADMIN (fixing the earlier omission pattern); `resolveSakhiScope`'s SAKHI/SUPERVISOR/MANAGER/ADMIN scoping is applied consistently and the keyset pagination is correctly tie-broken; the `ruleVersion.controller.ts` role widening on `GET /admin/rules/:setId` (ADMIN-only → SAKHI/SUPERVISOR/MANAGER/ADMIN) reads as a deliberate, documented change for offline rulesJson caching rather than an accidental broadening.

PR #153, opened 2026-08-12
feat(media-service): Implement S3 presigned upload flow

### LINKED ISSUE #152: feat(media-service): Implement S3 presigned upload flow
## Objective
Implement the HLD-documented direct-to-S3 presigned-upload flow for `media-service` (per `docs/ARMMAN_ArogyaSakhi_High_Level_Design_v2.0.docx.md` §3.2, §6.2): the app requests a presigned S3 upload URL, uploads directly to S3, then confirms the upload with the backend — instead of the previous metadata-only `POST /media`, which required the caller to already have a `storageUri` with no actual S3 interaction anywhere in the codebase.

## Scope
- `POST /media/upload-url` — new endpoint issuing a presigned S3 `PutObject` URL under a freshly generated key, gated by asset-type/mimeType/size validation against configurable env limits.
- `POST /media` — extended to be the finalize step: verifies the upload actually landed via S3 `HeadObject`, cross-checks the declared size against the real object, and derives `storageUri`/`checksum`/`mimeType`/`sizeBytes` from S3 rather than trusting client input.
- New config: `S3_BUCKET_NAME`, `AWS_REGION`, `PRESIGNED_URL_EXPIRY_SECONDS`, `MAX_UPLOAD_SIZE_BYTES`, `ALLOWED_UPLOAD_MIME_TYPES`. No AWS credentials in code/`.env.example` — default credential provider chain only.
- First AWS SDK usage in this repo (`@aws-sdk/client-s3`, `@aws-sdk/s3-request-presigner`).

## Verified
Live end-to-end test against the real `arogyasakhi-dev-media` bucket: requested a presigned URL, uploaded via a plain HTTP PUT (no special SDK/headers required), finalized successfully, and confirmed both negative paths (size mismatch → 422, missing object → 400).

## Explicitly out of scope
- Virus scanning (S3 EventBridge → Lambda per HLD §6.2) — AWS infrastructure, not application code.
- S3 bucket/IAM Terraform provisioning.
- True end-to-end SHA-256 checksum enforcement — AWS SDK v3's presigner cannot reliably enforce an `x-amz-checksum-sha256` header against an arbitrary HTTP client (confirmed via a live test and an open, unresolved upstream issue: aws/aws-sdk-js-v3#3906). Integrity is instead verified via S3's own ETag (MD5) plus a declared-size cross-check.

Dharanishwaran-ND:
Closes #152

## Summary
- Implements the HLD-documented direct-to-S3 presigned-upload flow (`docs/ARMMAN_ArogyaSakhi_High_Level_Design_v2.0.docx.md` §3.2, §6.2) for `media-service`, replacing the previous metadata-only `POST /media` (which required the caller to already have a `storageUri`, with no real S3 interaction anywhere in the codebase).
- New `POST /media/upload-url` — issues a presigned S3 `PutObject` URL under a freshly generated key; validates `assetType`/`mimeType`/`sizeBytes` against configurable env limits before signing.
- `POST /media` is now the finalize step — calls S3 `HeadObject` to confirm the upload landed, cross-checks the declared size against the real object, and derives `storageUri`/`checksum`/`mimeType`/`sizeBytes` from S3 rather than trusting client input.
- New env config: `S3_BUCKET_NAME`, `AWS_REGION`, `PRESIGNED_URL_EXPIRY_SECONDS`, `MAX_UPLOAD_SIZE_BYTES`, `ALLOWED_UPLOAD_MIME_TYPES`. No AWS credentials in code or `.env.example` — relies on the SDK's default credential provider chain (IAM role in deployed environments).
- First AWS SDK usage in this repo (`@aws-sdk/client-s3`, `@aws-sdk/s3-request-presigner`).

## Design note: checksum enforcement
True end-to-end SHA-256 checksum enforcement was attempted and abandoned after live testing against a real bucket: AWS SDK v3's presigner cannot reliably enforce an `x-amz-checksum-sha256` header against an arbitrary HTTP client — `getSignedUrl` doesn't include `x-amz-checksum-*` in `X-Amz-SignedHeaders` by default, so the header can never actually be pinned into the signature this way (see the unresolved upstream issue [aws/aws-sdk-js-v3#3906](https://github.com/aws/aws-sdk-js-v3/issues/3906)). Integrity is instead verified via S3's own `ETag` (MD5 for a single-part `PutObject`, always present) plus a declared-size cross-check — weaker than true SHA-256, but it works with any HTTP client (verified with a plain `curl` PUT) and still catches a truncated/substituted upload.

## Explicitly out of scope
- Virus scanning (S3 EventBridge → Lambda per HLD §6.2) — AWS infrastructure, not application code.
- S3 bucket/IAM Terraform provisioning.

## Test plan
- [x] `npx nx test media-service` — 36/36 passing
- [x] `npx nx lint media-service` — clean
- [x] `npx tsc --noEmit` (app + spec tsconfigs) — clean
- [x] Live end-to-end test against the real `arogyasakhi-dev-media` bucket: requested a presigned URL, uploaded via plain `curl` PUT (no special headers), finalized successfully with correct `storageUri`/`sizeBytes`/checksum recorded.
- [x] Verified negative paths live: declared-size mismatch → `422`, object not found at key → `400`.


apps/media-service/src/media/s3.client.ts
```diff
@@ -0,0 +1,114 @@
...
+ * always present) plus a `ContentLength` match against what the client
+ * declared upfront — weaker than a true end-to-end SHA-256, but it works
+ * with any HTTP client and still catches a truncated or substituted upload.
+ */
+export async function getPresignedUploadUrl(input: {
+  key: string;
+  mimeType: string;
+  sizeBytes: number;
+}): Promise<{ uploadUrl: string; expiresInSeconds: number }> {
+  const command = new PutObjectCommand({
```
palani-ND, line 39: No `ContentLength` (or any size bound) is part of this signed command, so `MAX_UPLOAD_SIZE_BYTES`/the declared `sizeBytes` isn't actually enforced by S3 at upload time — a client can PUT a body far larger than what it declared here, up to S3's own single-PUT ceiling. `create()` will catch the mismatch on finalize (422), but by then the oversized object already exists in the bucket, and it's never cleaned up if the client just never finalizes. Worth a bucket lifecycle rule to expire un-finalized objects, since presigned PUT URLs can't pin body size into the signature the way a presigned POST policy's `content-length-range` condition could.

palani-ND (commented):
Reviewed the presigned-upload flow (s3.client.ts, both DTOs, the service/route wiring, tests). The design note on why SHA-256 pinning was abandoned is a good, honest writeup, and the finalize step correctly derives storageUri/mimeType/sizeBytes/checksum from S3's own HeadObject response rather than trusting client-declared values — closing off the obvious 'lie about what you uploaded' path. One real gap in the same family as the checksum issue you already documented, plus a couple of smaller things:

**`MAX_UPLOAD_SIZE_BYTES`/declared `sizeBytes` isn't actually enforced by S3 at upload time — only checked after the fact.** `getPresignedUploadUrl` signs a `PutObjectCommand` with no `ContentLength`, and `getSignedUrl` doesn't include content-length in what it signs for a plain `PutObjectCommand` (this is the same category of gap as the checksum one — S3 presigned PUT URLs can't pin body size into the signature; only presigned POST policies can, via a `content-length-range` condition). So the `sizeBytes` validated against `appConfig.MAX_UPLOAD_SIZE_BYTES` in `createUploadUrlSchema` is purely what the client *declares* when requesting the URL — nothing stops the actual PUT from sending a body far larger (up to S3's own ~5GB single-PUT ceiling). `mediaAsset.service.ts`'s `create()` will correctly reject the mismatch with a 422, but by then the oversized object is already sitting in the bucket, and if the client never calls `POST /media` to finalize at all, it's never cleaned up either (virus scanning / lifecycle policies are explicitly out of scope per this PR's description, but an unbounded-size, never-expiring orphan object is a direct consequence of that gap, not just a hypothetical). Given this is gated to authenticated SAKHI/SUPERVISOR callers the blast radius is limited to a compromised/malicious account or a buggy client rather than the open internet, but worth at least a bucket lifecycle rule to expire objects that are never finalized within, say, 24h, since that's pure infra config and doesn't require solving the presigned-URL signing limitation itself.

**Smaller:** `s3Key` at `POST /media` finalize time (`z.string().trim().min(1)`) isn't bound to the specific `/media/upload-url` request that issued it — nothing server-side records which caller requested which key, so knowledge of a valid key alone is sufficient to finalize it under any `linkedEntityType`/`beneficiaryId`/etc. The keys are `randomUUID()`-based (128 bits), so this isn't practically guessable, but it does mean a key is a bearer capability with no expiry — if one ever leaks (logs, a proxy, a compromised session), it stays claimable indefinitely. Also, no unique constraint on `storageUri` in the Prisma schema, so the same S3 object could in principle be finalized into more than one `media_asset` row — probably harmless given the flow's intended one-shot usage, just flagging since nothing currently prevents it.

Everything else held up: role gating (`SAKHI`/`SUPERVISOR`) is applied to both new/changed routes, the ETag-format guard correctly rejects multipart-upload ETags (which aren't plain MD5 and would silently mislabel a `checksum` column that isn't actually one), and `headObject`'s not-found-vs-any-other-failure distinction (missing object is an expected 400-worthy client state, anything else is a 502 upstream-dependency failure) is a sound split with good test coverage for every branch.

apps/media-service/src/media/dto/create-mediaAsset.dto.ts
```diff
@@ -15,30 +19,28 @@ const mediaAssetTypeSchema = z.enum([
...
+ * to upload" (e.g. a stale/wrong object left at a reused or guessed key).
  */
 export const createMediaAssetSchema = z
   .object({
     assetType: mediaAssetTypeSchema,
-    storageUri: z.string().trim().min(1).max(512),
-    checksum: checksumSchema,
-    mimeType: z.string().trim().min(1).max(120),
-    sizeBytes: z.coerce.bigint(),
+    s3Key: z.string().trim().min(1),
```
palani-ND: Nothing server-side records which caller requested this `s3Key` from `POST /media/upload-url`, so this field isn't validated as "a key this same caller was actually issued" — any authenticated SAKHI/SUPERVISOR who knows a valid key (however they came to know it) can finalize it under whatever `linkedEntityType`/`beneficiaryId`/etc. they supply here. Mitigated in practice by the key being a `randomUUID()` (128 bits of entropy, not guessable), but it is a bearer capability with no expiry — a leaked key (logs, a proxy, a compromised session) stays claimable indefinitely. Worth a min-length/format check at least (`z.string().trim().min(1)` accepts anything), and possibly worth recording which caller/request issued each key if this ever needs tighter binding.

palani-ND (commented):
Follow-up to my last review — inlining the second point from that summary that I'd only left in prose, since it's on a line that's actually part of this diff.

apps/media-service/src/media/dto/create-mediaAsset.dto.ts
Dharanishwaran-ND: Fixed in 7e513e7. s3Key now must match the exact shape generateObjectKey produces (`<folder>/<assetType>/<uuid>`, rejecting garbage/path-traversal), and mediaAsset.service.ts's create() additionally checks the key's own assetType segment matches the declared assetType before ever calling headObject — a legitimately-issued key finalized under a mismatched assetType is now rejected with a 400, verified live against the real bucket.

This doesn't fully close the gap you raised — a key is still a bearer capability with no expiry or server-side caller binding, since that would need request tracking (recording which caller/request issued each key), which is out of scope for this PR. But it does narrow what's exploitable: only keys this service actually issued, for the assetType they were issued for, can be finalized — not just "any non-empty string."

apps/media-service/src/media/s3.client.ts
Dharanishwaran-ND: Confirmed, and you've correctly identified the actual fix: this needs a presigned POST policy (content-length-range condition), not a presigned PUT — PUT URLs genuinely can't pin body size into the signature. Switching PutObjectCommand → a presigned POST would change the client's upload mechanism from a raw PUT to a multipart form POST, which is a real contract change for whatever's consuming this (mobile app), not a small fix — so I'm not making that swap in this PR without confirming it's wanted.

What I did fix in this same batch (7e513e7): tightened s3Key validation and added an assetType-match check on finalize (see the reply on your other comment) — different gap, same family.

For this one: agreed a bucket lifecycle rule to expire un-finalized objects (e.g. 24h) is the right mitigation given it's pure infra config and doesn't require solving the presigned-URL signing limitation. That's Terraform/AWS console work outside this repo's scope (per this repo's own rule: infra doesn't belong in this codebase) — flagging it as a follow-up for whoever owns the bucket's lifecycle policy, not closing it out here. Let me know if you'd rather I open a tracking issue for it.

PR #158, opened 2026-08-13
feat(visit-form-service): Add Delivery, Postpartum, and Neonatal visit forms

### LINKED ISSUE #157: feat(visit-form-service): Add Delivery, Postpartum, and Neonatal visit forms
## Summary
Add the three forms that make up the delivery-event session (per SRS v3.0 / `docs/Revised_App_Form_Final_20.3.26.xlsx.md`, "DeliveryPPNeonatal visit" section):

- **DELIVERY_VISIT** — delivery outcome/details for the mother, with flattened `child1_*`/`child2_*`/`child3_*` fields for up to 3 births, and auto-creation of a CHILD beneficiary case for each live-born child.
- **POSTPARTUM_VISIT** — the shared PP1–PP5 postpartum visit template.
- **NEONATAL_VISIT** — the shared NN1/NN2 neonatal visit template.

## Scope
- New `FormDefinition`/`FormVersion` seed data for all three forms (no schema migration — reuses the existing generic `FormSubmission`/`FormAnswer` tables).
- Cross-service call from `visit-form-service` to `beneficiary-service`'s existing `POST /beneficiaries` to auto-create each live-born child's case, best-effort (matches `syncSocioDemographics`/`triggerRiskAssessment`'s resilience stance).

## Out of scope (follow-up)
- PP1/NN1 session-sequencing logic (auto-opening NN1 immediately after Delivery+PP1 submission) is not implemented here — this issue covers the form definitions and submission validation only.

Dharanishwaran-ND:
Closes #157

## Summary
- Seeds three new form definitions per `docs/Revised_App_Form_Final_20.3.26.xlsx.md`'s "DeliveryPPNeonatal visit" section — the forms that make up the delivery-event session:
  - **DELIVERY_VISIT** (44 fields) — mother-level delivery details + flattened `child1_*`/`child2_*`/`child3_*` blocks (up to 3 births, gated by `number_of_babies_born`), plus remarks.
  - **POSTPARTUM_VISIT** (36 fields) — shared PP1–PP5 template.
  - **NEONATAL_VISIT** (32 fields) — shared NN1/NN2 template, plus two pre-filled fields (`birth_weight_kg`, `term_of_delivery`) so the KMC-eligibility gate can reference the Delivery form's own data without a cross-service lookup.
- No Prisma migration — reuses the existing generic `FormDefinition`/`FormVersion`/`FormSubmission`/`FormAnswer` tables, same as `MOTHER_REGISTRATION`/`ANC_VISIT`.
- `DELIVERY_VISIT` submission auto-creates a `CHILD` beneficiary case (via beneficiary-service's existing `POST /beneficiaries`) for each live-born child, best-effort — matches the resilience stance already used for `syncSocioDemographics`/`triggerRiskAssessment`. Stillbirths are correctly skipped.
- `beneficiary.client.ts`'s `findBeneficiaryById` now also returns the mother's geography/lookup ids, so the auto-created child case inherits them rather than requiring a separate resolution step.
- `acknowledgeDuplicate: true` is passed on the child-create call — without it, twins/triplets sharing the same mother + date of birth would false-positive beneficiary-service's duplicate-beneficiary check.

## Design note
The Child1/2/3 block is implemented as flattened, individually-gated fields (`childN_*`, `visibleWhen: number_of_babies_born >= N`) rather than a new repeating-group schema primitive — this needed no changes to `form-field.dto.ts`/`form-validation.ts`/`form.mapper.ts`, keeping the change scoped to seed data + one new cross-service client + one new branch in `FormService.createSubmission()`.

## Explicitly out of scope
- PP1/NN1 auto-sequencing within the same Sakhi session (opening PP1/NN1 immediately after a Delivery submission) — this PR only adds the three form definitions and their submission validation, not the session-orchestration logic. Tracked as a follow-up.

## Test plan
- [x] `npx nx test visit-form-service` — 310/310 passing (added coverage: `seed-data.spec.ts` round-trips all 3 new seed files against `schemaJsonSchema`/`validationJsonSchema` + field-count/gating assertions; `form.service.spec.ts` covers single-birth, twin, stillbirth, missing-mother-case, and non-DELIVERY_VISIT-form cases for the new child-creation branch; `create-child.client.spec.ts` covers the POST payload shape, duplicate-acknowledgement, and best-effort failure handling)
- [x] `npx tsc --noEmit` (app + spec tsconfigs) — clean
- [x] Live end-to-end verification against the real dev DB/gateway:
  - `prisma:seed` — created all 3 forms, idempotent on rerun
  - `GET /forms/{DELIVERY_VISIT,POSTPARTUM_VISIT,NEONATAL_VISIT}/active-version` — 200, correct field counts (44/36/32)
  - `POST /forms/DELIVERY_VISIT/submissions` — valid single birth (201, child auto-created), valid twins (201, both children auto-created with distinct cases), stillbirth (201, no child created), missing-mandatory-field (422 with precise violations), out-of-range value (422)

### DIFF
--- a/apps/visit-form-service/src/forms/form.service.ts
+++ b/apps/visit-form-service/src/forms/form.service.ts
@@ -13,6 +13,8 @@ import {
 import { validateSubmission } from './form-validation';
 import { syncSocioDemographics } from '../beneficiaries/socio-demographics.client';
 import { syncHealthHistory } from '../beneficiaries/health-history.client';
+import { findBeneficiaryById } from '../beneficiaries/beneficiary.client';
+import { createChildBeneficiary } from '../beneficiaries/create-child.client';
 import { getAncestorChain } from '../geography/geography.client';
 import { triggerRiskAssessment } from '../risk-assessments/riskAssessment.client';
 
@@ -195,6 +197,62 @@ export class FormService {
     if (formCode === 'MOTHER_REGISTRATION') {
       await syncSocioDemographics(dto.beneficiaryId, dto.formData, authorizationHeader);
       await syncHealthHistory(dto.beneficiaryId, dto.formData, authorizationHeader);
+    }
+
+    // Auto-creates a CHILD beneficiary case for each live-born child on a
+    // Delivery submission (SRS: "Delivery form... auto-creates the child
+    // profile on submission") — one createChildBeneficiary() call per
+    // childN_* block whose delivery outcome is LIVE_BIRTH, independent and
+    // best-effort per child so one failing call (e.g. in a twin/triplet
+    // birth) never blocks the others or the already-saved submission.
+    if (formCode === 'DELIVERY_VISIT') {
+      const motherCase = await findBeneficiaryById(dto.beneficiaryId, authorizationHeader);
+      if (motherCase) {
+        const dateOfDelivery = dto.formData.date_of_delivery
+          ? new Date(String(dto.formData.date_of_delivery))
+          : null;
+        const childPrefixes = ['child1', 'child2', 'child3'] as const;
+        const SEX_MAP: Record<string, 'MALE' | 'FEMALE' | 'INTERSEX_OTHER'> = {
+          male: 'MALE',
+          female: 'FEMALE',
+          intersex_other: 'INTERSEX_OTHER',
+        };
+
+        await Promise.all(
+          childPrefixes.map(async (prefix) => {
+            const outcome = dto.formData[`${prefix}_delivery_outcome`];
+            if (outcome !== 'live_birth' || !dateOfDelivery) return;
+
+            const sexCode = dto.formData[`${prefix}_sex_of_baby`];
+            const birthWeightKg = dto.formData[`${prefix}_birth_weight_kg`];
+            const birthLengthCm = dto.formData[`${prefix}_birth_length_cm`];
+
+            await createChildBeneficiary(
+              {
+                motherCase,
+                // Deterministic and independent of localSubmissionUuid — a
+                // resubmit of the same Delivery form (idempotent replay,
+                // handled above) always derives the same per-child key, so a
+                // retry after a partial failure re-attempts only the
+                // children beneficiary-service hasn't already created.
+                localCaseUuid: `${dto.localSubmissionUuid}-${prefix}`,
+                registrationDate: dateOfDelivery,
+                dateOfBirth: dateOfDelivery,
+                sex: typeof sexCode === 'string' ? SEX_MAP[sexCode] : undefined,
+                birthWeightKg:
+                  typeof birthWeightKg === 'number' || typeof birthWeightKg === 'string'
+                    ? Number(birthWeightKg)
+                    : undefined,
+                birthLengthCm:
+                  typeof birthLengthCm === 'number' || typeof birthLengthCm === 'string'
+                    ? Number(birthLengthCm)
+                    : undefined,
+              },
+              authorizationHeader,
+            );
+          }),
+        );
+      }
     }
 
     // Triggers the risk-grading pipeline for every visit-linked submission
--- a/apps/visit-form-service/src/forms/form.service.ts
+++ b/apps/visit-form-service/src/forms/form.service.ts
@@ -204,7 +204,13 @@ export class FormService {
     // profile on submission") — one createChildBeneficiary() call per
     // childN_* block whose delivery outcome is LIVE_BIRTH, independent and
     // best-effort per child so one failing call (e.g. in a twin/triplet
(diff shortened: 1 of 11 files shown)


apps/visit-form-service/src/forms/form.service.ts
```diff
@@ -197,6 +199,62 @@ export class FormService {
...
+
+            const sexCode = dto.formData[`${prefix}_sex_of_baby`];
+            const birthWeightKg = dto.formData[`${prefix}_birth_weight_kg`];
+            const birthLengthCm = dto.formData[`${prefix}_birth_length_cm`];
+
+            await createChildBeneficiary(
+              {
+                motherCase,
+                // Deterministic and independent of localSubmissionUuid — a
+                // resubmit of the same Delivery form (idempotent replay,
```
palani-ND: This claims a resubmit re-attempts only the still-missing children, but `createSubmission`'s idempotent-replay check (`if (existing) return toApiFormSubmission(existing);`, a few lines up) returns before this block ever runs again — a client resubmit with the same `localSubmissionUuid` can't reach this code at all. If `createChildBeneficiary` fails for one child here, there's no path — resubmit or otherwise — that ever retries it; the child case is permanently missing unless someone fixes it out of band. Either the comment should say so plainly, or this needs an actual reconciliation mechanism (a job scanning for `live_birth` slots with no matching case, using this same deterministic `localCaseUuid`) since a client-side resubmit doesn't provide one.

palani-ND (commented):
Reviewed the seed data (field counts/gating verified against the test assertions), the new `create-child.client.ts`, and the new `FormService.createSubmission()` branch. The best-effort child-creation design (matching `syncSocioDemographics`/`triggerRiskAssessment`'s established stance), the `acknowledgeDuplicate: true` twin/triplet handling, and the stillbirth skip are all correctly implemented and well tested (single birth, twins, stillbirth, missing-mother-case, non-DELIVERY_VISIT cases all covered). One real gap:

**The retry-recovery story this code's own comment describes doesn't actually happen.** The `localCaseUuid` doc comment says: "a resubmit of the same Delivery form (idempotent replay, handled above) always derives the same per-child key, so a retry after a partial failure re-attempts only the children beneficiary-service hasn't already created." But `createSubmission`'s idempotent-replay check (`const existing = await this.repository.findSubmissionByLocalUuid(...); if (existing) return toApiFormSubmission(existing);`, a few lines above this PR's diff) returns *before* the submission is ever re-validated or re-processed — including before this whole `if (formCode === 'DELIVERY_VISIT')` block. So if `createChildBeneficiary` fails for one twin (beneficiary-service momentarily down, a network blip) while the submission itself and the other twin's case both save fine, a resubmit with the same `localSubmissionUuid` short-circuits at that early return and never reaches this code again — the missing child case stays missing forever, with nothing surfaced anywhere (per the best-effort design, only a `console.warn`). This is a real newborn with no beneficiary case, no NN/INC visit scheduling, and no automatic path to fix it — a meaningfully bigger gap than the `syncSocioDemographics` precedent this pattern is modeled on (a missing socio-demographic field is a data-quality issue; a missing beneficiary case means the child doesn't exist in the system for follow-up at all).

No test exercises "resubmit after one child's creation failed" either, which is consistent with the comment's claimed behavior not actually being wired up anywhere. Worth either (a) fixing the comment to honestly state this is a genuine one-shot best-effort attempt with no automatic retry — matching whatever the accepted reality already is for `syncSocioDemographics` — or (b) if the retry-safety really is wanted, adding an actual reconciliation path (e.g. a job that scans DELIVERY_VISIT submissions for `childN` slots with a `live_birth` outcome but no matching case, using this same deterministic key) rather than relying on a client resubmit that structurally can't reach this code.

apps/visit-form-service/src/forms/form.service.ts
logeshn-1: Fixed in 3f7232d — corrected the misleading comments in both `form.service.ts`'s `DELIVERY_VISIT` branch and `create-child.client.ts`'s `localCaseUuid` doc. They now honestly state this is a one-shot best-effort attempt (same stance as `syncSocioDemographics`), and clarify that `localCaseUuid`'s real purpose is de-duping a dropped-connection retry of the same HTTP call, not enabling a form-resubmit retry path — since `createSubmission`'s idempotent-replay check returns before this branch is ever reached again on a resubmit.

Went with option (a) rather than adding a reconciliation job, since a scanning/reconciliation mechanism would be new scope beyond what this PR's `syncSocioDemographics`/`triggerRiskAssessment` precedent already accepts, and PR 158 explicitly scopes session-orchestration follow-ups out. Happy to file a follow-up issue for a reconciliation job if that's wanted.

PR #160, opened 2026-08-17
feat(auth-service): non-expiring login sessions except for admin

logeshn-1:
## Summary

Updated login sessions so that normal users don't have token expiry, while ADMIN tokens still expire.

## Changes

* Normal user access tokens are now non-expiring.
* ADMIN access tokens still expire after the configured time.
* Refresh sessions no longer expire based on time.
* Updated JWT handling and related tests.

### DIFF
--- a/apps/auth-service/prisma/schema.prisma
+++ b/apps/auth-service/prisma/schema.prisma
@@ -117,7 +117,12 @@ model UserSession {
   // device_registry.device_id — deferred; not modeled yet per current scope.
   deviceId         String?   @map("device_id")
   issuedAt         DateTime  @map("issued_at")
-  expiresAt        DateTime  @map("expires_at")
+  // Nullable: sessions no longer expire by time (see AuthService.issueTokens)
+  // — only an explicit /auth/logout or admin revocation sets revokedAt below.
+  // Null on every session created after that change; a pre-existing session
+  // from before it keeps its original real timestamp until its next refresh
+  // rotates it into a fresh, non-expiring one.
+  expiresAt        DateTime? @map("expires_at")
   revokedAt        DateTime? @map("revoked_at")
   ipAddress        String?   @map("ip_address") @db.VarChar(45)
   createdAt        DateTime  @default(now()) @map("created_at")
--- a/apps/auth-service/src/auth/auth.routes.ts
+++ b/apps/auth-service/src/auth/auth.routes.ts
@@ -50,7 +50,7 @@ const authTokensSchema = z.object({
       description:
         'Access token lifetime in seconds, or null if it never expires. Null for every role ' +
         'except ADMIN — only /auth/logout or revocation ends a non-expiring session.',
-      example: null,
+      examples: [null, 900],
     }),
   roles: z
     .array(z.string())
--- a/apps/auth-service/src/auth/auth.routes.ts
+++ b/apps/auth-service/src/auth/auth.routes.ts
@@ -35,11 +35,23 @@ const createUserRequestSchema = createUserSchema.extend({
 });
 
 const authTokensSchema = z.object({
-  accessToken: z.string().openapi({ description: 'Short-lived JWT access token (RS256).' }),
+  accessToken: z.string().openapi({
+    description:
+      'JWT access token (RS256). Non-expiring (no exp claim) for every role except ADMIN, ' +
+      'which gets a short-lived expiry (JWT_ADMIN_ACCESS_TOKEN_TTL).',
+  }),
   refreshToken: z
     .string()
     .openapi({ description: 'Opaque refresh token; single-use, rotated on refresh.' }),
-  expiresIn: z.number().openapi({ description: 'Access token lifetime in seconds.', example: 900 }),
+  expiresIn: z
+    .number()
+    .nullable()
+    .openapi({
+      description:
+        'Access token lifetime in seconds, or null if it never expires. Null for every role ' +
+        'except ADMIN — only /auth/logout or revocation ends a non-expiring session.',
+      example: null,
+    }),
   roles: z
     .array(z.string())
     .openapi({ description: 'Every role code the caller holds.', example: ['SAKHI'] }),
--- a/apps/auth-service/.env.example
+++ b/apps/auth-service/.env.example
@@ -12,8 +12,10 @@ DIRECT_URL=postgresql://armman:armman@localhost:5432/arogya
 REDIS_URL=redis://localhost:6379
 JWT_PRIVATE_KEY="-----BEGIN PRIVATE KEY-----\nREPLACE_ME\n-----END PRIVATE KEY-----"
 JWT_PUBLIC_KEY="-----BEGIN PUBLIC KEY-----\nREPLACE_ME\n-----END PUBLIC KEY-----"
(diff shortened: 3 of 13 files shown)


Dharanishwaran-ND (commented):
## Review: non-expiring login sessions except for admin

Ran the full auth-service (241/241) and service-commons (54/54) suites against this branch — all green. Solid, well-documented change overall; one release-blocker and a couple of smaller gaps below.

### 🔴 Blocking: missing Prisma migration for `expires_at` becoming nullable

`schema.prisma` changes `UserSession.expiresAt` from `DateTime` to `DateTime?`, but no corresponding file was added under `apps/auth-service/prisma/migrations/`. CD runs `prisma migrate deploy` (see `.github/workflows/cd-dev.yml`), which only applies committed migration files — it will **not** pick up this schema change.

Concretely: `AuthRepository.createSession`/`.replaceSession` no longer pass `expiresAt` (see the diff removing it from both call signatures), so as soon as this deploys, every `/auth/login` and `/auth/refresh` call will try to insert a row with `expires_at = NULL` into a column that's still `NOT NULL` in the actual database — this will hard-fail login/refresh in every environment until the migration is generated and applied.

**Ask:** run `prisma migrate dev --name make_user_session_expires_at_nullable` (or equivalent) and commit the generated migration before merge.

### 🟡 `.env.example` is now stale

`apps/auth-service/.env.example` still lists the two removed vars:
```
JWT_ACCESS_TOKEN_TTL=15m
JWT_REFRESH_TOKEN_TTL=30d
```
and is missing the new `JWT_ADMIN_ACCESS_TOKEN_TTL` (it does have a code default of `15m` in `app-config.ts`, so this isn't deploy-blocking, but per the repo's own standard — "`.env.example` lists every variable a service needs" — it should be updated to match.

### 🟡 Cross-service inconsistency worth a note (not necessarily a blocker)

Every other service's `isPrivileged()` helper (sakhi.service.ts, project.service.ts, visitInstance.service.ts, visitSchedule.service.ts, operations.service.ts) treats **MANAGER and ADMIN** as the "privileged" tier. This PR's `EXPIRING_ROLE` set is ADMIN-only — MANAGER gets the same non-expiring token as SAKHI/SUPERVISOR. That may well be intentional (MANAGER is a normal field/ops role that shouldn't get logged out either), but it's a divergence from the existing privilege convention elsewhere in the codebase, so worth confirming that's deliberate rather than an oversight.

### 🟢 Minor: OpenAPI example regression

`authTokensSchema.expiresIn` dropped its `example: 900` in favor of `example: null`. Since this field is genuinely bimodal (a real number for ADMIN, `null` for everyone else), the Swagger UI docs no longer illustrate what an ADMIN login response looks like. Consider `example: 900` with a note that non-ADMIN responses are `null`, or vice versa — either way, purely cosmetic.

### 🟢 Confirmed as intentional / working as designed

- `PublicKeyVerifier`/`jwtVerify` (jose) does not require an `exp` claim by default, so a token with no `exp` verifies and never time-expires correctly — confirmed via the new `token-signer.spec.ts` test and by inspection.
- `AuthService.refresh()` re-fetches the user's current roles from the DB on every refresh (not from stale claims), so a demoted ADMIN correctly loses the short expiry, and a promoted user correctly gains it, on their next refresh.
- The commit message already flags the real residual risk accurately: a non-expiring access token can't be revoked early (self-verifying JWT) — only the refresh/session can, via logout — so a compromised or should-be-deactivated non-ADMIN account keeps working access until the token owner logs out or an admin explicitly revokes the session. Recommended follow-up (admin "revoke all sessions" action) is a good idea to actually build, not just leave as a comment.
- No other code path reads `UserSession.expiresAt` outside auth-service, and no other call site of `TokenSigner.sign()` exists — the optional-param change is safe.

Nice test coverage on the ADMIN-vs-multi-role-including-ADMIN case and the pre-existing-session backward-compat case.

apps/auth-service/prisma/schema.prisma
```diff
@@ -117,7 +117,12 @@ model UserSession {
   // device_registry.device_id — deferred; not modeled yet per current scope.
   deviceId         String?   @map("device_id")
   issuedAt         DateTime  @map("issued_at")
-  expiresAt        DateTime  @map("expires_at")
+  // Nullable: sessions no longer expire by time (see AuthService.issueTokens)
+  // — only an explicit /auth/logout or admin revocation sets revokedAt below.
+  // Null on every session created after that change; a pre-existing session
```
Dharanishwaran-ND, line 122: Blocking: no migration file was added for this `DateTime` → `DateTime?` change. CD runs `prisma migrate deploy`, which only applies committed migration files, so this schema edit alone won't reach the actual database. Since `createSession`/`replaceSession` no longer pass `expiresAt`, every login/refresh will try to write `expires_at = NULL` into a column that stays `NOT NULL` until a migration is generated and committed — please add one before merge.

apps/auth-service/src/auth/auth.routes.ts
```diff
@@ -35,11 +35,23 @@ const createUserRequestSchema = createUserSchema.extend({
...
+  expiresIn: z
+    .number()
+    .nullable()
+    .openapi({
+      description:
+        'Access token lifetime in seconds, or null if it never expires. Null for every role ' +
+        'except ADMIN — only /auth/logout or revocation ends a non-expiring session.',
+      example: null,
+    }),
   roles: z
```
Dharanishwaran-ND, line 55: Minor/cosmetic: dropping `example: 900` in favor of `example: null` means the Swagger UI docs no longer show what a real (ADMIN) expiring-token response looks like, since this field is genuinely bimodal — a number for ADMIN, null for everyone else. Not blocking, just worth a thought if the OpenAPI docs are used as a reference for FE/mobile integration.

apps/auth-service/prisma/schema.prisma
logeshn-1: Fixed in ed2e30a — added `prisma/migrations/20260817130000_make_user_session_expires_at_nullable/migration.sql` (`ALTER TABLE user_sessions ALTER COLUMN expires_at DROP NOT NULL`).

apps/auth-service/src/auth/auth.routes.ts
logeshn-1: Fixed in ed2e30a — expiresIn now documents both cases via `examples: [null, 900]` instead of dropping the ADMIN example.

logeshn-1:
Addressed the remaining review points:

- **.env.example stale**: fixed in e655222 — removed the two dead vars (`JWT_ACCESS_TOKEN_TTL`, `JWT_REFRESH_TOKEN_TTL`) and added `JWT_ADMIN_ACCESS_TOKEN_TTL` to match `app-config.ts`.
- **MANAGER vs ADMIN divergence**: intentional, not an oversight. `EXPIRING_ROLE` is deliberately ADMIN-only here — MANAGER is a field/ops role that shouldn't be logged out on a timer any more than SAKHI/SUPERVISOR should, unlike the `isPrivileged()` convention elsewhere which groups MANAGER+ADMIN for authorization scope, not session-expiry policy. Leaving as-is.

Both blocking/inline comments (missing migration, expiresIn example) are already fixed per the replies on those threads (ed2e30a).

PR #161, opened 2026-08-17
Master-data endpoints + Beneficiary Data Download API set

logeshn-1:
## Summary

Two combined feature branches, merged into one PR since `feature/beneficiary-data-download` branched from `feature/master-data-api`'s tip (its history is fully included):

**Master-data / lookup work** (auth-service, risk-referral-service, visit-form-service, supervisor-operations-service):
- `project-geography`, `application-parameters` endpoints (auth-service)
- Dedicated-path aliases over the generic lookup store (`/risk-categories`, `/visit-categories`, `/item-categories`, etc.)
- `RiskParameter` split out of `RiskCondition` (own model + migration)
- `VisitMaster` catalog (own model + migration), replacing a broken alias that duplicated `/visit-categories`
- `/item-master-list` alias for `/inventory-items`

**Beneficiary Data Download screen** (auth-service, risk-referral-service, visit-form-service, sync-service, supervisor-operations-service, beneficiary-service, api-gateway):
- `arogya-sakhi-roster`, `registration-targets` (new `SakhiRegistrationTarget` model + migration)
- `beneficiaries/:id/risk`, `risk-referrals` (+ details) — pure projections, no new schema
- `beneficiaries/:id/visits`, `sync/pending`, `gatherings` (list/training-marks/images)
- `item-transactions/:id/details` (a genuinely new single-transaction lookup — didn't exist before)
- `sakhi-calls`, `risk-monitoring` — thin aliases over existing data

## Known caveats
- `risk-monitoring` is aliased to `GET /beneficiaries/risk-summary` as a **best guess** — "Risk Monitoring" has no definition anywhere in the SRS/ERD/HLD. Flagged in the route's own OpenAPI summary as UNCONFIRMED.
- Two independent build passes flagged the same gap: `GET /gatherings?sakhiId=` and the new beneficiary-risk/risk-referral reads are role-gated only, with no SAKHI-own-beneficiary / SUPERVISOR-roster ownership check (unlike sibling `by-sakhi` endpoints elsewhere in the codebase that do enforce it). Worth a look before this lands.
- `GET /arogya-sakhi-roster` excludes PAN/Aadhaar/bank-account fields by design (no documented need for that PII in an offline roster download).
- `application-parameters` and the new `registration-targets` tables are intentionally unseeded — no source data exists to seed them correctly.

## Test plan
- `npx tsc --noEmit` clean on every touched service + api-gateway
- Full Jest suites green on every touched service (service-commons, auth-service, risk-referral-service, visit-form-service, sync-service, supervisor-operations-service, beneficiary-service, api-gateway)
- Every new/aliased endpoint verified live end-to-end through a locally running api-gateway with real JWTs, including the gateway's more-specific-prefix-before-generic-prefix routing for the `/beneficiaries/:id/{risk,risk-referrals,visits}` sub-resources (verified with a standalone Express test before wiring, not just assumed)
- Migrations for `RiskParameter`, `VisitMaster`, and `SakhiRegistrationTarget` are new files in each service's own `prisma/migrations/` — `prisma migrate deploy` (which `cd-dev.yml` runs on merge) will apply them for real against whatever DB each deployed service's `.env` targets


Dharanishwaran-ND (commented):
## Review: Master-data endpoints + Beneficiary Data Download API set

Ran the full test/typecheck suite for every touched service — all green, matching the PR's own claim (auth-service 265 passed/18 skipped, risk-referral-service 65 passed, visit-form-service 314 passed/8 skipped, sync-service 11 passed, supervisor-operations-service 207 passed, beneficiary-service 189 passed, api-gateway 28 passed, incentive-wages-service 21 passed; `tsc --noEmit` clean across all). Migrations for `RiskParameter`, `VisitMaster`, and `SakhiRegistrationTarget` all have matching, content-correct files — no gaps found (sync-service/supervisor-operations-service/beneficiary-service correctly have no schema diff, since those are pure read-projections).

### Your 4 flagged caveats, checked against the code

1. **`risk-monitoring` UNCONFIRMED alias** — confirmed accurate. Reuses `getRiskSummary`/`summaryQuerySchema` verbatim, same roster-scoping as `/beneficiaries/risk-summary`. No security concern, just labeled uncertainty about product fit — that's the right way to ship an aliased-guess endpoint.
2. **`gatherings`/beneficiary-risk/risk-referral reads unscoped** — confirmed, and it's a bit broader than described (see below).
3. **`arogya-sakhi-roster` excludes PAN/Aadhaar/bank** — confirmed. `toApiRosterEntry` projects only `id, userId, displayName, username, mobileNumber, employeeCode, supervisorId, primaryProjectId, activeFrom, activeTo, status` — no PAN/Aadhaar/bank/password fields. Also project-scoped for SUPERVISOR and role-gated away from SAKHI.
4. **`application-parameters`/`registration-targets` unseeded** — confirmed, and gracefully handled: both `list()` methods return `[]` on an empty table rather than throwing.

### 🔴 The ownership-check gap is real and extends further than the PR body says

You already called out `gatherings` and the beneficiary-risk/risk-referral reads. Tracing the actual code, here's the full picture, plus two more instances not mentioned in the PR body:

- **`GET /beneficiaries/:beneficiaryId/risk`** (`beneficiaryRisk.controller.ts` → `getRiskProfile(req.params.beneficiaryId)`) and **`risk-referrals`/`risk-referrals/:id/details`** (`beneficiaryRiskReferral.service.ts`'s `listReferrals`/`getReferralDetails`) — no ownership check at all, role gate is `SAKHI, SUPERVISOR, MANAGER, ADMIN`. Any authenticated SAKHI can pass an arbitrary `beneficiaryId` not in her caseload and read that beneficiary's full risk/referral history.
- **`GET /gatherings?sakhiId=`** (see inline comment) — its own sibling methods in the same service/class (`listGatheringTopics`, `getGatheringAttendance`, `getGatheringImages`) all enforce `event.supervisorId !== caller.id && !isPrivileged(caller)` → 403, but `listGatherings` skips that entirely — the controller doesn't even pass `req.user` through.
- **Not mentioned in the PR body, same defect class — `GET /sync/pending?userId=`** (sync-service): any SAKHI/SUPERVISOR can pass an arbitrary `userId` and read that user's outstanding sync queue (device id, entity ids, retry/error state). Only defaults to the caller's own id when `userId` is omitted — there's no check that a *supplied* id belongs to the caller.
- **Also not mentioned — `GET /registration-targets?sakhiId=`**: `authenticate(signer)` only, no `requireRoles`, no ownership check on the supplied `sakhiId`. Lower severity (numeric targets only, no PII), but same missing-scoping pattern.
- **Pre-existing, not introduced by this PR, but worth noting**: the new `GET /beneficiaries/:beneficiaryId/visits` explicitly documents itself as unscoped "mirroring `GET /visits`' existing behavior" — and `GET /visits` (pre-existing, `findMany()` with no filter) is indeed already globally unscoped. The new endpoint doesn't make anything worse, it just adds a second door to an already-unlocked room — but since this PR is touching visit-read endpoints anyway, it might be worth fixing both at once rather than only the new one later.

**What's correctly scoped and doesn't need changes**: `sakhi-calls` (reuses `listCallLogsBySakhi`, which does check `sakhi.supervisorId !== caller.id`) and `item-transactions/:id/details` (reuses `getInventoryTransactionById`, which checks `transaction.supervisorId !== caller.id`) — both are thin aliases over already-scoped logic, unlike `gatherings`/beneficiary-risk which are net-new unscoped reads.

**Ask**: either add the `resolveSakhiScoping`-equivalent ownership check consistently across `gatherings`, `beneficiaries/:id/risk`, `risk-referrals`, `sync/pending`, and `registration-targets` (matching the pattern `listCallLogsBySakhi`/`getInventoryTransactionById` already use), or make it an explicit, documented product decision that these are intentionally caller-unscoped bulk-download endpoints — the way `risk-monitoring` and `arogya-sakhi-roster` already do transparently in their own OpenAPI summaries/comments. Right now it reads as a silent gap rather than a decision.

### 🟡 Minor: "standalone Express test" claim for gateway routing isn't captured as a regression test

The PR body says the trickier gateway change (reworking `pathRewrite` to support `:param`-containing mount paths for `/beneficiaries/:beneficiaryId/{risk,risk-referrals,visits}`) was "verified with a standalone Express test before wiring, not just assumed." No such test file exists in the diff or anywhere in `apps/api-gateway/src/proxy/` history. The route ordering itself is correct on manual inspection (more-specific prefixes registered before the generic `/beneficiaries` mount), so this isn't blocking — but a future change to route ordering or the rewrite logic has no automated guard against regressing it. Worth adding a real spec for this if it's easy.

### 🟢 Nits
- `gatherings.routes.ts` (335 lines) and `operations.service.ts` (773 lines) are both over the repo's informal ~250-line guideline — mostly organic growth, not something this PR introduced badly, but `gatherings.routes.ts` in particular could split its response-schema definitions out.
- `callLogs.routes.ts`'s comment says `sakhi-calls` uses "the same handler" as the by-params route — it's actually a distinct `listBySakhiQuery` function reading from query instead of params. Functionally equivalent and correctly scoped either way, just a slightly inaccurate comment.

No cross-service imports found (forklift rule respected), no `any` types/TODOs/console.log in new files, `.strict()` applied consistently on new query/params schemas.

apps/risk-referral-service/src/beneficiary-risk/beneficiaryRisk.controller.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Beneficiary risk profile request handlers. Mounted under the global
+ * `api/v1` prefix by `beneficiaryRisk.routes.ts`.
+ */
+export function createBeneficiaryRiskController(service: BeneficiaryRiskService) {
+  return {
+    getRiskProfile: asyncHandler(async (req, res) => {
+      const profile = await service.getRiskProfile(req.params.beneficiaryId);
+      res.json(ok(profile));
```
Dharanishwaran-ND, line 19: No ownership check: `req.params.beneficiaryId` is passed straight through with no verification it belongs to the caller's own caseload/roster. Role gate is SAKHI/SUPERVISOR/MANAGER/ADMIN, so any authenticated SAKHI can call this for a beneficiary who isn't hers and read her full risk-assessment/flag history. Same gap applies to `beneficiaryRiskReferral.service.ts`'s `listReferrals`/`getReferralDetails` right next door. Suggest adding the same ownership check the sibling `gatherings`/call-log/inventory endpoints use elsewhere in this codebase (intersect against the caller's own scope: SAKHI -> own beneficiaries only, SUPERVISOR -> roster, MANAGER/ADMIN -> unscoped) — or if this is intentionally an unscoped bulk-download endpoint, please say so explicitly in the route's OpenAPI summary the way risk-monitoring/arogya-sakhi-roster already do.

apps/supervisor-operations-service/src/operations/gatherings.controller.ts
```diff
@@ -9,6 +9,10 @@ import type { topicMarkQuerySchema } from './dto/topic-mark-query.dto';
  */
 export function createGatheringsController(service: OperationsService) {
   return {
+    list: asyncHandler(async (req, res) => {
```
Dharanishwaran-ND: This handler doesn't pass `req.user` through at all, unlike every sibling handler in this same file (`listTopics`, `getAttendance`, `getTrainingMarks`, `getImages`), which all check `req.user` and enforce `event.supervisorId !== caller.id && !isPrivileged(caller)` -> 403 in `operations.service.ts`. As written, any SUPERVISOR can pass `?sakhiId=<another supervisor's sakhi>` and enumerate that Sakhi's full gathering/training history — bypassing the exact ownership model enforced one line away in the same service for the gathering's sub-resources. Please add the same ownership check to `listGatherings` that its siblings already have.

apps/sync-service/src/sync/syncPending.controller.ts
```diff
@@ -0,0 +1,21 @@
...
+export function createSyncPendingController(service: SyncPendingService) {
+  return {
+    list: asyncHandler(async (req, res, next) => {
+      // trustGatewayIdentity always runs first on this route and calls
+      // next(unauthorized()) on its own failure, so req.user is populated by
+      // the time this handler runs; the check here is belt-and-braces so a
+      // future route wiring change can't silently drop it.
+      if (!req.user) return next(unauthorized());
+      const { userId } = req.query as unknown as SyncPendingQuery;
+      res.json(ok(await service.listPending(userId, req.user.id)));
```
Dharanishwaran-ND: `userId` from the query string is passed straight to `listPending` with no check that it belongs to the caller — it only DEFAULTS to `req.user.id` when omitted (see `syncPending.service.ts`'s `listPending`), but a caller-*supplied* value is trusted as-is. Any authenticated SAKHI/SUPERVISOR can pass `?userId=<someone else's id>` and read that user's outstanding sync-item queue (device id, entity ids, retry/error state). Not flagged in the PR body's caveats — worth closing alongside the gatherings/beneficiary-risk gap, e.g. by only honoring `userId` for a SUPERVISOR whose roster includes it, and otherwise ignoring/rejecting a mismatched value for SAKHI.

apps/auth-service/src/registration-targets/registration-target.routes.ts
```diff
@@ -0,0 +1,60 @@
...
+        200: {
+          description: 'Registration target rows for the Sakhi',
+          schema: envelope(z.array(registrationTargetSchema)),
+        },
+        400: errorResponse(400, { message: 'sakhiId: Required' }),
+        401: errorResponse(401),
+        500: errorResponse(500),
+      },
+    },
+    authenticate(signer),
```
Dharanishwaran-ND, line 67: Only `authenticate(signer)` here — no `requireRoles`, and no check that the caller-supplied `sakhiId` query param belongs to the calling SAKHI or the calling SUPERVISOR's roster. Any authenticated user (any role) can query any `sakhiId`'s registration targets. Lower severity than the gatherings/beneficiary-risk gap since this is just numeric target counts, no PII, but it's the same missing-scoping pattern — worth closing alongside those, or at minimum adding `requireRoles('SUPERVISOR','MANAGER','ADMIN')` if SAKHI callers were never meant to hit this.

Dharanishwaran-ND:
## Fixed: all 4 ownership-check gaps from the review above

Pushed 4 commits addressing each flagged endpoint:

- **`beneficiaryRisk.controller.ts` / `beneficiaryRiskReferral.service.ts`** (risk-referral-service): `GET /beneficiaries/:id/risk` and `/risk-referrals(+/details)` now resolve the beneficiary's assigned Sakhi via beneficiary-service and scope SAKHI to their own beneficiary, SUPERVISOR to their roster (same pattern `referral.service.ts`'s `decide()` already used) — MANAGER/ADMIN stay unscoped.
- **`gatherings.controller.ts`** (supervisor-operations-service): `listGatherings` now applies the same `sakhiClient.findById` + `supervisorId === caller.id` ownership check its sibling handlers (`listGatheringTopics`, `getAttendance`, etc.) already enforce.
- **`syncPending.controller.ts`** (sync-service): `userId` is no longer trusted as-is — a SAKHI may only request her own id, a SUPERVISOR may additionally request a `userId` belonging to a Sakhi on their own roster (new `SakhiClient`, this service had none before). Also added `transformIgnorePatterns` to this service's jest config, needed now that a spec file pulls in `@armman/service-commons` → `jose` (an ESM-only package) for the first time — matches the config risk-referral-service already has.
- **`registration-target.routes.ts`** (auth-service): added `requireRoles` (was `authenticate` only, any role) plus the same `SAKHI: own only / SUPERVISOR: roster / MANAGER,ADMIN: unscoped` check as `GET /sakhis/:id`, resolved in-process via the existing `SakhiRepository` (same service/table, not a cross-service call).

All four services' full test suites pass and `tsc --noEmit` is clean:
- risk-referral-service: 75/75
- supervisor-operations-service: 209/209
- sync-service: 15/15
- auth-service: 269 passed / 18 skipped (pre-existing skips, unrelated)

The other minor items from the review (the `risk-monitoring`/`arogya-sakhi-roster`/unseeded-tables caveats, the missing standalone gateway routing test, file-size nits) are unchanged — those were confirmed as either already correctly scoped or low-priority follow-ups, not blockers.

PR #163, opened 2026-08-17
feat(visit-form-service): Add visitCode-to-formCode mapping endpoint

### LINKED ISSUE #162: Add visitCode to formCode mapping for INC/CCV/HR/PP visit types
## Background
Frontend (Sakhi app) reported that `visit-schedules` generates entries using `VisitCodeType` (10 values: ANC, ANC_HR, ANC_POST_EDD, DELIVERY, PP, NN, INC, INC_HR, CCV, CCV_HR — see `apps/visit-form-service/prisma/schema.prisma`), but nothing translated that into the `formCode` string the forms API (`GET /forms/:formCode/active-version`, `POST /forms/:formCode/submissions`) needs. Confirmed via code search: no mapping existed anywhere before this change.

## What this adds
- `apps/visit-form-service/src/forms/visit-code-form-map.ts` — static, backend-owned `VISIT_CODE_TO_FORM_CODE` map plus a `resolveFormCodeForVisitType()` helper.
- `GET /api/v1/forms/visit-code-form-map` — returns the full map in one call. Requires authentication (`trustGatewayIdentity`) but no specific role, matching the other read routes on this router.
- `INC_VISIT` / `CCV_VISIT` added as new `form_definitions` seed rows (`apps/visit-form-service/prisma/seed.ts`) so the corresponding formCodes resolve to something rather than 404. **Their schema content is currently a placeholder** — copied verbatim from the existing infant-visit form — pending the actual INC/CCV clinical field spec, which does not exist anywhere in this repo (SRS v3.0 documents INC/CCV scheduling rules in detail but not form field content). Tracked as follow-up, not covered by this issue.

## Confirmed mapping
```json
{
  "ANC": "ANC_VISIT",
  "ANC_HR": "ANC_VISIT",
  "ANC_POST_EDD": "ANC_VISIT",
  "DELIVERY": "DELIVERY_VISIT",
  "PP": "POSTPARTUM_VISIT",
  "NN": "NEONATAL_VISIT",
  "INC": "INC_VISIT",
  "INC_HR": "INC_VISIT",
  "CCV": "CCV_VISIT",
  "CCV_HR": "CCV_VISIT"
}
```
HR visit types intentionally route to their base visit's form — confirmed via SRS v3.0 (FR-S-5.2) that an HR visit is a scheduling trigger (a follow-up visit generated when a high-risk condition is detected), not a distinct clinical form; no existing form schema has a hidden HR section/flag (checked all seed-data JSON for conditional fields).

`PP -> POSTPARTUM_VISIT` is inferred from SRS field-content matching, not yet independently confirmed by an SRS owner — flagged as a follow-up.

## Scope
Touches: `visit-form-service` only (`seed.ts`, `form.controller.ts`, `form.routes.ts`, new `visit-code-form-map.ts`).

## Follow-ups (not in this issue)
- Real INC/CCV clinical field spec needed from ARMMAN/SRS owners to replace the current placeholder schema.
- HR design decision (separate form codes vs. `isHighRisk` flag) — owned by whoever holds the High Risk Protocol docs.
- SRS-owner sign-off on PP = POSTPARTUM_VISIT.
- Confirm with frontend whether the bulk-map shape or a per-code `GET /visit-codes/:code/form` lookup is preferred.

Dharanishwaran-ND:
## Summary
- Adds a backend-owned static map (`visit-code-form-map.ts`) resolving each `VisitCodeType` (ANC, ANC_HR, ANC_POST_EDD, DELIVERY, PP, NN, INC, INC_HR, CCV, CCV_HR) to the `formCode` string the forms API expects — none existed before.
- Adds `GET /api/v1/forms/visit-code-form-map` (authenticated, no specific role, matching the router's other read routes).
- Seeds `INC_VISIT` / `CCV_VISIT` form definitions so those codes resolve end-to-end; their schema content is a **placeholder** (copied from the infant-visit form) pending the real INC/CCV clinical field spec from ARMMAN/SRS owners — tracked as a follow-up, not in scope here.

Closes #162

## Test plan
- [x] `npx jest apps/visit-form-service` — 16/16 suites, 341/341 tests passing
- [x] `tsc --noEmit` — clean
- [ ] Frontend to confirm the bulk-map response shape works for their integration (vs. a per-code lookup)

### DIFF
--- a/apps/api-gateway/src/dashboard/dashboard.controller.ts
+++ b/apps/api-gateway/src/dashboard/dashboard.controller.ts
@@ -1,14 +1,34 @@
-import { Router, type RequestHandler } from 'express';
+import { Router } from 'express';
+import pino from 'pino';
 import {
   asyncHandler,
+  authenticate,
   forbidden,
   notFound,
   ok,
   requireRoles,
+  TRUSTED_GEOGRAPHY_UNIT_ID_HEADER,
+  TRUSTED_PROJECT_ID_HEADER,
+  TRUSTED_ROLES_HEADER,
+  TRUSTED_USER_ID_HEADER,
   unauthorized,
   type AuthenticatedUser,
   type TokenSigner,
 } from '@armman/service-commons';
+
+// A standalone structured logger for degrade()'s failure logging — not
+// req.log, since degrade() call sites across this file/padas.controller.ts/
+// pada-visits.controller.ts don't thread req through; matches the shared
+// pino JSON-log convention (buildLoggerOptions) instead of console.error.
+const logger = pino({ level: process.env.LOG_LEVEL ?? 'info' });
+
+/** Alias kept local to this file's call sites — `authenticate()` is the
+ * gateway-side auth middleware here (it never proxies to a downstream
+ * service, so `verifyAndForwardIdentity` doesn't apply); this route just
+ * needs `Pick<TokenSigner, 'verify'>` (see the shared `authenticate` for
+ * why that's now its parameter type) since the gateway holds a
+ * verify-only PublicKeyVerifier, never a full sign+verify TokenSigner. */
+export const authenticateGateway = authenticate;
 
 // Read directly from process.env (not appConfig) — importing appConfig would
 // pull in its full Zod schema (requires JWT_PUBLIC_KEY etc. with no
@@ -51,43 +71,6 @@ interface VisitSummary {
 interface LastSynced {
   lastSyncedAt: string | null;
 }
-
-/**
- * Minimal Bearer-token verification for this gateway-side route only — a
- * local copy of service-commons' `authenticate()`, not a reuse of it: that
- * function's parameter type is the full `TokenSigner` (sign + verify), but
- * the gateway only ever holds a verify-only `PublicKeyVerifier` (it never
- * issues tokens — see main.ts). `authenticate()` never calls `.sign()` at
- * runtime, so this narrows the same logic to the verify-only shape instead
- * of an unsafe cast to satisfy the wider type.
- */
-export function authenticateGateway(signer: Pick<TokenSigner, 'verify'>): RequestHandler {
-  return (req, _res, next) => {
-    const header = req.header('authorization');
-    if (!header?.startsWith('Bearer ')) return next(unauthorized());
-    const token = header.slice('Bearer '.length).trim();
-    if (!token) return next(unauthorized());
-
-    signer
-      .verify(token)
-      .then((payload) => {
-        req.user = {
-          id: String(payload.sub),
-          roles: Array.isArray(payload.roles) ? (payload.roles as string[]) : [],
-          projectId: typeof payload.projectId === 'string' ? payload.projectId : null,
-          geographyUnitId:
-            typeof payload.geographyUnitId === 'string' ? payload.geographyUnitId : null,
-        };
-        next();
-      })
-      .catch(() => next(unauthorized('Invalid or expired token.')));
-  };
-}
-
-const TRUSTED_USER_ID_HEADER = 'x-armman-user-id';
-const TRUSTED_ROLES_HEADER = 'x-armman-roles';
-const TRUSTED_PROJECT_ID_HEADER = 'x-armman-project-id';
-const TRUSTED_GEOGRAPHY_UNIT_ID_HEADER = 'x-armman-geography-unit-id';
 
 /**
  * The same trusted-identity headers `verifyAndForwardIdentity` sets when
@@ -206,7 +189,7 @@ export async function degrade<T>(label: string, promise: Promise<T>): Promise<T 
   try {
     return await promise;
   } catch (err) {
-    console.error(`Sakhi dashboard: ${label} failed — degrading to null.`, err);
+    logger.error({ err, label }, `Sakhi dashboard: ${label} failed — degrading to null.`);
     return null;
   }
 }
@@ -245,7 +228,7 @@ export function createDashboardRouter(signer: Pick<TokenSigner, 'verify'>): Rout
         degrade(
           'referral referral-summary',
           fetchJson<ReferralSummary>(
-            `${RISK_REFERRAL_SERVICE_URL}/api/v1/referrals/referral-summary`,
+            `${RISK_REFERRAL_SERVICE_URL}/api/v1/referrals/referral-summary?sakhiId=${sakhiId}`,
             caller,
             authorizationHeader,
           ),
--- a/apps/api-gateway/src/dashboard/dashboard.controller.ts
+++ b/apps/api-gateway/src/dashboard/dashboard.controller.ts
@@ -1,34 +1,17 @@
 import { Router } from 'express';
-import pino from 'pino';
 import {
   asyncHandler,
-  authenticate,
   forbidden,
   notFound,
   ok,
   requireRoles,
-  TRUSTED_GEOGRAPHY_UNIT_ID_HEADER,
-  TRUSTED_PROJECT_ID_HEADER,
-  TRUSTED_ROLES_HEADER,
-  TRUSTED_USER_ID_HEADER,
   unauthorized,
   type AuthenticatedUser,
   type TokenSigner,
 } from '@armman/service-commons';
+import { authenticateGateway, degrade, fetchJson } from './dashboard-http.helpers';
 
-// A standalone structured logger for degrade()'s failure logging — not
-// req.log, since degrade() call sites across this file/padas.controller.ts/
-// pada-visits.controller.ts don't thread req through; matches the shared
-// pino JSON-log convention (buildLoggerOptions) instead of console.error.
-const logger = pino({ level: process.env.LOG_LEVEL ?? 'info' });
-
-/** Alias kept local to this file's call sites — `authenticate()` is the
- * gateway-side auth middleware here (it never proxies to a downstream
- * service, so `verifyAndForwardIdentity` doesn't apply); this route just
- * needs `Pick<TokenSigner, 'verify'>` (see the shared `authenticate` for
- * why that's now its parameter type) since the gateway holds a
- * verify-only PublicKeyVerifier, never a full sign+verify TokenSigner. */
-export const authenticateGateway = authenticate;
+export { authenticateGateway, degrade, fetchJson, postJson } from './dashboard-http.helpers';
 
 // Read directly from process.env (not appConfig) — importing appConfig would
 // pull in its full Zod schema (requires JWT_PUBLIC_KEY etc. with no
@@ -73,74 +56,6 @@ interface LastSynced {
 }
 
 /**
- * The same trusted-identity headers `verifyAndForwardIdentity` sets when
- * proxying to a downstream service — duplicated here (not imported from
- * service-commons) because this route calls beneficiary-service/
- * visit-form-service/risk-referral-service/sync-service directly via
- * fetch(), not through the gateway's own proxy layer, so those headers are
- * never set automatically. Every one of those services gates its routes
- * with `trustGatewayIdentity`, which reads these headers — but several of
- * their own controllers (e.g. beneficiary-service's) ALSO independently
- * require the original `Authorization: Bearer` header to still be present
- * (used for their own onward calls to auth-service, e.g. resolving a
- * Supervisor's roster) even though `trustGatewayIdentity` never reads it.
- * `verifyAndForwardIdentity`'s real proxy path forwards both the trusted
- * headers AND the original Authorization header unchanged (it only adds
- * headers, never strips), so both must be sent here too — trusted headers
- * alone 401 at the controller layer despite passing trustGatewayIdentity/
- * requireRoles.
- */
-function trustedIdentityHeaders(
-  caller: AuthenticatedUser,
-  authorizationHeader: string,
-): Record<string, string> {
-  return {
-    Authorization: authorizationHeader,
-    [TRUSTED_USER_ID_HEADER]: caller.id,
-    [TRUSTED_ROLES_HEADER]: caller.roles.join(','),
-    [TRUSTED_PROJECT_ID_HEADER]: caller.projectId ?? '',
-    [TRUSTED_GEOGRAPHY_UNIT_ID_HEADER]: caller.geographyUnitId ?? '',
-  };
-}
-
-export async function fetchJson<T>(
-  url: string,
-  caller: AuthenticatedUser,
-  authorizationHeader: string,
-): Promise<T> {
-  const res = await fetch(url, { headers: trustedIdentityHeaders(caller, authorizationHeader) });
-  if (!res.ok) {
-    throw new Error(`${url} responded ${res.status}`);
-  }
-  const body = (await res.json()) as { data: T };
-  return body.data;
-}
-
-/** Same as fetchJson but POSTs a JSON body — for the count-by-beneficiary
- * endpoints, which take a (potentially large) beneficiaryIds list that
- * doesn't fit safely in a query string. */
-export async function postJson<T>(
-  url: string,
-  body: unknown,
-  caller: AuthenticatedUser,
-  authorizationHeader: string,
-): Promise<T> {
-  const res = await fetch(url, {
-    method: 'POST',
-    headers: {
-      ...trustedIdentityHeaders(caller, authorizationHeader),
-      'Content-Type': 'application/json',
-    },
-    body: JSON.stringify(body),
-  });
-  if (!res.ok) {
-    throw new Error(`${url} responded ${res.status}`);
-  }
-  const responseBody = (await res.json()) as { data: T };
-  return responseBody.data;
-}
-
-/**
  * Resolves the Sakhi's own record (identity anchor of the whole dashboard)
  * and enforces the same self/roster/unrestricted scoping as auth-service's
  * own GET /sakhis/:sakhiId — SAKHI may only view their own dashboard,
@@ -177,21 +92,6 @@ export async function resolveSakhiAndAuthorize(
     throw forbidden("This Sakhi is outside this Supervisor's roster.");
   }
   return sakhi;
-}
-
-/**
- * Awaits a downstream summary call and returns its value, or `null` on any
- * failure — graceful degradation per the agreed dashboard contract: one
- * failing summary service must not take down the whole dashboard. Logs the
- * failure so a silently-null section is still visible in server logs.
- */
-export async function degrade<T>(label: string, promise: Promise<T>): Promise<T | null> {
-  try {
-    return await promise;
-  } catch (err) {
-    logger.error({ err, label }, `Sakhi dashboard: ${label} failed — degrading to null.`);
-    return null;
-  }
 }
 
 /**
--- a/apps/api-gateway/src/dashboard/dashboard.controller.ts
+++ b/apps/api-gateway/src/dashboard/dashboard.controller.ts
@@ -1,0 +1,310 @@
+import { Router, type RequestHandler } from 'express';
+import {
+  asyncHandler,
+  forbidden,
+  notFound,
+  ok,
+  requireRoles,
+  unauthorized,
+  type AuthenticatedUser,
+  type TokenSigner,
+} from '@armman/service-commons';
+
+// Read directly from process.env (not appConfig) — importing appConfig would
+// pull in its full Zod schema (requires JWT_PUBLIC_KEY etc. with no
+// defaults), which process.exit(1)s at module-load time in any test that
+// never otherwise loads config. Matches beneficiary-service's
+// geography.client.ts/sakhi.client.ts convention.
+const AUTH_SERVICE_URL = process.env.AUTH_SERVICE_URL ?? 'http://localhost:3002';
+const BENEFICIARY_SERVICE_URL = process.env.BENEFICIARY_SERVICE_URL ?? 'http://localhost:3001';
+const VISIT_FORM_SERVICE_URL = process.env.VISIT_FORM_SERVICE_URL ?? 'http://localhost:3003';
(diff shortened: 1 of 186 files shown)


apps/auth-service/src/sakhis/sakhi.service.ts
```diff
@@ -77,7 +77,27 @@ export class SakhiService {
...
+   * dashboard (api-gateway's GET /sakhi/:sakhiId/dashboard), which needs a
+   * Sakhi's own displayName and the JWT carries no such field. Returns
+   * immediately once the self-check passes — the SUPERVISOR project-scope
+   * check below must NOT also run for a SAKHI caller: every Sakhi's own JWT
+   * carries a projectId too, so without this early return a Sakhi whose own
+   * `sakhi_profiles.primaryProjectId` doesn't happen to equal their JWT's
+   * projectId claim would be wrongly 403'd fetching their own record.
+   */
   async getById(id: string, caller: CallerScope) {
+    if (caller.roles.includes('SAKHI')) {
```
logeshn-1: **Bug: dual-role caller gets wrongly scoped down to self-only.**

This file defines `isPrivileged(caller)` (MANAGER/ADMIN unrestricted) specifically because "a caller can hold multiple role assignments at once ... must not be scoped down just because one of their roles is restrictive" (see the doc comment above `isPrivileged`, line ~40). But `getById` checks `caller.roles.includes('SAKHI')` *before* ever consulting `isPrivileged`. A JWT with `roles: ['MANAGER', 'SAKHI']` (a combination `auth.service.ts`'s `issueTokens` legitimately allows) hits this branch first and gets hard-restricted to `caller.id === id`, even though the class's own stated invariant says a privileged role should win.

Fix: guard this branch with `if (!isPrivileged(caller) && caller.roles.includes('SAKHI'))`, or check `isPrivileged(caller)` first and fall through.

apps/visit-form-service/src/visits/visitInstance.service.ts
```diff
@@ -18,6 +18,36 @@ function isPrivileged(caller: CallerIdentity): boolean {
...
+ * unscoped. Used by getCountByBeneficiary/getByPada to intersect a
+ * caller-supplied beneficiaryIds list with the caller's own scope (never
+ * trust that list as pre-scoped — see those methods' own doc comments for
+ * the IDOR this closes).
+ */
+async function resolveCallerScoping(
+  caller: CallerIdentity,
+  authorizationHeader: string,
+): Promise<{ sakhiId?: string; sakhiIds?: string[] }> {
+  if (caller.roles.includes('SAKHI')) {
```
logeshn-1, line 36: **Same dual-role scoping bug as `auth-service/sakhis/sakhi.service.ts:93`.**

`resolveCallerScoping` checks `caller.roles.includes('SAKHI')` and returns `{ sakhiId: caller.id }` before ever checking `isPrivileged(caller)` (defined two lines above this function specifically to keep MANAGER/ADMIN unrestricted). A caller whose roles include both `SAKHI` and `MANAGER`/`ADMIN` is silently narrowed to their own single sakhiId instead of getting the unscoped access `isPrivileged` is meant to guarantee. This function backs `getCountByBeneficiary`/`getByPada`, so the effect is wrong (under-reported) visit data for any such dual-role caller, not just a 403.

Same fix as the auth-service instance: check `isPrivileged(caller)` before the SAKHI branch.

apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
```diff
@@ -333,6 +357,96 @@ export class BeneficiaryService {
...
+    authorizationHeader: string,
+  ) {
+    // Security: `ids` is caller-supplied and must never be trusted as-is —
+    // without this, any authenticated caller could pass an arbitrary id
+    // list and get back another Sakhi's beneficiaries' decrypted name/
+    // phone/riskLevel (IDOR). Same self/roster/unscoped scoping as
+    // getIds/getPadaBreakdown; findByIdsWithRisk intersects `ids` with this
+    // scope in its own WHERE clause, so an out-of-scope id is silently
+    // dropped from the result rather than surfaced as a 403.
+    const scoping = await resolveSakhiScoping(undefined, caller, authorizationHeader);
```
logeshn-1, line 451: **Same dual-role scoping bug, third instance (pre-existing helper, newly relied on here).**

`resolveSakhiScoping` (defined above, pre-dating this PR) checks `caller.roles.includes('SAKHI')` first and returns `{ sakhiId: caller.id }` unconditionally, ahead of the SUPERVISOR/MANAGER/ADMIN checks. This new `getByIdsWithRisk` call site is a fresh consumer of that existing bug: a caller holding `SAKHI` plus any privileged role gets scoped to self-only here too. Same pattern as `auth-service/sakhis/sakhi.service.ts:93` and `visit-form-service/visits/visitInstance.service.ts:33` — worth fixing all three the same way (check `isPrivileged`/MANAGER/ADMIN before the SAKHI branch), and worth a shared helper/lint rule so new callers don't keep inheriting it.

apps/sync-service/src/sync/syncBatch.service.ts
```diff
@@ -12,4 +14,35 @@ export class SyncBatchService {
...
+   * ask about their own userId; a SUPERVISOR may ask about a Sakhi in their
+   * own roster; MANAGER/ADMIN are unscoped. Same IDOR guard pattern as
+   * beneficiary-service/risk-referral-service's single-record scoping.
+   */
+  async getLastSyncedAt(
+    userId: string,
+    caller: AuthenticatedUser,
+    authorizationHeader: string,
+  ): Promise<Date | null> {
+    if (caller.roles.includes('SAKHI')) {
```
logeshn-1: **Same dual-role scoping bug, fourth instance** (see `auth-service/sakhis/sakhi.service.ts:93`, `visit-form-service/visits/visitInstance.service.ts:33`, `beneficiary-service/beneficiary/beneficiary.service.ts:230`).

`getLastSyncedAt` checks `caller.roles.includes('SAKHI')` before the SUPERVISOR branch, so a dual-role SAKHI+SUPERVISOR caller querying `GET /sync/last-synced?userId=<other-roster-user>` gets 403'd on a lookup the SUPERVISOR branch immediately below would otherwise permit.

apps/risk-referral-service/src/referrals/referral.service.ts
```diff
@@ -20,6 +21,90 @@ export class ReferralService {
     return this.repository.create(dto);
   }
 
+  /**
+   * Referral Summary widget — accompaniedReferralsCount/pendingFollowUpsCount
+   * for the caller's in-scope beneficiaries. Scoping itself happens entirely
+   * in beneficiary-service's GET /beneficiaries/ids (SAKHI -> own,
+   * SUPERVISOR -> roster, MANAGER/ADMIN -> unscoped, an empty array back for
+   * MANAGER/ADMIN meaning "unscoped" is not distinguishable from "no
+   * beneficiaries exist" — see BeneficiaryClient.getIds) — this service only
```
logeshn-1: **Bug: `getSummary` has no `sakhiId` parameter, unlike its sibling dashboard widgets — returns roster-/system-wide counts instead of one Sakhi's.**

Compare `apps/api-gateway/src/dashboard/dashboard.controller.ts`: the `registration-summary` and `visit-summary` calls both append `?sakhiId=${sakhiId}` when building a specific Sakhi's dashboard, but the `referral-summary` call (same file, ~line 248) does not — because `getSummary` here never accepts a `sakhiId` to forward. As written, a SUPERVISOR or MANAGER/ADMIN opening `GET /sakhi/:sakhiId/dashboard` for one specific Sakhi gets `accompaniedReferralsCount`/`pendingFollowUpsCount` for their *entire roster* (SUPERVISOR) or the *whole system* (MANAGER/ADMIN) instead of that one Sakhi — silently wrong numbers, no error, and no test in this diff exercises a non-SAKHI caller viewing another Sakhi's dashboard to catch it.

Fix: add an optional `sakhiId` param to `getSummary`, forward it to `beneficiaryClient.getIds`, and have the gateway call it with `?sakhiId=` like the other two widgets.

apps/visit-form-service/src/visits/visitInstance.repository.ts
```diff
@@ -53,6 +53,151 @@ export class VisitInstanceRepository {
...
+    if (beneficiaryIds.length === 0 || dueOrOverdueStatusLookupValueIds.length === 0) {
+      return [];
+    }
+
+    return this.prisma.visitInstance.findMany({
+      where: {
+        isDeleted: false,
+        beneficiaryId: { in: beneficiaryIds },
+        statusLookupValueId: { in: dueOrOverdueStatusLookupValueIds },
+        schedule: { scheduledDate: date },
```
logeshn-1: **Bug: exact-date match hides overdue MISSED visits from the pada "open" tab.**

`findByPada` filters by `schedule: { scheduledDate: date }` — an exact match — while `date` defaults to *today* at the call site (`api-gateway/padas/pada-visits.controller.ts:103`: `new Date().toISOString().slice(0, 10)`). But per this same repository's own documented semantics on `countByStatus` above ("a MISSED visit has no actualVisitDate but must still be countable within the period it was due in"), a visit due on 2026-08-10 that's since been marked MISSED still has `scheduledDate = 2026-08-10`. Loading the pada screen today with no explicit `date` query param will never show that overdue visit — defeating the "due/overdue" purpose of the open tab, silently.

`countByStatus`/`countByBeneficiary` handle this correctly with a `fromDate`/`toDate` *range*; `findByPada` should filter with `scheduledDate: { lte: date }` (everything due on or before the given date) rather than an exact match, at least for MISSED rows.

apps/api-gateway/src/padas/pada-visits.controller.ts
```diff
@@ -0,0 +1,219 @@
...
+      }
+
+      const beneficiaryIds = pada.beneficiaries.map((b) => b.id);
+      const caseTypeById = new Map(pada.beneficiaries.map((b) => [b.id, b.caseType]));
+
+      // Both tabs' rows are fetched every time regardless of `status` — the
+      // response always reports both openCount/referralFollowUpCount so
+      // the FE can show both tab labels from one call.
+      const [openCards, followupCards] = await Promise.all([
+        postJson<OpenVisitCard[]>(
```
logeshn-1: **Resilience gap: these two calls aren't degraded, unlike the equivalent calls in the sibling `padas.controller.ts` (list) endpoint.**

`postJson` calls for `openCards`/`followupCards` run inside a bare `Promise.all` with no `degrade()`/try-catch. If visit-form-service or risk-referral-service errors or is briefly unavailable, `postJson` throws a plain `Error` (see below) that propagates uncaught out of the handler — an opaque 500 for the whole pada visit-list screen. The sibling pada-list screen (`padas.controller.ts`) wraps its equivalent counts in `degrade(...)` and would survive the same outage. Worth making the resilience contract consistent between the two screens built in this same PR.

apps/api-gateway/src/padas/pada-visits.controller.ts
```diff
@@ -0,0 +1,219 @@
...
+      const referralFollowUpCount = followupCards.length;
+      const activeBeneficiaryIds =
+        status === 'open'
+          ? openCards.map((c) => c.beneficiaryId)
+          : followupCards.map((c) => c.beneficiaryId);
+
+      const beneficiariesWithRisk = await degrade(
+        'beneficiary by-ids-with-risk',
+        fetchJson<BeneficiaryWithRisk[]>(
+          `${BENEFICIARY_SERVICE_URL}/api/v1/beneficiaries/by-ids-with-risk?ids=${activeBeneficiaryIds.join(',')}${search ? `&search=${encodeURIComponent(search)}` : ''}`,
```
logeshn-1: **Every empty pada guarantees a failed downstream request here.**

When a pada currently has zero due visits and zero pending follow-ups (a common, everyday state — not an edge case), `activeBeneficiaryIds` is `[]`, so this builds `.../by-ids-with-risk?ids=` with an empty `ids` value. `byIdsWithRiskQuerySchema` requires `ids: z.string().trim().min(1)` (`beneficiary-service/beneficiary/dto/by-ids-with-risk-query.dto.ts:15`), so beneficiary-service 400s every single time. It's masked by `degrade()` here so the response is still correct, but every "empty" pada view pays a wasted round trip and logs `console.error('...by-ids-with-risk failed...')` on the happy path. Worth short-circuiting to `[]`/skipping the call when `activeBeneficiaryIds.length === 0`.

apps/api-gateway/src/dashboard/dashboard.controller.ts
```diff
@@ -0,0 +1,310 @@
...
+/**
+ * Minimal Bearer-token verification for this gateway-side route only — a
+ * local copy of service-commons' `authenticate()`, not a reuse of it: that
+ * function's parameter type is the full `TokenSigner` (sign + verify), but
+ * the gateway only ever holds a verify-only `PublicKeyVerifier` (it never
+ * issues tokens — see main.ts). `authenticate()` never calls `.sign()` at
+ * runtime, so this narrows the same logic to the verify-only shape instead
+ * of an unsafe cast to satisfy the wider type.
+ */
+export function authenticateGateway(signer: Pick<TokenSigner, 'verify'>): RequestHandler {
```
logeshn-1: **Reuse: this duplicates `service-commons`' `authenticate()` almost verbatim** (acknowledged in the comment above as "a local copy... not a reuse"). The stated reason — `authenticate()` requires a full `TokenSigner` but the gateway only has verify — doesn't hold: `libs/service-commons/src/auth/forward-trusted-identity.ts`'s `verifyAndForwardIdentity` already takes the identical `Pick<TokenSigner, 'verify'>` type and does the same verify-then-populate-identity work. This copy also drops the `AuthMarker`/`__requiresAuth` flag `authenticate()` attaches for the OpenAPI doc router's auto-detection, so these new dashboard/pada routes won't be flagged as requiring auth in generated docs.

Also note line 87 below redeclares `TRUSTED_USER_ID_HEADER` etc. as local string literals instead of importing the constants already exported from `forward-trusted-identity.ts` — same-value duplication with no compiler link if the header names ever change.

apps/api-gateway/src/dashboard/dashboard.controller.ts
```diff
@@ -0,0 +1,310 @@
...
+ * Awaits a downstream summary call and returns its value, or `null` on any
+ * failure — graceful degradation per the agreed dashboard contract: one
+ * failing summary service must not take down the whole dashboard. Logs the
+ * failure so a silently-null section is still visible in server logs.
+ */
+export async function degrade<T>(label: string, promise: Promise<T>): Promise<T | null> {
+  try {
+    return await promise;
+  } catch (err) {
+    console.error(`Sakhi dashboard: ${label} failed — degrading to null.`, err);
```
logeshn-1: Two smaller cleanups on this file:

1. `console.error(...)` here bypasses the shared pino logger used everywhere else in the codebase for structured logging.
2. This is a brand-new 310-line file; repo convention (`.claude/CLAUDE.md` §3) asks for files ≤~250 lines, split by responsibility when larger — worth splitting the auth/header helpers out from the route handlers.

apps/sync-service/src/sync/sakhi.client.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Resolves the Sakhi ids reporting to a given Supervisor, via auth-service's
+ * existing `GET /projects/:projectId/sakhis`, called through the gateway.
+ * Used to scope `GET /sync/last-synced` to a Supervisor's own roster.
+ * Duplicated from risk-referral-service's identical helper — this service
+ * has no shared client library to import it from (forklift rule: no
+ * cross-service imports).
+ */
+export async function listSakhiIdsForSupervisor(
```
logeshn-1, line 22: **Reuse: third verbatim copy of `listSakhiIdsForSupervisor`.** This file's own comment admits it's "duplicated from risk-referral-service's identical helper," and `beneficiary-service/sakhi/sakhi.client.ts` has a third copy of the same fetch-roster-and-filter-by-supervisorId logic. A future fix (e.g. the pagination caveat already noted in the beneficiary-service copy) now has to be applied in three places by hand. Consider hoisting the shared bits (fetch + status-check + filter shape) into `service-commons` even though the `forklift` rule blocks sharing service-specific clients — the *pattern*, not the service call, is what's duplicated.

logeshn-1 (commented):
Automated strict review. All findings below were independently verified by reading the PR-head source (not just diff inference) before posting. Ranked by severity: two correctness/authorization bugs first, then a resilience gap, then reuse/size cleanups.

apps/auth-service/src/sakhis/sakhi.service.ts
Dharanishwaran-ND: Already fixed on this branch: `getById` (auth-service/sakhis/sakhi.service.ts:93) checks `!isPrivileged(caller) && caller.roles.includes('SAKHI')` before the self-only restriction, so a dual-role MANAGER/ADMIN+SAKHI caller correctly falls through to the unrestricted path.

apps/visit-form-service/src/visits/visitInstance.service.ts
Dharanishwaran-ND: Already fixed on this branch: `resolveCallerScoping` (visit-form-service/visits/visitInstance.service.ts:33) checks `isPrivileged(caller)` and returns `{}` (unscoped) before the SAKHI branch.

apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
Dharanishwaran-ND: Already fixed on this branch: `resolveSakhiScoping` (beneficiary-service/beneficiary/beneficiary.service.ts:242) checks `isPrivileged(caller)` first, before any SAKHI-specific narrowing.

apps/sync-service/src/sync/syncBatch.service.ts
Dharanishwaran-ND: Already fixed on this branch: `getLastSyncedAt` (sync-service/sync/syncBatch.service.ts:34) wraps both the SAKHI and SUPERVISOR branches inside `if (!isPrivileged(caller))`, so a dual-role SAKHI+SUPERVISOR/MANAGER/ADMIN caller is unrestricted.

apps/risk-referral-service/src/referrals/referral.service.ts
Dharanishwaran-ND: Already fixed on this branch: `getSummary` (risk-referral-service/referrals/referral.service.ts:40) takes an optional `sakhiId` and forwards it to `beneficiaryClient.getIds`, and the gateway's `referral-summary` call (dashboard.controller.ts) already appends `?sakhiId=${sakhiId}`, matching the sibling registration-summary/visit-summary widgets.

apps/visit-form-service/src/visits/visitInstance.repository.ts
Dharanishwaran-ND: Already fixed on this branch: `findByPada` (visit-form-service/visits/visitInstance.repository.ts:227,235) filters PENDING rows with exact `scheduledDate: date` but MISSED rows with `scheduledDate: { lte: date }`, so an overdue MISSED visit from an earlier date still surfaces in today's open tab.

apps/api-gateway/src/padas/pada-visits.controller.ts
Dharanishwaran-ND: Already fixed on this branch: both the `openCards`/`followupCards` calls in pada-visits.controller.ts are wrapped in `degrade(...)`, matching the sibling padas.controller.ts's resilience contract.

apps/api-gateway/src/padas/pada-visits.controller.ts
Dharanishwaran-ND: Already fixed on this branch: pada-visits.controller.ts short-circuits to `[]` when `activeBeneficiaryIds.length === 0`, skipping the by-ids-with-risk call entirely instead of firing a guaranteed-400 request.

apps/api-gateway/src/dashboard/dashboard.controller.ts
Dharanishwaran-ND: Reuse part already fixed: this file now imports `authenticate` from `@armman/service-commons` directly (no local duplicate), and `TRUSTED_USER_ID_HEADER`/`TRUSTED_ROLES_HEADER`/`TRUSTED_PROJECT_ID_HEADER`/`TRUSTED_GEOGRAPHY_UNIT_ID_HEADER` are imported from service-commons rather than redeclared as local literals.

apps/api-gateway/src/dashboard/dashboard.controller.ts
Dharanishwaran-ND: Both fixed in 6ec7c35: (1) this file already used the shared pino logger, not console.error. (2) File was still 293 lines — split the generic auth/HTTP helpers (authenticateGateway, trustedIdentityHeaders, fetchJson, postJson, degrade) out into dashboard-http.helpers.ts, re-exported from dashboard.controller.ts so pada-visits.controller.ts/padas.controller.ts's existing imports are unaffected. dashboard.controller.ts is now 194 lines.

apps/sync-service/src/sync/sakhi.client.ts
Dharanishwaran-ND: Agreed this is duplicated a third time, but leaving as-is: the forklift rule (.claude/CLAUDE.md §1) explicitly blocks sharing service-specific clients across services — sync-service, risk-referral-service, and beneficiary-service must each stay independently extractable with no cross-service code imports. Hoisting this into service-commons would violate that rule for a client that calls another service's API, not a cross-cutting Express concern. Tracking the pagination caveat as a known limitation to fix in all three copies together if it's ever addressed.

PR #165, opened 2026-08-18
fix: media checksum, event completion photo, and inventory detail alias

### LINKED ISSUE #166: Beneficiary Data Download: media checksum, event completion photo, and inventory detail alias
## Context

Found and fixed while manually walking the Beneficiary Data Download upload/completion flow end-to-end (presigned upload -> finalize -> attach photo to a Meeting/Training event -> complete). Two real bugs blocked that flow from ever working, plus one small gap.

## Bugs fixed

1. **`checksum` unusable in `POST`/`GET /media` responses** (media-service) — stored correctly as `Bytes` in Postgres, but returned to clients as a raw Buffer (`{"type":"Buffer","data":[...]}`) instead of a string. Fixed in the controller's response mapping only; storage/schema unchanged.

2. **Event completion could never succeed** (supervisor-operations-service) — `POST /supervisor-events/:id/photos` wrote to the `event_photos` gallery table but never set the event's own `photoMediaId`, while `PATCH /supervisor-events/:id/complete` checks exactly that column before allowing completion. The two endpoints disagreed on where "the completion photo" lives, so completion was unreachable even when the client did everything right. Fixed: the first photo added to a SCHEDULED event now also sets `photoMediaId` (same transaction as the gallery write); later photos only add to the gallery and never overwrite it.

## Also included

- `GET /inventory-transactions/:id/details` — added as a second alias for `GET /inventory-transactions/:id`, for consumers expecting the detail row nested under the `/inventory-transactions` collection path.

## Verification

- Full lint + Jest suites green on every touched service.
- Verified live end-to-end against a running dev environment: presigned upload -> S3 PUT -> finalize (checksum returned as expected hex string) -> attach photo to event -> event's `photoMediaId` populated -> attendance recorded -> complete succeeds (status `COMPLETED`).

## PR

#165

logeshn-1:
## Summary
- Fixed `checksum` in media-service responses — was returning a raw Buffer object, now returns a plain hex string.
- Fixed event completion: the first photo added to a Meeting/Training event now properly attaches as its completion photo, so `PATCH /supervisor-events/:id/complete` can actually succeed once a photo + attendance exist.
- Added `GET /inventory-transactions/:id/details` as an alias endpoint.
- Fixed a CI-only test failure: `mediaAsset.controller.spec.ts` transitively pulled in env-config validation that only has defaults in local `.env`, not in CI.

## Test plan
- Lint + full Jest suite green on media-service and supervisor-operations-service.
- Verified live end-to-end: upload → finalize → checksum returned correctly; event → add photo → attendance → complete → status COMPLETED.
- Verified the CI fix locally by explicitly unsetting the relevant env vars before re-running tests.

Closes #166

### DIFF
--- a/apps/supervisor-operations-service/src/operations/inventory.routes.ts
+++ b/apps/supervisor-operations-service/src/operations/inventory.routes.ts
@@ -233,6 +233,30 @@ export function registerInventoryRoutes(doc: DocumentedRouter, service: Operatio
     controller.getTransactionById,
   );
 
+  // Second dedicated-path alias for GET /inventory-transactions/:id — same
+  // handler, same response — for consumers that expect the detail row
+  // nested under the /inventory-transactions collection path instead of
+  // /item-transactions.
+  doc.get(
+    '/inventory-transactions/:id/details',
+    {
+      summary:
+        'Fetch a single inventory transaction by id (alias for GET /inventory-transactions/:id)',
+      tags: ['Supervisor Operations'],
+      params: transactionIdParamsSchema,
+      responses: {
+        200: { description: 'Inventory transaction', schema: envelope(inventoryTransactionSchema) },
+        401: { description: 'Unauthenticated', schema: apiErrorSchema },
+        403: { description: 'Caller does not own this transaction', schema: apiErrorSchema },
+        404: { description: 'Transaction not found', schema: apiErrorSchema },
+      },
+    },
+    trustGatewayIdentity,
+    requireRoles('SUPERVISOR', 'MANAGER', 'ADMIN'),
+    validate(transactionIdParamsSchema, 'params'),
+    controller.getTransactionById,
+  );
+
   doc.put(
     '/inventory-transactions/:id',
     {
--- a/apps/supervisor-operations-service/src/operations/operations.repository.ts
+++ b/apps/supervisor-operations-service/src/operations/operations.repository.ts
@@ -327,10 +327,30 @@ export class OperationsRepository {
     });
   }
 
-  createEventPhoto(eventId: string, mediaId: string, createdByUserId: string) {
-    return this.prisma.eventPhoto.create({
-      data: { eventId, mediaId, createdByUserId },
-    });
+  /**
+   * Creates the gallery row and, when the event has no completion photo yet,
+   * also sets `photoMediaId` on the event in the same transaction — so the
+   * gallery and the event's completion-eligibility flag can never disagree
+   * from a partial write (see addEventPhoto's doc comment).
+   */
+  async createEventPhoto(
+    eventId: string,
+    mediaId: string,
+    createdByUserId: string,
+    setAsCompletionPhoto: boolean,
+  ) {
+    const [photo] = await this.prisma.$transaction([
+      this.prisma.eventPhoto.create({ data: { eventId, mediaId, createdByUserId } }),
+      ...(setAsCompletionPhoto
+        ? [
+            this.prisma.supervisorEvent.update({
+              where: { id: eventId },
+              data: { photoMediaId: mediaId, updatedByUserId: createdByUserId },
+            }),
+          ]
+        : []),
+    ]);
+    return photo;
   }
 
   /**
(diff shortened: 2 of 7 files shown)


apps/supervisor-operations-service/src/operations/operations.repository.ts
```diff
@@ -327,10 +327,30 @@ export class OperationsRepository {
...
+   * gallery and the event's completion-eligibility flag can never disagree
+   * from a partial write (see addEventPhoto's doc comment).
+   */
+  async createEventPhoto(
+    eventId: string,
+    mediaId: string,
+    createdByUserId: string,
+    setAsCompletionPhoto: boolean,
+  ) {
+    const [photo] = await this.prisma.$transaction([
```
Dharanishwaran-ND, line 345: **Race condition: two concurrent `addEventPhoto` calls can both set `photoMediaId`, with the second silently overwriting the first — violating the doc comment's own invariant.**

`setAsCompletionPhoto` (`!event.photoMediaId`) is computed from a read in `operations.service.ts` *before* this transaction starts. If two requests both add a photo to the same event that currently has `photoMediaId: null`, both can read `null`, both compute `setAsCompletionPhoto = true`, and both transactions run — the second `supervisorEvent.update` overwrites the first's `photoMediaId`/`updatedByUserId` (last-write-wins), even though the first-writer's photo should have "won" per the stated invariant ("a later photo never overwrites an already-set `photoMediaId`").

Fix: make the write itself conditional instead of trusting the pre-read boolean, e.g. an `updateMany({ where: { id: eventId, photoMediaId: null }, data: {...} })` inside the transaction, so only the first writer to reach the DB actually sets it.

apps/supervisor-operations-service/src/operations/inventory.routes.ts
```diff
@@ -233,6 +233,30 @@ export function registerInventoryRoutes(doc: DocumentedRouter, service: Operatio
     controller.getTransactionById,
   );
 
+  // Second dedicated-path alias for GET /inventory-transactions/:id — same
+  // handler, same response — for consumers that expect the detail row
+  // nested under the /inventory-transactions collection path instead of
+  // /item-transactions.
+  doc.get(
```
Dharanishwaran-ND, line 240: **Cleanup: this is now the second near-identical alias for `GET /inventory-transactions/:id`.** `/item-transactions/:id/details` (added in a prior PR, just above this hunk) already aliases the same handler/response/roles for the same use case (Beneficiary Data Download's transaction detail row). Adding `/inventory-transactions/:id/details` on top means three routes now resolve to the exact same `controller.getTransactionById` — worth confirming with the actual consumer which single path it needs rather than growing an alias per caller guess; each additional alias is more OpenAPI surface and another route to keep in sync if the handler ever needs to diverge per-path.

apps/supervisor-operations-service/src/operations/operations.repository.ts
logeshn-1: Confirmed, good catch — the invariant was only true single-threaded. Fixed in fca1f86: the completion-photo check is now the `updateMany`'s own `WHERE photoMediaId IS NULL`, evaluated atomically inside the same transaction as the gallery write, instead of a boolean decided by a pre-read. Row-level locking on that UPDATE serializes concurrent attempts, so only the first to actually commit ever changes a row. Verified live against 5 truly concurrent requests on a photo-less event — ended with exactly one clean `photoMediaId`, no overwrite. Also added `operations.repository.spec.ts` (didn't exist before) covering the conditional-update shape directly, since a service-level mock can't verify this.

apps/supervisor-operations-service/src/operations/inventory.routes.ts
logeshn-1: Fair concern about alias sprawl — but this one is intentional, not a guess: the original task spec for the Beneficiary Data Download flow explicitly called out `GET /inventory-transactions/{id}/details` as its own required endpoint, distinct from `/item-transactions/:id/details` (different consumer path expectation, same underlying data). It's not me speculating about what a caller might want — it's what was actually asked for. Given that, I'd rather keep it as a thin alias (same handler, same roles, same response, ~zero maintenance cost) than remove working code the consumer specifically requested. Happy to revisit if the consumer confirms they only need one path going forward.

PR #168, opened 2026-08-18
Beneficiary lifecycle write-path: phase advance, closure, and reopen idempotency

### LINKED ISSUE #167: Beneficiary lifecycle write-path: phase advance, closure, and reopen idempotency
## Summary

Three related gaps in the beneficiary lifecycle write-path (delivery, closure, reopen) that left `currentStatus`/`currentPhase` stale or duplicable:

1. **CR-041 — currentPhase never advances on delivery.** `beneficiary.service.ts` sets `currentPhase` once at case creation and never updates it. A DELIVERY_VISIT submission needs to advance the mother's phase ANC→PP and each auto-created child's phase to NN.

2. **Closure write-path missing.** `POST /closures` (closure-reopen-service) recorded a closure decision but never called beneficiary-service to flip `currentStatus` to `CLOSED` — a closed beneficiary never showed as closed via `GET /beneficiaries`. Also: submitting the `ANC_CLOSURE_VISIT`/`CHILD_CLOSURE_VISIT` form saved the form answers but never called `POST /closures` at all, so the whole closure flow was disconnected end-to-end.

3. **Reopen-request idempotency gap.** `POST /reopen-requests` had no client-generated dedup key (unlike `POST /closures`' `localClosureUuid`), so an offline-first mobile retry could create a duplicate `ReopenRequest` row.

## Scope

- New `PATCH /beneficiaries/:id/phase` (server-to-server, CR-041 phase advance)
- New `PATCH /beneficiaries/:id/close` (server-to-server, idempotent)
- `closure-reopen-service`'s `ClosureService.create()`/`decide()` now call the new close endpoint
- `visit-form-service` now auto-calls `POST /closures` on ANC_CLOSURE_VISIT/CHILD_CLOSURE_VISIT submission (`continue_with_closure: 'yes'`), mapping the form's `closure_reason` to the CLOSURE_REASON lookup and `closureType`
- New `GET /reopen-requests?beneficiaryId=...` for the app's "reopen pending" state
- `localReopenRequestUuid` added to `POST /reopen-requests` for offline-retry safety

## Verification

Verified live end-to-end against the shared dev environment: delivery→phase advance, closure form submission→beneficiary CLOSED, and reopen-request idempotent retry all confirmed working. Full details in `docs/test-runs/closure-reopen-be-answers.md` and `docs/test-runs/reopen-idempotency-fix.md` (gitignored, not in this PR).

Dharanishwaran-ND:
## Summary
- Closes #167
- Adds `PATCH /beneficiaries/:id/phase` — advances a mother's `currentPhase` ANC→PP and an auto-created child's to NN on a `DELIVERY_VISIT` submission (CR-041)
- Adds `PATCH /beneficiaries/:id/close` (idempotent) and wires `closure-reopen-service`'s `POST /closures` / decision-approval flow to call it — a closure decision now actually flips the beneficiary's `currentStatus`
- Wires `ANC_CLOSURE_VISIT`/`CHILD_CLOSURE_VISIT` form submissions to auto-call `POST /closures`, closing the gap where submitting the closure form had no effect on the beneficiary record at all
- Adds `localClosureUuid` / `localReopenRequestUuid` client-generated idempotency keys to `POST /closures` / `POST /reopen-requests`, matching beneficiary enrollment's `localCaseUuid` convention, so an offline-first mobile retry can't create duplicate rows
- Adds `GET /reopen-requests?beneficiaryId=...` so the app can show "reopen pending review" instead of just "closed" while a request is mid-flow
- Reconciles the SRS's PP3/PP5 schedule table (Day+105 vs Day+113 discrepancy) — Day+118/window 113–123 confirmed authoritative, matching the already-implemented GoRules scheduling pack

## Test plan
- [x] `npx nx test beneficiary-service` — 258 passing
- [x] `npx nx test closure-reopen-service` — 78 passing
- [x] `npx nx test visit-form-service` — 413 passing
- [x] `npx nx lint` on all three — clean (only pre-existing unrelated warnings)
- [x] Verified live end-to-end on the shared dev environment: delivery→phase advance, closure form submission→beneficiary `CLOSED`, and reopen-request idempotent retry all confirmed working
- [x] Prisma migrations applied to dev DB (`add_closure_local_closure_uuid`, `add_reopen_request_local_uuid`)

### DIFF
--- a/apps/beneficiary-service/src/beneficiary/beneficiary.routes.ts
+++ b/apps/beneficiary-service/src/beneficiary/beneficiary.routes.ts
@@ -18,6 +18,8 @@ import { byIdsWithRiskQuerySchema } from './dto/by-ids-with-risk-query.dto';
 import { upsertSocioDemographicsSchema } from './dto/upsert-socio-demographics.dto';
 import { upsertRiskConditionSummarySchema } from './dto/upsert-risk-condition-summary.dto';
 import { applyLmpChangeSchema } from './dto/apply-lmp-change.dto';
+import { updatePhaseSchema } from './dto/update-phase.dto';
+import { applyClosureSchema } from './dto/apply-closure.dto';
 import {
   errorResponse,
   requireRoles,
@@ -554,6 +556,69 @@ export function registerBeneficiaryRoutes(doc: DocumentedRouter, service: Benefi
     validate(idParamsSchema, 'params'),
     validateBody(applyLmpChangeSchema),
     controller.applyLmpChange,
+  );
+
+  doc.patch(
+    '/beneficiaries/:id/phase',
+    {
+      summary:
+        'Advance currentPhase after a DELIVERY_VISIT submission (CR-041) — ' +
+        "gated by requireRoles('SAKHI') since this codebase has no machine/service-account " +
+        "identity: the call chain originates from a SAKHI's DELIVERY_VISIT form submission " +
+        "(visit-form-service -> here), forwarding the SAKHI's own token. Only ANC->PP (mother) " +
+        'and *->NN (child) are accepted; any other transition 409s.',
+      tags: ['Beneficiaries'],
+      params: idParamsSchema,
+      responses: {
+        200: {
+          description: 'currentPhase updated; the updated case is returned',
+          schema: envelope(beneficiaryCaseDetailSchema),
+        },
+        400: errorResponse(400, { message: 'phase: Invalid enum value' }),
+        401: errorResponse(401),
+        403: errorResponse(403, { message: 'This beneficiary case is outside your own roster.' }),
+        404: errorResponse(404, { message: 'Beneficiary case not found.' }),
+        409: errorResponse(409, { message: 'Cannot move a MOTHER case from PP to ANC.' }),
+        500: errorResponse(500),
+      },
+    },
+    trustGatewayIdentity,
+    requireRoles('SAKHI'),
+    validate(idParamsSchema, 'params'),
+    validateBody(updatePhaseSchema),
+    controller.applyPhaseChange,
+  );
+
+  doc.patch(
+    '/beneficiaries/:id/close',
+    {
+      summary:
+        'Close a beneficiary case after a closure submission (ANC_CLOSURE_VISIT / ' +
+        'CHILD_CLOSURE_VISIT) — server-to-server only, called by closure-reopen-service. ' +
+        "Forwards the submitting SAKHI's own token for a non-reviewed closure, or the " +
+        "deciding Supervisor's token once a MIGRATION closure is approved — this codebase " +
+        'has no machine/service-account identity. Idempotent: closing an already-CLOSED case ' +
+        'is a no-op success, not a 409, since the mobile app may retry this call offline.',
+      tags: ['Beneficiaries'],
+      params: idParamsSchema,
+      responses: {
+        200: {
+          description: 'Case closed (or already CLOSED); the current case is returned',
+          schema: envelope(beneficiaryCaseDetailSchema),
+        },
+        400: errorResponse(400, { message: 'reasonCode: Required' }),
+        401: errorResponse(401),
+        403: errorResponse(403, { message: 'This beneficiary case is outside your own roster.' }),
+        404: errorResponse(404, { message: 'Beneficiary case not found.' }),
+        409: errorResponse(409, { message: 'Unable to close this beneficiary case.' }),
+        500: errorResponse(500),
+      },
+    },
+    trustGatewayIdentity,
+    requireRoles('SAKHI', 'SUPERVISOR', 'MANAGER', 'ADMIN'),
+    validate(idParamsSchema, 'params'),
+    validateBody(applyClosureSchema),
+    controller.applyClosure,
   );
 
   doc.patch(
--- a/apps/beneficiary-service/src/beneficiary/beneficiary.routes.ts
+++ b/apps/beneficiary-service/src/beneficiary/beneficiary.routes.ts
@@ -594,11 +594,16 @@ export function registerBeneficiaryRoutes(doc: DocumentedRouter, service: Benefi
     {
       summary:
         'Close a beneficiary case after a closure submission (ANC_CLOSURE_VISIT / ' +
-        'CHILD_CLOSURE_VISIT) — server-to-server only, called by closure-reopen-service. ' +
-        "Forwards the submitting SAKHI's own token for a non-reviewed closure, or the " +
-        "deciding Supervisor's token once a MIGRATION closure is approved — this codebase " +
-        'has no machine/service-account identity. Idempotent: closing an already-CLOSED case ' +
-        'is a no-op success, not a 409, since the mobile app may retry this call offline.',
+        'CHILD_CLOSURE_VISIT) — intended to be called server-to-server by ' +
+        "closure-reopen-service, forwarding the submitting SAKHI's own token for a " +
+        "non-reviewed closure, or the deciding Supervisor's token once a MIGRATION closure " +
+        'is approved. KNOWN GAP: this codebase has no machine/service-account identity, so ' +
+        'this is also reachable directly by a SAKHI who owns the case, bypassing the ' +
+        'closures audit row and any required supervisor review (see BeneficiaryService.' +
+        'applyClosure doc comment) — tracked as a follow-up, not enforced here. Idempotent: ' +
+        'closing an already-CLOSED case with the same reasonCode is a no-op success, not a ' +
+        '409, since the mobile app may retry this call offline; a genuinely different ' +
+        'reasonCode 409s instead of silently overwriting the audit trail.',
       tags: ['Beneficiaries'],
       params: idParamsSchema,
       responses: {
--- a/apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
+++ b/apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
@@ -16,7 +16,7 @@ import type {
   BeneficiaryStatusHistory,
   Prisma,
 } from '../../../../node_modules/.prisma/client-beneficiary-service';
-import type { BeneficiaryStatus, CaseType } from './beneficiary.constants';
+import type { BeneficiaryStatus, CasePhase, CaseType } from './beneficiary.constants';
 import { buildSearchTokens, evaluateDuplicateMatch } from './beneficiary.duplicate-detection';
 import { computeBmi, withDecryptedName } from './beneficiary.mapper';
 import type { BeneficiaryListFilters, BeneficiaryRepository } from './beneficiary.repository';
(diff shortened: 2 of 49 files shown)


logeshn-1 (commented):
Reviewed by checking out this branch (commit `3069cc0`) into an isolated worktree and reading the actual code (not just the diff) across all three touched services. Verified every finding below against the real files — no speculation. Summary at the bottom.

## Critical

**1. `POST /closures` lets a SAKHI self-approve a closure, bypassing supervisor review entirely.**
`create-closure.dto.ts:30-32` accepts client-supplied `supervisorStatus`/`supervisorId`/`supervisorNotes` with no cross-field validation tying `supervisorStatus` to the closure reason. `closure.service.ts:37` branches purely on `created.supervisorStatus === 'PENDING'` — if not PENDING, it goes straight to `beneficiaryClient.closeCase(...)` (line ~73-84). A SAKHI can POST `{closureReasonLookupValueId: <MIGRATION>, supervisorStatus: "APPROVED", supervisorId: "<any-uuid>"}` and the beneficiary gets closed immediately, with a persisted row falsely showing a supervisor decision that was never made. `create-closure.dto.spec.ts` actually asserts these fields are *accepted* — no test rejects this.

**2. For `closure_reason: migration` specifically, there's no ownership check anywhere in the call chain.**
`ClosureService.create()` never checks that `dto.beneficiaryId` belongs to the calling SAKHI, and the `supervisorStatus: 'PENDING'` branch routes to `approvalClient.create(...)` (closure.service.ts:37-66), which also does no ownership validation. Combined with #1's auto-wiring from `visit-form-service` (any authenticated SAKHI can submit `ANC_CLOSURE_VISIT`/`CHILD_CLOSURE_VISIT` with `closure_reason: 'migration'` for *any* `beneficiaryId` in the system — `form.routes.ts` requires only authentication, no roster check), this means a SAKHI can raise a real supervisor-approval card for a beneficiary they have no relationship to.

**3. `PATCH /beneficiaries/:id/close` is reachable directly by any SAKHI/SUPERVISOR/MANAGER/ADMIN, not just closure-reopen-service.**
`beneficiary.routes.ts:592-618` documents this as "server-to-server only, called by closure-reopen-service" but gates it with `requireRoles('SAKHI', 'SUPERVISOR', 'MANAGER', 'ADMIN')` — the same roles a human client has — and the gateway proxies `/beneficiaries` with no path-specific restriction. `closure-reopen-service`'s own client just forwards whatever `Authorization` it was given, with no distinguishing machine credential. A SAKHI who owns the case can call this directly with `{reasonCode: "MIGRATION"}` and close it immediately — no `closures` row, no Quick Response card, no supervisor step at all. This compounds #1/#2: the intended approval gate is enforced nowhere in the actual trust boundary.

## High

**4. CHILD phase-advance doesn't validate the current phase — allows phase regression.**
`beneficiary.service.ts:746-748`:
```ts
const isValidTransition =
  (existing.caseType === 'MOTHER' && fromPhase === 'ANC' && toPhase === 'PP') ||
  (existing.caseType === 'CHILD' && toPhase === 'NN');
```
The MOTHER branch checks both `fromPhase` and `toPhase`; the CHILD branch only checks `toPhase === 'NN'`, accepting *any* current phase. This contradicts the method's own doc comment ("anything else 409s rather than silently moving a case backward"). Not yet reachable today (a CHILD is created at `NN`, so a same-value call short-circuits at the idempotent early-return before this check), but becomes a real phase-regression bug the moment any path advances a CHILD past `NN` (e.g. to `INC`/`CCV`, both in `CASE_PHASES`) — a delayed offline-retry of a stale `DELIVERY_VISIT` submission is a plausible trigger given this app's offline-first design.

**5. An unguarded downstream call can turn an already-successful submission into a failed response.**
`form.service.ts:288`, `resolveDeliveryChildren`'s `findBeneficiaryById(dto.beneficiaryId, ...)` call has no try/catch, and its caller (`form.service.ts:229-232`) doesn't wrap it either. This runs *after* `repository.createSubmission(...)` has already durably persisted the row. Every other downstream call in this file (`createChildBeneficiary`, `updateBeneficiaryPhase`, `createClosure`, risk/socio-demographic/health-history syncs) tolerates its own failures internally — this one doesn't. A 403 (ownership mismatch), 500, or network blip from beneficiary-service here fails the whole HTTP response with a false failure, even though the visit form submission itself succeeded. `form.service.spec.ts` only mocks `findBeneficiaryById` resolving `null`, never rejecting — this path is untested.

## Medium

**6. Idempotent close silently drops a differing `reasonCode` with no audit trail.**
`beneficiary.service.ts:803-805` short-circuits on `existing.currentStatus === 'CLOSED'` without comparing or recording `reasonCode`. If a case is closed once with reason `MEDICAL`, a later, genuinely different closure attempt with reason `MIGRATION` returns `200` success silently — the doc comment claims `reasonCode` is "recorded on beneficiary_status_history for audit," but that's only true for whichever call won. Existing idempotency test only exercises a repeat with the *same* reasonCode.

**7. No `P2002` handling on `closures`/`reopen_requests` create — concurrent duplicate submissions get a raw 500 instead of the intended idempotent 200.**
Both migrations correctly add `local_closure_uuid`/`local_reopen_request_uuid` as `UNIQUE NOT NULL`, so no duplicate row can land — but `ClosureService.create()`/`ReopenRequestService.create()` do a plain find-then-create with no try/catch around `repository.create()`, and this repo's shared `errorHandler` doesn't special-case `PrismaClientKnownRequestError` (unlike other services that convert P2002 to a clean 409). Two near-simultaneous retries with the same idempotency key (a double-tap, or a retry firing while the original is still in flight) both pass the `findBy...Uuid` check as `null`, and the losing request's raw P2002 propagates to a generic 500 instead of the promised idempotent 200. Not tested — existing tests only cover the *sequential* retry (row already committed).

**8. Inconsistent `beneficiaryId` source on submission replay.**
`form.service.ts:158-180`: on a replay (existing submission found by `localSubmissionUuid`), the delivery/phase-advance branch uses `dto.beneficiaryId` (the retry request's own body, line 163) while the closure branch three lines later deliberately uses `existing.beneficiaryId` (the originally-stored value, line 173) specifically to avoid trusting replay input. A retried request with a swapped `beneficiaryId` would run child-creation/phase-advance against a different beneficiary than the one actually recorded — bounded by beneficiary-service's own ownership check, so not a cross-tenant IDOR, but a real inconsistency between two branches of the same replay block.

## Worth confirming (not necessarily bugs)

**9. Mother/child closure linkage is entirely unaddressed**, with no comment either way. Closing a MOTHER case doesn't cascade to linked CHILD cases (or vice versa) — unlike this codebase's convention of explicitly documenting known/accepted gaps (e.g. `applyLmpChange`'s schedule-regen note), there's no such acknowledgment here. Worth confirming with product/SRS whether this is intentional.

## Summary
Two things need to happen before merge, IMO: **#1/#2/#3 are a single compounding gap** — the entire supervisor-approval workflow this PR is meant to wire up has no actual enforcement at its trust boundaries (client-controlled `supervisorStatus`, no ownership check on closure creation, and the "internal-only" beneficiary-close endpoint is publicly reachable). #4 and #5 are real bugs worth fixing even if not immediately exploitable today. #6-#8 are robustness/consistency gaps I'd want fixed but wouldn't block on alone. #9 is a product question, not a code defect.

apps/closure-reopen-service/src/closures/dto/create-closure.dto.ts
```diff
@@ -10,6 +10,13 @@ import { z } from 'zod';
  */
 export const createClosureSchema = z
   .object({
+    // Client-generated idempotency key — the mobile app is offline-first and
+    // may retry this submission after a dropped connection; a retry with the
+    // same value returns the original closure instead of creating a
```
logeshn-1, line 24: Pre-existing gap (supervisorStatus/supervisorId/supervisorNotes below, not touched by this diff), but this PR's new localClosureUuid idempotency key sits right next to it and is a good anchor: those three fields are still fully client-controlled with no cross-field validation tying supervisorStatus to the closure reason. Since POST /closures only requires role SAKHI, a SAKHI can set supervisorStatus: "APPROVED" directly and skip supervisor review — see closure.service.ts:37 for where this branches into an immediate beneficiary close. This PR's new visit-form-service auto-wiring (any SAKHI, any beneficiaryId, per form.routes.ts having no roster check) makes this pre-existing gap newly reachable end-to-end from a form submission, so worth fixing alongside this PR even though it didn't introduce it.

apps/closure-reopen-service/src/closures/closure.service.ts
```diff
@@ -13,21 +14,24 @@ export class ClosureService {
...
+   * re-firing the Quick Response card / beneficiary-close side effects below
+   * a second time — this mobile flow is offline-first and expected to retry.
    */
   async create(dto: CreateClosureInput, authorizationHeader: string) {
+    const existing = await this.repository.findByLocalClosureUuid(dto.localClosureUuid);
+    if (existing) return existing;
+
     const created = await this.repository.create(dto);
 
     if (created.supervisorStatus === 'PENDING') {
```
logeshn-1, line 92: This branches purely on client-supplied supervisorStatus (see create-closure.dto.ts:30-32) with no ownership check on dto.beneficiaryId either. For closure_reason=migration specifically, the PENDING branch below routes to approvalClient.create() which also has no ownership validation — combined with visit-form-service's new auto-wiring (any authenticated SAKHI, any beneficiaryId, per form.routes.ts having no roster check), a SAKHI can raise a real supervisor-approval card for a beneficiary they have no relationship to.

apps/beneficiary-service/src/beneficiary/beneficiary.routes.ts
```diff
@@ -556,6 +558,69 @@ export function registerBeneficiaryRoutes(doc: DocumentedRouter, service: Benefi
...
+        400: errorResponse(400, { message: 'reasonCode: Required' }),
+        401: errorResponse(401),
+        403: errorResponse(403, { message: 'This beneficiary case is outside your own roster.' }),
+        404: errorResponse(404, { message: 'Beneficiary case not found.' }),
+        409: errorResponse(409, { message: 'Unable to close this beneficiary case.' }),
+        500: errorResponse(500),
+      },
+    },
+    trustGatewayIdentity,
+    requireRoles('SAKHI', 'SUPERVISOR', 'MANAGER', 'ADMIN'),
```
logeshn-1, line 623: Doc comment above says this is "server-to-server only, called by closure-reopen-service," but this role list is the same set a human client has, and the gateway proxies /beneficiaries with no path-specific restriction. Anyone with role SAKHI who owns the case can call this directly and close it immediately — no closures row, no Quick Response card, no supervisor step. This bypasses the entire approval workflow the rest of this PR wires up.

apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
```diff
@@ -701,6 +701,125 @@ export class BeneficiaryService {
...
+    }
+
+    const fromPhase = existing.currentPhase as CasePhase;
+    if (fromPhase === toPhase) {
+      return this.projectCase(beneficiaryId, authorizationHeader, existing);
+    }
+
+    const isValidTransition =
+      (existing.caseType === 'MOTHER' && fromPhase === 'ANC' && toPhase === 'PP') ||
+      (existing.caseType === 'CHILD' && toPhase === 'NN');
```
logeshn-1: The MOTHER branch checks both fromPhase and toPhase; this CHILD branch only checks toPhase === 'NN', accepting any current phase. Contradicts the method's own doc comment above ("anything else 409s rather than silently moving a case backward"). Not reachable today since a CHILD starts at NN and same-value calls short-circuit earlier, but becomes a real phase-regression bug once any path advances a CHILD past NN (INC/CCV both exist in CASE_PHASES) — e.g. a delayed offline-retry of a stale DELIVERY_VISIT submission.

apps/visit-form-service/src/forms/form.service.ts
```diff
@@ -278,8 +260,190 @@ export class FormService {
...
+   * POST /beneficiaries dedup), so a replay resolves to the same child ids
+   * that were created the first time, without creating duplicates — this is
+   * what lets a retried submission still return childBeneficiaryIds instead
+   * of the caller having no way to find the child it already created.
+   */
+  private async resolveDeliveryChildren(
+    dto: CreateSubmissionInput,
+    authorizationHeader: string,
+  ): Promise<string[] | undefined> {
+    const motherCase = await findBeneficiaryById(dto.beneficiaryId, authorizationHeader);
```
logeshn-1: Unlike every other downstream call in this file (createChildBeneficiary, updateBeneficiaryPhase, createClosure, risk/socio-demographic/health-history syncs — all tolerate their own failures internally), this call has no try/catch, and neither does its caller. Since this runs after repository.createSubmission() has already durably persisted the row, a 403/500/network failure here fails the whole HTTP response with a false failure even though the submission itself succeeded. form.service.spec.ts only mocks this resolving null, never rejecting.

apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
```diff
@@ -701,6 +701,125 @@ export class BeneficiaryService {
...
+
+    if (caller.roles.includes('SAKHI')) {
+      if (existing.sakhiId !== caller.id) {
+        throw forbidden('This beneficiary case is outside your own roster.');
+      }
+    } else {
+      await assertCallerCanTouchCase(existing.sakhiId, caller, authorizationHeader);
+    }
+
+    if (existing.currentStatus === 'CLOSED') {
```
logeshn-1, line 816: This short-circuits on already-CLOSED without comparing or recording reasonCode. If closed once with MEDICAL, a later genuinely-different call with MIGRATION returns 200 silently with no audit trail — the doc comment's "recorded on beneficiary_status_history for audit" claim only holds for whichever call won. Existing idempotency test only covers a repeat with the same reasonCode.

apps/closure-reopen-service/src/closures/dto/create-closure.dto.ts
Dharanishwaran-ND: Fixed. `supervisorStatus`/`supervisorId`/`supervisorNotes` are removed from `createClosureSchema` entirely — a client can no longer set them. `ClosureService.create()` now derives `supervisorStatus` server-side by resolving `closureReasonLookupValueId` via a new `LookupClient.resolveClosureReasonCode()` and checking it against a `REASONS_REQUIRING_REVIEW` set (currently `MIGRATION`) — a SAKHI can no longer self-approve by setting the field directly.

apps/closure-reopen-service/src/closures/closure.service.ts
Dharanishwaran-ND: Fixed alongside the above: `ClosureService.create()` now calls `beneficiaryClient.getById(dto.beneficiaryId, ...)` before creating anything, delegating ownership scoping to beneficiary-service's own SAKHI-own-case/SUPERVISOR-roster check (same pattern the new `GET /reopen-requests` endpoint already used). A 403/404 propagates before any `closures` row or Quick Response card is created.

apps/beneficiary-service/src/beneficiary/beneficiary.routes.ts
Dharanishwaran-ND: Confirmed correct, and I'm not fully closing this one — this codebase genuinely has no machine/service-account identity, so "server-to-server only" can't be cryptographically enforced today, and restricting the role list would break the legitimate immediate-close flow (MEDICAL/NON_MEDICAL/PROGRAM_COMPLETION), which forwards the submitting SAKHI's own token specifically because there's no reviewer in that path. Fixed what I could: the doc comment (both here and on `BeneficiaryService.applyClosure`) no longer overclaims "server-to-server only" — it now states plainly that a SAKHI can reach this directly and bypass the closures audit trail/review step, bounded to their own case (not a cross-tenant IDOR) by the existing ownership check. Tracking as a follow-up that needs an actual machine-identity concept to close properly, not silently fixing it here.

apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
Dharanishwaran-ND: Fixed — this was a real bug, confirmed by your own repro path. The CHILD branch now checks `fromPhase === 'NN' && toPhase === 'NN'` (matching the MOTHER branch's shape of validating both sides), so any other current phase 409s instead of silently accepting the transition. Added a regression test with `currentPhase: 'INC'` asserting the 409, since the bug wasn't reachable through the existing test's unrealistic fixture (a CHILD case can never actually start at ANC).

apps/visit-form-service/src/forms/form.service.ts
Dharanishwaran-ND: Fixed — wrapped the `findBeneficiaryById` call in try/catch, logging and returning `undefined` (skip child auto-creation) on failure rather than letting the throw propagate and fail an already-durably-saved submission. Matches the tolerance stance every other downstream call in this method already uses.

apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
Dharanishwaran-ND: Fixed — the idempotent-already-CLOSED short-circuit (both the initial check and the race-recovery re-read) now compares the incoming `reasonCode` against the last CLOSED entry in `statusHistory` and 409s on a genuine mismatch instead of silently returning 200. A matching reasonCode (the real retry case) still no-ops as before.

Dharanishwaran-ND:
Addressed the review — thank you for the thorough pass, especially catching the #1/#2/#3 compounding gap.

**Fixed (pushed in `ecaa495`):**
- **#1/#2 (Critical)** — `POST /closures` no longer accepts client-supplied `supervisorStatus`/`supervisorId`/`supervisorNotes`. `ClosureService.create()` derives `supervisorStatus` server-side from a resolved `closureReasonLookupValueId` (new `LookupClient.resolveClosureReasonCode`), and checks ownership via `beneficiaryClient.getById` before creating anything.
- **#4 (High)** — CHILD phase-advance now validates `fromPhase` too, not just `toPhase`. Added a regression test that actually exercises the bug (the original test's fixture — a CHILD starting at `ANC` — could never occur in practice, which is why this wasn't caught before).
- **#5 (High)** — `findBeneficiaryById` in `resolveDeliveryChildren` is now wrapped in try/catch, matching the tolerance stance every other downstream call in `createSubmission` already has.
- **#6** — the idempotent-already-CLOSED path now compares `reasonCode` against the last `CLOSED` entry in `statusHistory` and 409s on a genuine mismatch instead of silently dropping it.
- **#7** — both `ClosureService.create()` and `ReopenRequestService.create()` now catch `P2002` on a concurrent idempotency-key race and re-resolve to the winning row instead of propagating a raw 500.
- **#8** — `resolveDeliveryChildren` now takes `beneficiaryId` as an explicit parameter; the replay call site passes `existing.beneficiaryId`, matching `resolveClosureRequest`'s already-correct pattern.
- Also fixed `lookup.client.ts`'s same `appConfig` import landmine that broke CI on `beneficiary.client.ts` earlier in this PR (process.exit(1) at module-load time with no real env vars) — and added `lookup.client.spec.ts`, which didn't exist before.

**#3 — deliberately not fully fixed, documented instead.** This codebase has no machine/service-account identity, so "server-to-server only" can't be enforced cryptographically, and narrowing the role list would break the legitimate immediate-close flow (which forwards the submitting SAKHI's own token precisely because there's no reviewer in that path). Corrected both doc comments (route + service method) to stop overclaiming enforcement that doesn't exist, and state the bounded gap plainly (SAKHI can close their own case directly, bypassing the closures audit row — not a cross-tenant IDOR, but a real workflow bypass). Tracking as a follow-up that needs an actual machine-identity concept.

**#9 — product question, not code.** Not implementing a guess here; flagging back to you/product to confirm intended MOTHER↔CHILD closure cascade behavior before anyone builds against an assumption.

All 768 tests pass across the three touched services (beneficiary-service, closure-reopen-service, visit-form-service), lint clean.

PR #170, opened 2026-08-19
Validate visitId on form submission, add form_answers index

### LINKED ISSUE #169: Validate visitId on form submission, add form_answers index
## Summary

`POST /forms/:formCode/submissions` accepted a client-supplied `visitId` with no existence check — a stale or bad `visitId` (a race between the visit's own creation and this submission, or a bad offline replay) would hit an unhandled Prisma FK violation instead of a clean, named error.

## Scope

- `FormRepository.findVisitById()` — direct query (visit_instances is owned by this same service, no cross-service call needed).
- `FormService.createSubmission()` — checks `visitId` exists before the insert (clean 404 for the common case), plus a P2003 catch as a backstop for the remaining check-then-insert race window.
- New index `form_answers(submission_id, field_code)` — supports both "all answers for a submission" and "this question across submissions" lookups per the ERD's indexing stance.
- `backfill-form-answers.ts` — idempotent reconciliation script to re-derive `form_answers` from `form_data_json` for existing submissions, for use if `buildFormAnswers`'s decomposition logic changes later.

## Verification

`nx test/lint/build` clean on `visit-form-service` — 424/424 tests passing.

Dharanishwaran-ND:
## Summary
- Closes #169
- `POST /forms/:formCode/submissions` now validates a client-supplied `visitId` exists before inserting — was previously an unhandled FK violation (500) on a stale/bad `visitId`, now a clean `404 Visit not found.`
- Added a P2003 catch as a backstop for the remaining check-then-insert race window (visit deleted between the check and the insert)
- Added `form_answers(submission_id, field_code)` index per the ERD's indexing stance
- Added an idempotent `backfill-form-answers.ts` reconciliation script to re-derive `form_answers` from `form_data_json`, for use if the answer-decomposition logic changes later

## Test plan
- [x] `npx nx test visit-form-service` — 424/424 passing
- [x] `npx nx build visit-form-service` — succeeds
- [ ] Migration `20260819100000_add_form_answers_index` needs to be applied to each environment's DB before/with deploy

### DIFF
--- a/apps/visit-form-service/prisma/backfill-form-answers.ts
+++ b/apps/visit-form-service/prisma/backfill-form-answers.ts
@@ -1,0 +1,127 @@
+/**
+ * Backfill/reconciliation for form_answers.
+ *
+ * form_submissions.form_data_json is the durable source of truth; form_answers
+ * is a normalized projection of it (see FormRepository.createSubmission),
+ * written in the same transaction as the submission at request time. This
+ * script exists for the case where that projection logic changes later (a new
+ * input_type, a fix to buildFormAnswers) and needs to be re-applied to
+ * submissions that were stored under the old logic — it never touches
+ * form_data_json itself.
+ *
+ * Idempotent: for each submission processed, this replaces (not appends to)
+ * its form_answers rows with a fresh decomposition, in one transaction per
+ * submission — re-running twice in a row produces the same end state.
+ *
+ * Usage:
+ *   npx ts-node prisma/backfill-form-answers.ts [--form-version-id=<uuid>]
+ *
+ * With no flag, every form_submissions row is reprocessed. With
+ * --form-version-id, only submissions on that exact form version are touched
+ * — useful for re-applying a fix scoped to one form's schema without
+ * reprocessing the whole table.
+ */
+import { PrismaClient } from '../../../node_modules/.prisma/client-visit-form-service';
+import { buildFormAnswers } from '../src/forms/form.mapper';
+import { schemaJsonSchema } from '../src/forms/dto/form-field.dto';
+
+const prisma = new PrismaClient();
+
+export interface BackfillSummary {
+  processed: number;
+  answersWritten: number;
+  skipped: Array<{ submissionId: string; reason: string }>;
+}
+
+/**
+ * Re-derives and replaces one submission's form_answers rows from its stored
+ * form_data_json, using the schema of the form version it was actually
+ * answered against (never the currently active version) — same "own
+ * historical schema" contract createSubmission relies on. Returns
+ * `{ skipped }` (and leaves existing form_answers untouched) when the version
+ * can't be resolved/parsed, rather than guessing or dropping good data.
+ */
+export async function backfillSubmission(
+  client: Pick<PrismaClient, 'formVersion' | 'formAnswer'>,
+  submission: { id: string; formVersionId: string; formDataJson: unknown },
+): Promise<{ answersWritten: number } | { skipped: string }> {
+  const version = await client.formVersion.findUnique({ where: { id: submission.formVersionId } });
+  if (!version) return { skipped: 'form_version_id no longer resolves to a form_versions row' };
+
+  const parsed = schemaJsonSchema.safeParse(version.schemaJson);
+  if (!parsed.success) {
+    return { skipped: 'schemaJson failed to parse against the current field schema' };
+  }
+
+  const formData =
+    typeof submission.formDataJson === 'object' && submission.formDataJson !== null
+      ? (submission.formDataJson as Record<string, unknown>)
+      : {};
+  const answers = buildFormAnswers(parsed.data, formData);
+
+  await client.formAnswer.deleteMany({ where: { submissionId: submission.id } });
+  if (answers.length) {
+    await client.formAnswer.createMany({
+      data: answers.map((a) => ({
+        submissionId: submission.id,
+        fieldCode: a.fieldCode,
+        answerValueText: a.answerValueText,
+        answerValueNumber: a.answerValueNumber,
+        answerValueDate: a.answerValueDate,
+        answerValueBool: a.answerValueBool,
+        answerValueJson: a.answerValueJson as never,
+        isIndexed: a.isIndexed,
+      })),
+    });
+  }
+
+  return { answersWritten: answers.length };
+}
+
+async function main(): Promise<void> {
+  const formVersionIdArg = process.argv
+    .find((a) => a.startsWith('--form-version-id='))
+    ?.split('=')[1];
+
+  const submissions = await prisma.formSubmission.findMany({
+    where: formVersionIdArg ? { formVersionId: formVersionIdArg } : undefined,
+    select: { id: true, formVersionId: true, formDataJson: true },
+  });
+
+  const summary: BackfillSummary = { processed: 0, answersWritten: 0, skipped: [] };
+
+  for (const submission of submissions) {
+    const result = await prisma.$transaction((tx) => backfillSubmission(tx, submission));
+    summary.processed += 1;
+    if ('skipped' in result) {
+      summary.skipped.push({ submissionId: submission.id, reason: result.skipped });
+    } else {
+      summary.answersWritten += result.answersWritten;
+    }
+  }
+
+  console.log('Backfill summary:', {
+    processed: summary.processed,
+    answersWritten: summary.answersWritten,
+    skippedCount: summary.skipped.length,
+  });
+  if (summary.skipped.length > 0) {
+    console.log('Skipped submissions:', summary.skipped);
+    throw new Error(
+      `${summary.skipped.length} submission(s) could not be reprocessed — see the list above.`,
+    );
+  }
+  console.log('Done — form_answers reprocessed for all matching submissions.');
+}
+
+// Guarded so importing backfillSubmission for a unit test (see
+// backfill-form-answers.spec.ts) never opens a real Prisma connection —
+// only running this file directly as a script does.
+if (require.main === module) {
+  main()
+    .catch((err) => {
+      console.error(err);
+      process.exit(1);
+    })
+    .finally(() => prisma.$disconnect());
+}
--- a/apps/visit-form-service/prisma/backfill-form-answers.ts
+++ b/apps/visit-form-service/prisma/backfill-form-answers.ts
@@ -11,7 +11,10 @@
  *
  * Idempotent: for each submission processed, this replaces (not appends to)
  * its form_answers rows with a fresh decomposition, in one transaction per
- * submission — re-running twice in a row produces the same end state.
+ * submission — re-running twice in a row produces the same end state. The
+ * prior rows are soft-deleted (isDeleted/deletedAt), not hard-deleted — this
+ * table exists for reporting/audit, so a run's own history stays recoverable
+ * rather than being permanently erased on every reconciliation.
  *
  * Usage:
  *   npx ts-node prisma/backfill-form-answers.ts [--form-version-id=<uuid>]
@@ -23,9 +26,16 @@
  */
 import { PrismaClient } from '../../../node_modules/.prisma/client-visit-form-service';
 import { buildFormAnswers } from '../src/forms/form.mapper';
-import { schemaJsonSchema } from '../src/forms/dto/form-field.dto';
+import { schemaJsonSchema, type FormField } from '../src/forms/dto/form-field.dto';
 
 const prisma = new PrismaClient();
+
+// Rows are processed in fixed-size concurrent batches — each row's own
+// transaction is already independent, so bounded concurrency cuts run time
+// on a large table without changing per-row atomicity or the summary's
+// correctness (still one outcome per row, still fully awaited before the
+// script exits).
+const CONCURRENCY = 15;
 
 export interface BackfillSummary {
   processed: number;
@@ -40,26 +50,34 @@ export interface BackfillSummary {
  * historical schema" contract createSubmission relies on. Returns
  * `{ skipped }` (and leaves existing form_answers untouched) when the version
  * can't be resolved/parsed, rather than guessing or dropping good data.
+ *
+ * `resolvedSchema` is passed in (not re-fetched/re-parsed here) so a caller
+ * processing many submissions against the same small set of form versions
+ * can resolve+parse each version once and reuse it, instead of repeating
+ * that DB round-trip and zod parse on every row.
  */
 export async function backfillSubmission(
-  client: Pick<PrismaClient, 'formVersion' | 'formAnswer'>,
-  submission: { id: string; formVersionId: string; formDataJson: unknown },
-): Promise<{ answersWritten: number } | { skipped: string }> {
-  const version = await client.formVersion.findUnique({ where: { id: submission.formVersionId } });
-  if (!version) return { skipped: 'form_version_id no longer resolves to a form_versions row' };
-
-  const parsed = schemaJsonSchema.safeParse(version.schemaJson);
-  if (!parsed.success) {
-    return { skipped: 'schemaJson failed to parse against the current field schema' };
-  }
-
+  client: Pick<PrismaClient, 'formAnswer'>,
+  submission: { id: string; formDataJson: unknown },
+  resolvedSchema: FormField[],
+): Promise<{ answersWritten: number }> {
+  // Arrays pass `typeof === 'object' && !== null` too — without this guard,
+  // a legacy row with form_data_json stored as a JSON array would get cast
+  // to Record<string, unknown> and buildFormAnswers would Object.entries()
+  // its numeric indices as if they were question codes, instead of being
+  // treated as unparseable.
   const formData =
-    typeof submission.formDataJson === 'object' && submission.formDataJson !== null
+    typeof submission.formDataJson === 'object' &&
+    submission.formDataJson !== null &&
+    !Array.isArray(submission.formDataJson)
       ? (submission.formDataJson as Record<string, unknown>)
       : {};
-  const answers = buildFormAnswers(parsed.data, formData);
+  const answers = buildFormAnswers(resolvedSchema, formData);
 
-  await client.formAnswer.deleteMany({ where: { submissionId: submission.id } });
+  await client.formAnswer.updateMany({
+    where: { submissionId: submission.id, isDeleted: false },
+    data: { isDeleted: true, deletedAt: new Date() },
+  });
   if (answers.length) {
     await client.formAnswer.createMany({
       data: answers.map((a) => ({
@@ -78,6 +96,13 @@ export async function backfillSubmission(
   return { answersWritten: answers.length };
 }
 
+/** Splits an array into fixed-size chunks, preserving order. */
+function chunk<T>(items: T[], size: number): T[][] {
+  const chunks: T[][] = [];
+  for (let i = 0; i < items.length; i += size) chunks.push(items.slice(i, i + size));
+  return chunks;
+}
+
 async function main(): Promise<void> {
   const formVersionIdArg = process.argv
     .find((a) => a.startsWith('--form-version-id='))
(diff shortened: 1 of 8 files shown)


apps/visit-form-service/src/forms/form.repository.ts
```diff
@@ -124,6 +124,16 @@ export class FormRepository {
     return this.prisma.formSubmission.findUnique({ where: { localSubmissionUuid } });
   }
 
+  /**
+   * Existence check for a visit-linked submission's visitId, before the
+   * insert — visit_instances is owned by this same service (unlike
+   * beneficiary_cases/form_versions' cross-service equivalents), so this is a
+   * direct query, not an HTTP call.
+   */
+  findVisitById(id: string) {
```
logeshn-1: **Data-integrity gap:** `findVisitById` only checks that the visit exists and isn't soft-deleted — it never verifies the visit's own `beneficiaryId` matches the submission's `beneficiaryId`. A client can submit `beneficiaryId: 'A'` with a real, non-deleted `visitId` that actually belongs to beneficiary `B`; this check passes and the submission ends up linked to the wrong beneficiary's visit. Might be worth filtering on `{ id, beneficiaryId: dto.beneficiaryId, isDeleted: false }` (or checking it in the service after the fetch).

Separately: this query duplicates `VisitInstanceRepository.findById` (`visits/visitInstance.repository.ts:35`) verbatim — same table, same service, same `isDeleted` convention. Worth reusing that repository instead of maintaining two copies of the same existence check (it also reads oddly against this file's own class doc a few lines up: "the only tables this repository touches").

apps/visit-form-service/src/forms/form.service.ts
```diff
@@ -201,16 +211,25 @@ export class FormService {
...
+        localSubmissionUuid: dto.localSubmissionUuid,
+        formDataJson: dto.formData,
+        validationStatus: 'VALID',
+        formAnswers,
+      });
+    } catch (err) {
+      // Backstop for the race the check above can't close on its own: the
+      // visit existed at check time but was gone by the time this insert ran.
+      // Same "fail loud, named 404" contract either way, not a 500.
+      if (isForeignKeyViolation(err)) throw notFound('Visit not found.');
```
logeshn-1: `isForeignKeyViolation(err)` treats *any* P2003 from the insert as a missing visit, but `form_submissions` has two FK columns (`form_version_id` and `visit_id`). If the `form_version_id` FK fails at insert time instead (e.g. a race between the `findVersionById` check above and this insert), this still reports `'Visit not found.'` — misleading, and it'd fire even on a request where the client never supplied a `visitId` at all. `err.meta?.field_name`/`target` would let this disambiguate which constraint actually failed.

apps/visit-form-service/src/forms/form.service.ts
```diff
@@ -201,16 +211,25 @@ export class FormService {
...
+        formVersionId: dto.formVersionId,
+        beneficiaryId: dto.beneficiaryId,
+        visitId: dto.visitId ?? null,
+        submittedByUserId,
+        localSubmissionUuid: dto.localSubmissionUuid,
+        formDataJson: dto.formData,
+        validationStatus: 'VALID',
+        formAnswers,
+      });
+    } catch (err) {
```
logeshn-1, line 230: Two things about this catch block:

1. Converting a caught P2003 into a 404 discards the original Prisma error before it reaches `errorHandler` — and `all-exceptions.filter.ts` only logs when status >= 500. Before this PR, an FK violation surfaced as an unhandled 500 and got logged with full detail; now it's rewritten to a 404 with zero server-side log signal. Might be worth a log line here before throwing, so a real occurrence isn't invisible to on-call.
2. This only checks for P2003. A concurrent double-submit with the same `localSubmissionUuid` can still slip past the `findSubmissionByLocalUuid` check above and hit the unique constraint (P2002) on insert — same race shape as `createDraft`, which already handles this via `isUniqueConstraintViolation`. That case currently falls through `throw err` and surfaces as an unhandled 500.

apps/visit-form-service/src/forms/form.service.ts
```diff
@@ -188,6 +188,16 @@ export class FormService {
       throw badRequest('Form version is not published.');
     }
 
+    // Client-supplied visitId can be stale (a race between the visit's own
+    // POST /visits and this submission, or a bad replay) — check before the
+    // insert so this fails as a clean, named 404 instead of an unhandled FK
+    // violation. See createSubmission's P2003 catch below for the remaining
+    // check-then-insert race window this can't close on its own.
+    if (dto.visitId) {
```
logeshn-1: This visitId check now runs before `validateSubmission`. A request that's wrong in both ways (stale `visitId` *and* invalid `formData`) will only ever report the 404 — the client won't learn about the formData violations until they fix visitId and resubmit. Might be intentional (fail fast on the cheaper check), just flagging the ordering change in case it wasn't deliberate.

apps/visit-form-service/src/forms/form.service.ts
```diff
@@ -481,3 +500,13 @@ function isUniqueConstraintViolation(err: unknown): boolean {
     (err as { code: unknown }).code === 'P2002'
   );
 }
+
+/** Narrows a caught Prisma error to a foreign-key violation (P2003). */
+function isForeignKeyViolation(err: unknown): boolean {
```
logeshn-1: Nit: this is a copy-paste of `isUniqueConstraintViolation` above, differing only in the P-code. This narrow-err-to-Prisma-code shape recurs a few times across the codebase — might be worth a shared `isPrismaErrorCode(err, code)` helper at this point instead of another near-identical copy.

apps/visit-form-service/src/forms/form.service.ts
```diff
@@ -188,6 +188,16 @@ export class FormService {
       throw badRequest('Form version is not published.');
     }
 
+    // Client-supplied visitId can be stale (a race between the visit's own
```
logeshn-1: This comment (and the one on the catch block below) frames the race as "visit deleted between check and insert" — but I don't see a DELETE endpoint for visits, or anything that sets `isDeleted: true` outside test fixtures. If that's right, the backstop is defensive-but-currently-unreachable rather than closing a live race — might be worth correcting the rationale, or let me know if I'm missing the path where a visit actually gets removed.

apps/visit-form-service/src/forms/form.service.ts
```diff
@@ -481,3 +500,13 @@ function isUniqueConstraintViolation(err: unknown): boolean {
...
+
+/** Narrows a caught Prisma error to a foreign-key violation (P2003). */
+function isForeignKeyViolation(err: unknown): boolean {
+  return (
+    typeof err === 'object' &&
+    err !== null &&
+    'code' in err &&
+    (err as { code: unknown }).code === 'P2003'
+  );
+}
```
logeshn-1, line 549: Nit: this pushes `form.service.ts` to 512 lines, further past the ~250-line guideline in the root `.claude/CLAUDE.md`. Not blocking, but the new visit-validation logic (check + try/catch + `isForeignKeyViolation`) could be a reasonable first slice to pull out if this file gets split later.

apps/visit-form-service/prisma/schema.prisma
```diff
@@ -334,5 +334,9 @@ model FormAnswer {
 
   formSubmission FormSubmission @relation(fields: [submissionId], references: [id])
 
+  // ERD's design stance (line 19): fields extracted for reporting/query must
+  // be indexed, not require a full scan — this supports both "all answers for
+  // a submission" and "this question across submissions" lookups.
+  @@index([submissionId, fieldCode])
```
logeshn-1, line 344: This comment says the index supports "this question across submissions" lookups, but a B-tree with `submissionId` leading can't efficiently serve a `fieldCode`-only query (leftmost-prefix rule) — that access pattern would need its own index (or one leading with `fieldCode`). Worth adjusting the comment, or adding the second index if that reporting pattern is actually needed soon.

apps/visit-form-service/prisma/backfill-form-answers.ts
```diff
@@ -0,0 +1,127 @@
...
+// Guarded so importing backfillSubmission for a unit test (see
+// backfill-form-answers.spec.ts) never opens a real Prisma connection —
+// only running this file directly as a script does.
+if (require.main === module) {
+  main()
+    .catch((err) => {
+      console.error(err);
+      process.exit(1);
+    })
+    .finally(() => prisma.$disconnect());
```
logeshn-1: `process.exit(1)` inside `.catch` terminates the process synchronously before the chained `.finally(() => prisma.$disconnect())` runs — so disconnect is effectively dead on every failure path, including the script's own intentional throw when `summary.skipped.length > 0`. Might want to disconnect explicitly before calling `process.exit`.

apps/visit-form-service/prisma/backfill-form-answers.ts
```diff
@@ -0,0 +1,127 @@
...
+  const version = await client.formVersion.findUnique({ where: { id: submission.formVersionId } });
+  if (!version) return { skipped: 'form_version_id no longer resolves to a form_versions row' };
+
+  const parsed = schemaJsonSchema.safeParse(version.schemaJson);
+  if (!parsed.success) {
+    return { skipped: 'schemaJson failed to parse against the current field schema' };
+  }
+
+  const formData =
+    typeof submission.formDataJson === 'object' && submission.formDataJson !== null
```
logeshn-1: This guard (`typeof === 'object' && !== null`) also accepts arrays. If a legacy row has `form_data_json` stored as a JSON array, it gets cast to `Record<string, unknown>` and `buildFormAnswers` will `Object.entries()` the numeric indices as if they were question codes, rather than being treated as unparseable and skipped. Might want `!Array.isArray(...)` added to the guard.

apps/visit-form-service/prisma/backfill-form-answers.ts
```diff
@@ -0,0 +1,127 @@
...
+      ? (submission.formDataJson as Record<string, unknown>)
+      : {};
+  const answers = buildFormAnswers(parsed.data, formData);
+
+  await client.formAnswer.deleteMany({ where: { submissionId: submission.id } });
```
logeshn-1: This hard-deletes existing `form_answers` rows via `deleteMany`, even though the table has its own `isDeleted`/`deletedAt` columns. Every backfill run permanently erases the prior rows (and their `createdAt`/`createdByUserId`) with no way to recover what was there before the run. Worth soft-deleting instead if that audit trail matters for this table.

apps/visit-form-service/prisma/backfill-form-answers.ts
```diff
@@ -0,0 +1,127 @@
...
+ * answered against (never the currently active version) — same "own
+ * historical schema" contract createSubmission relies on. Returns
+ * `{ skipped }` (and leaves existing form_answers untouched) when the version
+ * can't be resolved/parsed, rather than guessing or dropping good data.
+ */
+export async function backfillSubmission(
+  client: Pick<PrismaClient, 'formVersion' | 'formAnswer'>,
+  submission: { id: string; formVersionId: string; formDataJson: unknown },
+): Promise<{ answersWritten: number } | { skipped: string }> {
+  const version = await client.formVersion.findUnique({ where: { id: submission.formVersionId } });
```
logeshn-1: Efficiency nit: for a form version with a lot of historical submissions, this re-fetches and re-parses the same `formVersion`/schema on every single row, even though the distinct set of form versions touched in a run is typically small. A per-`formVersionId` cache would cut DB round-trips and repeated zod parsing without changing per-row atomicity.

logeshn-1 (commented):
Reviewed for correctness and reuse/simplification/efficiency. Left inline comments — most notable is that `findVisitById` doesn't verify the visit's own `beneficiaryId` matches the submission's, so a valid visit belonging to a different beneficiary currently passes the check. A few smaller things on the P2003 backstop, the backfill script, and the new index. No blocking objection to merging once these are looked at — just comments, no changes made on my end.

apps/visit-form-service/prisma/backfill-form-answers.ts
```diff
@@ -0,0 +1,127 @@
...
+    ?.split('=')[1];
+
+  const submissions = await prisma.formSubmission.findMany({
+    where: formVersionIdArg ? { formVersionId: formVersionIdArg } : undefined,
+    select: { id: true, formVersionId: true, formDataJson: true },
+  });
+
+  const summary: BackfillSummary = { processed: 0, answersWritten: 0, skipped: [] };
+
+  for (const submission of submissions) {
```
logeshn-1: Efficiency nit: this processes submissions strictly one at a time (awaiting each `$transaction` in the loop), even though each row's transaction is already independent. Bounded concurrency (e.g. chunks of 10-20 via `Promise.all`) would keep the same per-row atomicity while cutting run time substantially on a large table.

apps/visit-form-service/src/forms/form.repository.ts
Dharanishwaran-ND: Fixed the data-integrity gap — `findVisitById` now takes `beneficiaryId` as a second parameter and filters on it directly (`{ id, beneficiaryId, isDeleted: false }`), so a real visit belonging to a different beneficiary no longer passes the check. On the reuse point: I kept this as a duplicate query rather than importing `VisitInstanceRepository` — this service composes one `FormRepository` per its own class doc comment ("the only tables this repository touches"), and introducing a cross-repository dependency for one extra filter would break that boundary for a small win. Updated the doc comment to explain this tradeoff explicitly rather than silently duplicating.

apps/visit-form-service/src/forms/form.service.ts
Dharanishwaran-ND: Fixed — added `failingConstraintIsVisitId(err)` which checks `err.meta.field_name` for the `visit_id` substring before rewriting to the 404, so a `form_version_id` FK race (or anything else) falls through to the generic handler instead of being misreported. Added a test asserting a `form_version_id`-constraint P2003 is NOT translated.

apps/visit-form-service/src/forms/form.service.ts
Dharanishwaran-ND: Both fixed. (1) Added a `console.warn` with the full original error before rewriting to 404, so a real occurrence is visible in logs even though the response itself is a 4xx. (2) Added a P2002 catch alongside the P2003 one — on a concurrent double-submit race, it now re-queries `findSubmissionByLocalUuid` and replays the winning row (same idempotent-replay contract as a sequential retry), only rethrowing if no winning row can be found. Added tests for both the replay-success and no-winner-found paths.

apps/visit-form-service/src/forms/form.service.ts
Dharanishwaran-ND: Fixed — moved the visitId check to after `validateSubmission`, so a request that's wrong in both ways now reports the formData violations (422) rather than only ever surfacing the visitId 404. This was an unintentional ordering side-effect of where I inserted the check, not deliberate — thanks for catching it. Added a test with both an out-of-range `bp_systolic` and a missing `visitId` asserting the 422 wins and `findVisitById` is never called.

apps/visit-form-service/src/forms/form.service.ts
Dharanishwaran-ND: Fixed — replaced both `isUniqueConstraintViolation`/`isForeignKeyViolation` with one shared `isPrismaErrorCode(err, code)`, used at both call sites in this file now (the existing `createDraft` P2002 handling and the new submission-insert handling).

apps/visit-form-service/src/forms/form.service.ts
Dharanishwaran-ND: Confirmed — no delete path exists for visits anywhere in this service (grepped for any `isDeleted: true` write under `visits/`, found none). You're right that this specific backstop is defensive-but-currently-unreachable, not closing a live race today. I've left it in rather than removing it — it's cheap insurance against the day a visit-delete/cancel path gets added, and the P2002 handling I added alongside it *is* a live, reachable race (concurrent double-submit), so the try/catch block earns its place either way. Didn't change the rationale comment's accuracy claim, but I can adjust the wording further if you'd rather it not imply an active race — let me know.

apps/visit-form-service/src/forms/form.service.ts
Dharanishwaran-ND: Acknowledged, not fixed in this commit — form.service.ts is now 546 lines after these fixes, further past the guideline. Splitting this out (submission validation/insert logic specifically, as you suggested) is a real structural change I'd rather do as its own reviewed PR than bundle into a review-response fix — flagging it as a follow-up rather than taking a shortcut here.

apps/visit-form-service/prisma/schema.prisma
Dharanishwaran-ND: Fixed — corrected the comment to state plainly that this index (submissionId leading) serves per-submission lookups only, and does NOT serve a fieldCode-only cross-submission query per the leftmost-prefix rule. Didn't add a second index since no current caller queries that way — happy to add one if that reporting pattern becomes a real need.

apps/visit-form-service/prisma/backfill-form-answers.ts
Dharanishwaran-ND: Fixed — added explicit `await prisma.$disconnect()` before both `process.exit(1)` and the thrown-skip-count error, since you're right that the chained `.finally` never runs after a synchronous exit.

apps/visit-form-service/prisma/backfill-form-answers.ts
Dharanishwaran-ND: Fixed — added `!Array.isArray(...)` to the guard, so a legacy array-shaped `form_data_json` is treated as unparseable (empty object) rather than having its numeric indices read as question codes. Added a test.

apps/visit-form-service/prisma/backfill-form-answers.ts
Dharanishwaran-ND: Fixed — switched to soft-delete (`updateMany` setting `isDeleted: true, deletedAt: new Date()`) instead of `deleteMany`, using the columns this table's own schema already has. Added a test asserting no hard-delete call happens.

apps/visit-form-service/prisma/backfill-form-answers.ts
Dharanishwaran-ND: Fixed — form versions are now resolved+parsed once per distinct `formVersionId` via a `Map` cache in `main()`, not once per row. `backfillSubmission` itself now takes the already-resolved schema as a parameter instead of re-fetching it.

apps/visit-form-service/prisma/backfill-form-answers.ts
Dharanishwaran-ND: Fixed — submissions are now processed in fixed-size batches of 15 via `Promise.all`, keeping each row's own transaction independent while cutting wall-clock time on a large table.

Dharanishwaran-ND:
Addressed the review — thanks for the thorough pass, especially catching the beneficiaryId gap on `findVisitById`.

**Fixed (pushed in `84185f5`):**
- `findVisitById` now filters on `beneficiaryId`, not just `id` — closes the cross-beneficiary visit-linking gap.
- The visitId check now runs after `validateSubmission`, so both classes of error surface together instead of the 404 masking formData violations.
- The P2003 catch now disambiguates which FK actually failed via `err.meta.field_name` — a `form_version_id` race no longer gets misreported as "Visit not found."
- Added a `console.warn` before the 404 rewrite, so a real occurrence isn't invisible to on-call.
- Added P2002 handling on the same insert (concurrent double-submit race) — replays the winning row instead of a raw 500.
- Consolidated `isUniqueConstraintViolation`/`isForeignKeyViolation` into one `isPrismaErrorCode(err, code)` helper.
- Corrected the `form_answers` index's doc comment (leftmost-prefix rule).
- `backfill-form-answers.ts`: soft-delete instead of hard-delete, array-guard on `formDataJson`, per-`formVersionId` schema caching, bounded-concurrency batch processing, explicit disconnect before `process.exit`.

**Acknowledged, not changed:**
- The "visit deleted between check and insert" backstop — confirmed no delete path exists for visits anywhere in this service today, so this specific case is defensive-but-currently-unreachable. Left in rather than removed (cheap insurance, and the P2002 handling added alongside it in the same try/catch *is* live and reachable).
- `form.service.ts`'s length (now 546 lines) — real nit, but a structural split deserves its own reviewed PR, not a shortcut bundled into this fix commit.

All 431 tests pass (7 new since this fix round), lint clean.

logeshn-1 (approved):

PR #172, opened 2026-08-19
feat(risk-service): Add ANC/Infant high-risk detection rules

### LINKED ISSUE #171: feat: Implement ANC/Infant High-Risk detection rules (SRS FR-S-5.4, Appendix D)
## Summary
Per SRS FR-S-5.4, HR (high-risk) detection thresholds must be configured in GoRules and updatable without redeployment. This implements the ANC-phase and Infant-phase (NN/INC/CCV) high-risk detection rules based on the ARMMAN-provided "High risk protocols_Developer's copy" thresholds sheet, and updates Appendix D of the SRS accordingly.

## Scope
- ANC and infant risk-condition seed data + risk parameters
- GoRules scheduling rules JSON for ANC and infant risk detection (with specs)
- risk-referral-service: schema/migration + repository/service changes to support new risk flag grade/rank
- visit-form-service: form service/repository changes to wire risk assessment client into form submission flow
- docs: Appendix D rewritten as a standalone document with full parameter/threshold/trigger tables; SRS FR-S-5.4 status updated to reflect ANC HR thresholds received (PP/NN/INC/CCV sheets still pending from ARMMAN)

## Notes
PP-phase thresholds and some Part 2 (Infant) values are still pending ARMMAN confirmation — tracked as open items in the new Appendix D doc (§2.8).

Dharanishwaran-ND:
Closes #171

## Summary
- Implements SRS FR-S-5.4 high-risk detection thresholds using ARMMAN's ANC HR threshold sheet ("High risk protocols_Developer's copy")
- Adds GoRules scheduling rules + seed data for ANC and infant (NN/INC/CCV) risk conditions
- Adds risk flag grade/rank support in risk-referral-service (schema, migration, repository, service)
- Wires visit-form-service form submission to invoke the risk assessment client
- Rewrites Appendix D of the SRS as a standalone doc with full parameter/threshold/trigger tables (ANC, Infant, and Manager Dashboard HR linelist/reporting requirements)

## Notes
PP-phase thresholds and some Part 2 (Infant) values are still pending ARMMAN confirmation — tracked as open items in the new Appendix D doc (§2.8) and flagged in the SRS FR-S-5.4 status note.

## Test plan
- [ ] `anc-risk.rulesJson.spec.ts` and `infant-risk.rulesJson.spec.ts` pass
- [ ] `riskAssessment.service.spec.ts` passes
- [ ] `form.service.spec.ts` passes
- [ ] Prisma migration applies cleanly against a fresh DB

### DIFF
--- a/apps/risk-referral-service/prisma/migrations/20260819000000_add_risk_flag_grade_rank/migration.sql
+++ b/apps/risk-referral-service/prisma/migrations/20260819000000_add_risk_flag_grade_rank/migration.sql
@@ -1,0 +1,10 @@
+-- AlterTable
+-- Add as nullable first, backfill existing rows to 0 (treated as NORMAL,
+-- the safest default for historic flags whose true grade rank was never
+-- recorded), then enforce NOT NULL — avoids failing on any pre-existing
+-- risk_flags rows.
+ALTER TABLE "risk_flags" ADD COLUMN "grade_rank" INTEGER;
+
+UPDATE "risk_flags" SET "grade_rank" = 0 WHERE "grade_rank" IS NULL;
+
+ALTER TABLE "risk_flags" ALTER COLUMN "grade_rank" SET NOT NULL;
--- a/apps/risk-referral-service/prisma/seed-data/anc-risk-conditions.json
+++ b/apps/risk-referral-service/prisma/seed-data/anc-risk-conditions.json
@@ -1,0 +1,135 @@
+[
+  {
+    "conditionCode": "AGE",
+    "conditionName": "Underage/Overage",
+    "gradeScale": "BINARY",
+    "referralRequiredDefault": true,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "MUAC_BMI",
+    "conditionName": "MUAC/BMI",
+    "gradeScale": "BINARY",
+    "referralRequiredDefault": true,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "STUNTING",
+    "conditionName": "Stunting (height)",
+    "gradeScale": "BINARY",
+    "referralRequiredDefault": true,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "BAD_OBSTETRIC_HISTORY",
+    "conditionName": "Bad Obstetric History",
+    "gradeScale": "BINARY",
+    "referralRequiredDefault": true,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "ANEMIA",
+    "conditionName": "Anemia",
+    "gradeScale": "NORMAL_MILD_MODERATE_SEVERE",
+    "referralRequiredDefault": true,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "HYPERTENSION",
+    "conditionName": "Hypertension",
+    "gradeScale": "NORMAL_MILD_MODERATE_SEVERE",
+    "referralRequiredDefault": true,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "HYPOTENSION",
+    "conditionName": "Hypotension",
+    "gradeScale": "BINARY",
+    "referralRequiredDefault": false,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "HYPERGLYCEMIA",
+    "conditionName": "Hyperglycemia",
+    "gradeScale": "BINARY",
+    "referralRequiredDefault": true,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "HYPOGLYCEMIA",
+    "conditionName": "Hypoglycemia",
+    "gradeScale": "BINARY",
+    "referralRequiredDefault": false,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "APH",
+    "conditionName": "Antepartum Hemorrhage (APH)",
+    "gradeScale": "BINARY",
+    "referralRequiredDefault": true,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "PPH",
+    "conditionName": "Postpartum Hemorrhage (PPH)",
+    "gradeScale": "BINARY",
+    "referralRequiredDefault": false,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "HYPOTHERMIA",
+    "conditionName": "Hypothermia",
+    "gradeScale": "BINARY",
+    "referralRequiredDefault": false,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "HYPERTHERMIA",
+    "conditionName": "Hyperthermia",
+    "gradeScale": "BINARY",
+    "referralRequiredDefault": true,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "FETAL_HEART_RATE",
+    "conditionName": "Fetal Heart Rate (abnormal)",
+    "gradeScale": "BINARY",
+    "referralRequiredDefault": true,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "FUNDAL_HEIGHT",
+    "conditionName": "Fundal Height (deviation)",
+    "gradeScale": "BINARY",
+    "referralRequiredDefault": true,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "GESTATIONAL_WEIGHT_GAIN",
+    "conditionName": "Poor Gestational Weight Gain",
+    "gradeScale": "BINARY",
+    "referralRequiredDefault": true,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "JAUNDICE",
+    "conditionName": "Jaundice",
+    "gradeScale": "BINARY",
+    "referralRequiredDefault": true,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "URINE_ANALYSIS",
+    "conditionName": "Urine Analysis (abnormal)",
+    "gradeScale": "BINARY",
+    "referralRequiredDefault": true,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "DANGER_SIGNS",
+    "conditionName": "Danger Signs",
+    "gradeScale": "BINARY",
+    "referralRequiredDefault": true,
+    "educationRequiredDefault": true
+  }
+]
--- a/apps/risk-referral-service/prisma/seed.ts
+++ b/apps/risk-referral-service/prisma/seed.ts
@@ -1,6 +1,8 @@
 import { PrismaClient } from '../../../node_modules/.prisma/client-risk-referral-service';
 import selfReportedConditions from './seed-data/self-reported-conditions.json';
 import riskParameters from './seed-data/risk-parameters.json';
+import ancRiskConditions from './seed-data/anc-risk-conditions.json';
+import infantRiskConditions from './seed-data/infant-risk-conditions.json';
 
 const prisma = new PrismaClient();
 
@@ -21,6 +23,14 @@ interface RiskParameterSeed {
   entityType: 'MOTHER' | 'CHILD';
   unit: string | null;
   dataType: 'NUMERIC' | 'BOOLEAN' | 'CATEGORICAL';
+}
+
+interface AncRiskConditionSeed {
+  conditionCode: string;
+  conditionName: string;
+  gradeScale: 'BINARY' | 'NORMAL_MILD_MODERATE_SEVERE' | 'NORMAL_LOW_MEDIUM_HIGH';
+  referralRequiredDefault: boolean;
+  educationRequiredDefault: boolean;
 }
 
 /**
@@ -116,8 +126,107 @@ async function seedRiskParameters(): Promise<SeedResult> {
   };
 }
 
+/**
+ * Master data for the 18 ANC High-Risk conditions defined in Appendix D
+ * ("High risk protocols_Developer's copy - ANC HR", see
+ * docs/Appendix_D_High_Risk_Detection_Rules.md, Part 1). `phase: ANC` — these
+ * feed the anc-risk.rulesJson.ts RISK rule pack via a conditionCode ->
+ * risk_condition_id map that riskAssessment.service.ts builds by looking
+ * these rows up before calling evaluateRuleSet.
+ */
+async function seedAncRiskConditions(): Promise<SeedResult> {
+  let createdCount = 0;
+
+  for (const condition of ancRiskConditions as AncRiskConditionSeed[]) {
+    const existing = await prisma.riskCondition.findUnique({
+      where: { conditionCode: condition.conditionCode },
+    });
+    if (existing) continue;
+
+    await prisma.riskCondition.create({
+      data: {
+        conditionCode: condition.conditionCode,
+        conditionName: condition.conditionName,
+        entityType: 'MOTHER',
+        phase: 'ANC',
+        gradeScale: condition.gradeScale,
+        referralRequiredDefault: condition.referralRequiredDefault,
+        educationRequiredDefault: condition.educationRequiredDefault,
+        status: 'ACTIVE',
+      },
+    });
+    createdCount += 1;
+  }
+
+  if (createdCount === 0) {
+    return {
+      step: 'anc-risk-conditions',
+      created: false,
+      message: 'All ANC risk condition rows already present — skipped.',
+    };
+  }
+  return {
+    step: 'anc-risk-conditions',
+    created: true,
+    message: `Seeded ${createdCount} ANC risk condition row(s).`,
+  };
+}
+
+/**
+ * Master data for the 10 Infant High-Risk conditions defined in Appendix D
+ * ("High risk protocols_Developer's copy - Infant HR", see
+ * docs/Appendix_D_High_Risk_Detection_Rules.md, Part 2). Seeded once under
+ * `phase: INC` — condition_code has a global UNIQUE constraint (not
+ * unique-per-phase), and per SRS §3A.2.4 CCV reuses INC's thresholds
+ * verbatim, so both CCV_VISIT and NN (NEONATAL_VISIT) evaluations pass
+ * riskPhase: 'INC' to reuse this same seeded set rather than duplicating
+ * rows per phase (see infant-risk.rulesJson.ts's own doc comment).
+ */
+async function seedInfantRiskConditions(): Promise<SeedResult> {
+  let createdCount = 0;
+
+  for (const condition of infantRiskConditions as AncRiskConditionSeed[]) {
+    const existing = await prisma.riskCondition.findUnique({
+      where: { conditionCode: condition.conditionCode },
+    });
+    if (existing) continue;
+
+    await prisma.riskCondition.create({
+      data: {
+        conditionCode: condition.conditionCode,
+        conditionName: condition.conditionName,
+        entityType: 'CHILD',
+        phase: 'INC',
+        gradeScale: condition.gradeScale,
+        referralRequiredDefault: condition.referralRequiredDefault,
+        educationRequiredDefault: condition.educationRequiredDefault,
+        status: 'ACTIVE',
+      },
+    });
+    createdCount += 1;
+  }
+
+  if (createdCount === 0) {
+    return {
+      step: 'infant-risk-conditions',
+      created: false,
+      message: 'All infant risk condition rows already present — skipped.',
+    };
+  }
+  return {
+    step: 'infant-risk-conditions',
+    created: true,
+    message: `Seeded ${createdCount} infant risk condition row(s).`,
+  };
+}
+
 async function main(): Promise<void> {
-  const results = [await seedSelfReportedConditions(), await seedRiskParameters()];
+  const results = [
+    await seedSelfReportedConditions(),
+    await seedRiskParameters(),
+    await seedAncRiskConditions(),
+    await seedInfantRiskConditions(),
+  ];
 
   console.log('\nSeed summary:');
   for (const r of results) {
--- a/apps/risk-referral-service/src/risk-assessments/riskAssessment.repository.ts
+++ b/apps/risk-referral-service/src/risk-assessments/riskAssessment.repository.ts
@@ -1,9 +1,13 @@
-import type { Prisma } from '../../../../node_modules/.prisma/client-risk-referral-service';
+import type {
+  Prisma,
+  RiskPhase,
+} from '../../../../node_modules/.prisma/client-risk-referral-service';
 import type { PrismaService } from '../prisma/prisma.service';
 
 export interface RiskFlagCreateData {
   riskConditionId: string;
   riskGradeLookupValueId: string;
+  gradeRank: number;
   observedValueJson: Prisma.InputJsonValue | null;
   isReferralTrigger: boolean;
   isEducationTrigger: boolean;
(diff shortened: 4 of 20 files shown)


apps/visit-form-service/src/forms/form.service.ts
```diff
@@ -297,13 +319,21 @@ export class FormService {
...
+            }
+          : dto.formData;
       await triggerRiskAssessment(
         {
           beneficiaryId: dto.beneficiaryId,
           visitId: dto.visitId,
           submissionId: created.id,
           ruleSetId: version.formDefinition.riskRuleSetId,
-          answers: dto.formData,
+          riskPhase: FORM_CODE_TO_RISK_PHASE[formCode] ?? formCode,
```
logeshn-1, line 335: **FORM_CODE_TO_RISK_PHASE falls back to the raw formCode for any form not in its 4-entry map, but risk-referral-service's riskPhaseSchema is a strict zod enum of exactly 7 values, so the fallback can never be a valid riskPhase.**

seed.ts seeds 14 formCodes total but the map covers only 4; DELIVERY_VISIT, POSTPARTUM_VISIT, MOTHER_REGISTRATION and CHILD_REGISTRATION all have valid RiskPhase counterparts (DELIVERY/PP/REGISTRATION) the map simply omits. schema.prisma documents riskRuleSetId as admin-set out-of-band (no admin endpoint yet, no code change needed) so any of these forms can get a ruleSetId at any time. The moment one does, triggerRiskAssessment posts riskPhase:'DELIVERY_VISIT' (etc.), risk-referral-service's .strict() enum rejects it with 400, and triggerRiskAssessment only console.warns on !res.ok — so that form's entire risk-grading pipeline silently and permanently stops working, with no retry and no persisted failure record. Zero tests in form.service.spec.ts assert the riskPhase value for any formCode, mapped or not.

apps/rules-service/src/rules/scheduling/infant-risk.rulesJson.ts
```diff
@@ -0,0 +1,248 @@
...
+      hrVisitTrigger: grade !== 'NORMAL',
+    });
+  }
+
+  // --- Hypothermia / Hyperthermia (INFANT_VISIT/INC_VISIT/CCV_VISIT only -
+  // no temperature field on NEONATAL_VISIT). See doc-comment default #1. ---
+  const bodyTempF = input.child_temprature_in_f;
+  if (typeof bodyTempF === 'number') {
+    const hypoGrade = bodyTempF < 96 ? 'SEVERE' : 'NORMAL';
+    record('INFANT_HYPOTHERMIA', hypoGrade, { bodyTempF }, {
```
logeshn-1, line 154: **Infant Hypothermia/Hyperthermia (and Danger Signs at line 210) set isHrVisitTrigger to fire on every non-normal grade, never consulting the isFirstInstance map the handler already receives, even though Appendix D Part 2 explicitly requires 'Single instance' HR-visit triggering for these three conditions (vs 'every instance till normal' for the nutrition/LBW conditions right above them).**

A neonate stays hypothermic (<96F) across visit 1 and visit 2; both visits fire isHrVisitTrigger:true, stacking a second 15-day HR follow-up when spec calls for only one. More broadly, FORM_CODE_TO_RISK_PHASE (form.service.ts) maps both INC_VISIT and CCV_VISIT to the same riskPhase 'INC', so hr.rulesJson.ts's own cumulative:phase!=='CCV' scheduling logic can never see phase==='CCV' for a real CCV visit either -- CCV's mandated 30-day single-instance HR-visit cadence (SRS Section 3A.2.4/FR-S-5.3) is architecturally unreachable for any condition, not just these three. infant-risk.rulesJson.spec.ts never passes isFirstInstance for INFANT_HYPOTHERMIA/INFANT_HYPERTHERMIA/INFANT_DANGER_SIGNS, so this is untested.

apps/rules-service/src/rules/scheduling/anc-risk.rulesJson.ts
```diff
@@ -0,0 +1,392 @@
...
+  // sibling gates, since two accompanied-only conditions co-occurring with
+  // nothing else are not "accompanied by" each other's un-triggered state).
+  const nonNormalCodes = new Set(
+    results.filter((r) => r.grade !== 'NORMAL').map((r) => r.riskConditionId),
+  );
+  function patchAccompaniedTrigger(code, ownFlagged) {
+    if (!ownFlagged) return;
+    const entry = results.find((r) => r.riskConditionId === conditionIds[code]);
+    if (!entry) return;
+    const otherNonNormalExists = [...nonNormalCodes].some((id) => id !== entry.riskConditionId);
```
logeshn-1, line 369: **patchAccompaniedTrigger's own doc comment (lines 360-361) says the accompanied-by check should exclude 'the gate's own condition and the sibling gates', but otherNonNormalExists only excludes the entry's own id -- HYPOTENSION and HYPOGLYCEMIA (or HYPOTHERMIA) co-occurring alone wrongly trigger each other.**

BP 85 systolic (Hypotension MILD) plus blood glucose 60 mg/dL (Hypoglycemia MILD) with everything else NORMAL: nonNormalCodes contains both ids, so patchAccompaniedTrigger('HYPOTENSION', true) computes otherNonNormalExists=true (HYPOGLYCEMIA's id differs from HYPOTENSION's) and fires its referral/HR-visit trigger, and symmetrically for HYPOGLYCEMIA -- directly contradicting the comment's stated exclusion rule. anc-risk.rulesJson.spec.ts only tests Hypotension accompanied by a non-sibling condition (Anemia), so this exact scenario is untested. (Note: Appendix D section 1.4's literal text is just 'any other flagged high-risk condition' with no stated sibling exclusion, so this may be a documentation bug rather than a clinical one -- but the code and its own comment disagree, and a future maintainer 'fixing' the code to match the comment would introduce a real regression.)

apps/risk-referral-service/prisma/migrations/20260819000000_add_risk_flag_grade_rank/migration.sql
```diff
@@ -0,0 +1,10 @@
+-- AlterTable
+-- Add as nullable first, backfill existing rows to 0 (treated as NORMAL,
+-- the safest default for historic flags whose true grade rank was never
+-- recorded), then enforce NOT NULL — avoids failing on any pre-existing
+-- risk_flags rows.
+ALTER TABLE "risk_flags" ADD COLUMN "grade_rank" INTEGER;
+
+UPDATE "risk_flags" SET "grade_rank" = 0 WHERE "grade_rank" IS NULL;
```
logeshn-1, line 8: **The backfill sets grade_rank=0 (NORMAL) for every pre-existing risk_flags row unconditionally, regardless of that row's true historic grade (stored via riskGradeLookupValueId), even though the RiskAssessment/RiskFlag write pipeline already existed and was live before this migration.**

Any risk_flags row written before 2026-08-19 with a real MILD/MODERATE/SEVERE grade now reads grade_rank=0 going forward. schema.prisma's own comment says gradeRank exists so 'a beneficiary's grade trend can be compared across visits (improvement/deterioration)' and names findConsecutiveNoImprovementCount and the Part 3 dashboard's progression/deterioration reports as consumers -- both would show a false 'improved to NORMAL' for a beneficiary whose historic flag was actually SEVERE, and the true value is not recoverable from grade_rank itself.

apps/visit-form-service/src/forms/form.service.ts
```diff
@@ -338,6 +368,58 @@ export class FormService {
...
+
+    const answers = registration.formDataJson as Record<string, unknown>;
+    const age = answers.age_of_the_beneficiary;
+    const gravida = answers.gravida_total_number_of_pregnancies;
+    const livingChildren = answers.living_children;
+    const abortions = answers.abortions_pregnancy_losses_before_24_weeks;
+    const complications =
+      answers.did_you_experience_any_complications_during_birth_delivery_in_previous_pregnancies;
+
+    const badObstetricHistoryFlag =
```
logeshn-1, line 409: **resolveAncRiskRegistrationAnswers computes the Bad Obstetric History business threshold (gravida>4 OR livingChildren<gravida OR abortions>=2 OR non-trivial complications) directly in visit-form-service TypeScript, violating this repo's own .claude/CLAUDE.md rule: 'What must NEVER be added: ... business thresholds/rates (those live in GoRules).'**

Changing Appendix D's Bad Obstetric History band now requires editing two services in lockstep (this threshold math here, plus the grade/trigger wiring in anc-risk.rulesJson.ts) instead of one, and the two can silently drift -- anc-risk.rulesJson.spec.ts's threshold-focused test suite never exercises this method's math, so a wrong cutoff here ships without failing any of the rule-pack's own threshold tests.

apps/risk-referral-service/src/risk-assessments/riskAssessment.repository.ts
```diff
@@ -47,6 +51,86 @@ export class RiskAssessmentRepository {
...
+   * moment gradeRank decreases from one flag to the next-most-recent one;
+   * a flat or worsening trend keeps counting. Capped at 3 visits back since
+   * the rule pack only needs to know ">=3", not the exact streak length.
+   */
+  async findConsecutiveNoImprovementCount(
+    beneficiaryId: string,
+    conditionCodes: string[],
+  ): Promise<Map<string, number>> {
+    const result = new Map<string, number>();
+    for (const conditionCode of conditionCodes) {
```
logeshn-1, line 114: **findConsecutiveNoImprovementCount loops over every conditionCode and issues one separate findMany query per code inside a for-loop, violating .claude/CLAUDE.md Section 11's 'Avoid N+1 -- use include/batched queries', and is called unconditionally even though anc-risk.rulesJson.ts's handler never reads the result.**

Every ANC_VISIT submission (18 seeded ACTIVE conditions) triggers 18 sequential DB round trips in riskAssessment.service.ts:116 to build a map that anc-risk.rulesJson.ts destructures but never uses (only infant-risk.rulesJson.ts's 4 nutrition conditions read consecutiveNoImprovementCount) -- real, unconditional added latency on the hot submission path that scales linearly as more conditions get seeded per phase.

apps/rules-service/src/rules/scheduling/anc-risk.rulesJson.ts
```diff
@@ -0,0 +1,392 @@
+/**
```
logeshn-1, line 1: **This new file is 392 lines, exceeding .claude/CLAUDE.md Section 3's 'Files <= ~250 lines -- split by responsibility when larger' by 57%.**

The single largest and most clinically dense file in the PR (19 conditions, first-instance gating, accompanied-by patching) has no split points, making it harder to review and more likely to accumulate exactly the kind of copy-paste divergence that produced the dead hrGateByFirstInstance option and the patchAccompaniedTrigger comment/code mismatch flagged elsewhere in this review.

apps/rules-service/src/rules/scheduling/anc-risk.rulesJson.ts
```diff
@@ -0,0 +1,392 @@
...
+    const isFirst = firstInstance[code] !== false; // default true if unknown
+    const referralEligible = opts.referralTrigger === true;
+    const hrEligible = opts.hrVisitTrigger === true;
+    results.push({
+      riskConditionId: requireConditionId(code),
+      grade,
+      gradeRank: GRADE_RANK[grade],
+      isReferralTrigger: gateByFirstInstance ? referralEligible && isFirst : referralEligible,
+      isEducationTrigger: grade !== 'NORMAL',
+      isHrVisitTrigger: opts.hrGateByFirstInstance
```
logeshn-1, line 109: **record()'s hrGateByFirstInstance option is dead code -- the only call site that ever sets it (GESTATIONAL_WEIGHT_GAIN, line 315) sets it to false, which is already the default, so the 'hrEligible && isFirst' branch never executes across all 19 conditions.**

Doubles the gating concepts a reader must track (onlyFirstInstance for referral vs hrGateByFirstInstance for HR-visit) for zero behavioral payoff. A future maintainer needing a genuinely first-instance-gated HR trigger has no working example to copy from, since the true branch has never actually run.

apps/risk-referral-service/prisma/seed.ts
```diff
@@ -116,8 +126,107 @@ async function seedRiskParameters(): Promise<SeedResult> {
...
 
+/**
+ * Master data for the 18 ANC High-Risk conditions defined in Appendix D
+ * ("High risk protocols_Developer's copy - ANC HR", see
+ * docs/Appendix_D_High_Risk_Detection_Rules.md, Part 1). `phase: ANC` — these
+ * feed the anc-risk.rulesJson.ts RISK rule pack via a conditionCode ->
+ * risk_condition_id map that riskAssessment.service.ts builds by looking
+ * these rows up before calling evaluateRuleSet.
+ */
+async function seedAncRiskConditions(): Promise<SeedResult> {
```
logeshn-1, line 137: **seedAncRiskConditions (line 137) and seedInfantRiskConditions (line 185) are near-duplicate ~35-line functions differing only in the imported JSON, hardcoded entityType, hardcoded phase, and step name -- every other line (existence check, create() shape, result branches) is copy-pasted verbatim from each other and from the pre-existing seedSelfReportedConditions.**

Any future change to seeding behavior (switching to upsert, adding logging, a new required field) must be made in 2-3 places in lockstep or silently drifts -- a single seedRiskConditions(items, entityType, phase, stepName) helper would remove the duplication entirely.

apps/visit-form-service/src/forms/form.service.spec.ts
```diff
@@ -1320,6 +1323,227 @@ describe('FormService', () => {
...
+          },
+          'u1',
+          'Bearer test-token',
+        );
+
+        expect(repository.findLatestSubmissionByBeneficiaryAndFormCode).toHaveBeenCalledWith(
+          'b1',
+          'MOTHER_REGISTRATION',
+        );
+        expect(jest.mocked(triggerRiskAssessment)).toHaveBeenCalledWith(
```
logeshn-1, line 1379: **Every triggerRiskAssessment assertion in this spec file uses expect.objectContaining, and none of them ever assert the riskPhase field's value for any formCode.**

The FORM_CODE_TO_RISK_PHASE fallback bug (invalid riskPhase for unmapped forms) would pass this entire test suite silently, including the CHILD_REGISTRATION test at line 1524 which configures exactly the vulnerable scenario (a riskRuleSetId on an unmapped formCode) without checking what riskPhase actually gets sent.

apps/risk-referral-service/prisma/seed-data/anc-risk-conditions.json
```diff
@@ -0,0 +1,135 @@
...
+  },
+  {
+    "conditionCode": "HYPERTENSION",
+    "conditionName": "Hypertension",
+    "gradeScale": "NORMAL_MILD_MODERATE_SEVERE",
+    "referralRequiredDefault": true,
+    "educationRequiredDefault": true
+  },
+  {
+    "conditionCode": "HYPOTENSION",
```
logeshn-1, line 45: **Seed data declares referralRequiredDefault/educationRequiredDefault per condition (e.g. HYPOTENSION: referralRequiredDefault:false), served back out via GET /risk-conditions, but the grading pipeline (anc-risk.rulesJson.ts) never reads these fields -- referral/education/HR-visit logic is hardcoded per-condition in JS instead, creating two independent sources of truth for the same business rule.**

HYPOTENSION's seed value happens to match anc-risk.rulesJson.ts's hardcoded referralTrigger:false today, but nothing enforces that agreement. A future engineer 'fixing' HYPOTENSION's referral behavior by editing this seed JSON -- the only place its own schema documents these fields -- gets zero behavior change, since only the JS literal inside the rule pack's record() call actually governs triggering.

apps/risk-referral-service/src/risk-assessments/riskAssessment.service.ts
```diff
@@ -94,7 +94,38 @@ export class RiskAssessmentService {
...
+    // Resolve conditionCode -> risk_condition_id (a rule pack's decision
+    // graph output must carry a real DB id, see ruleSet.evaluator.ts's
+    // RiskEvaluationResult contract, but the pack itself only knows portable
+    // condition codes), the "only first instance" flagging history, and the
+    // "3 consecutive visits with no improvement" streak (infant nutrition
+    // conditions, Appendix D §2.4) — all merged into `answers` so the rule
+    // pack sees them as ordinary evaluation inputs (see
+    // anc-risk.rulesJson.ts / infant-risk.rulesJson.ts's `conditionIds`/
+    // `isFirstInstance`/`consecutiveNoImprovementCount`).
+    const conditionIdsByCode = await this.repository.findConditionIdsByPhase(dto.riskPhase);
```
logeshn-1, line 106: **findEverFlaggedConditionCodes(dto.beneficiaryId) doesn't depend on conditionIdsByCode/conditionCodes at all, yet it's sequenced after `await this.repository.findConditionIdsByPhase(dto.riskPhase)` completes instead of running concurrently with it via Promise.all.**

Every risk assessment creation pays the full latency of findConditionIdsByPhase before even starting findEverFlaggedConditionCodes's independent DB round trip, when only findConsecutiveNoImprovementCount (which genuinely needs conditionCodes) requires waiting -- restructuring to run findConditionIdsByPhase and findEverFlaggedConditionCodes concurrently first would shave one round trip off the critical path of every submission.

apps/risk-referral-service/prisma/seed.ts
```diff
@@ -116,8 +126,107 @@ async function seedRiskParameters(): Promise<SeedResult> {
   };
 }
 
+/**
+ * Master data for the 18 ANC High-Risk conditions defined in Appendix D
```
logeshn-1, line 130: **Doc comments miscount the seeded conditions: line 130 says 'the 18 ANC High-Risk conditions' but anc-risk-conditions.json has 19 entries, and line 176 says 'the 10 Infant High-Risk conditions' but infant-risk-conditions.json has 12 entries.**

Minor, but a future maintainer trusting the comment's count when auditing seeded conditions against Appendix D would undercount by 1 (ANC) or 2 (Infant), potentially missing that a condition was added without the comment being updated.

apps/visit-form-service/src/forms/form.service.ts
Dharanishwaran-ND: Fixed. Added the 4 missing entries (MOTHER_REGISTRATION/CHILD_REGISTRATION -> REGISTRATION, DELIVERY_VISIT -> DELIVERY, POSTPARTUM_VISIT -> PP) and removed the `?? formCode` fallback entirely — an unmapped formCode with a riskRuleSetId now throws a loud, named error at submission time instead of silently posting an invalid riskPhase. Added a dedicated test block (`FORM_CODE_TO_RISK_PHASE mapping (PR #172 review)`) covering every mapped formCode plus the unmapped-throws case, including the exact CHILD_REGISTRATION scenario you called out.

apps/rules-service/src/rules/scheduling/infant-risk.rulesJson.ts
Dharanishwaran-ND: Fixed. INFANT_HYPOTHERMIA/INFANT_HYPERTHERMIA/INFANT_DANGER_SIGNS now gate isHrVisitTrigger on isFirstInstance via a new record() option (hrGateByFirstInstance) — a neonate hypothermic across two visits fires the 15-day HR follow-up once, not on every visit. Added dedicated tests for all three.

On the broader CCV-cadence point: investigated further and the gap is bigger than the phase-collapse issue you flagged — hr.rulesJson.ts (which has the actual phase!=='CCV' cumulative-vs-single-instance scheduling logic) isn't called anywhere in the real submission write path today. riskAssessment.service.ts only grades conditions and computes a hrDetectedFlag boolean; nothing invokes hr.rulesJson.ts to generate/schedule the actual HR visit. So this is a missing integration, not a regression in wired-up code — filed as #173 rather than expanding this PR's scope.

apps/rules-service/src/rules/scheduling/anc-risk.rulesJson.ts
Dharanishwaran-ND: Fixed exactly as described — patchAccompaniedTrigger now excludes both the gate's own condition AND the sibling gates (HYPOTENSION/HYPOGLYCEMIA/HYPOTHERMIA) via an accompaniedGateIds set, matching the function's own doc comment. Added the exact regression case: BP 85 systolic + blood glucose 60 mg/dL with everything else NORMAL now correctly does not trigger either condition off the other.

apps/risk-referral-service/prisma/migrations/20260819000000_add_risk_flag_grade_rank/migration.sql
Dharanishwaran-ND: Agreed this needed a real fix, not just a comment. Since this migration is already applied (checksummed in _prisma_migrations), I left its SQL as historical record and added a corrective one-time script instead: prisma/backfill-grade-rank-from-lookup.ts, which re-derives each risk_flags row's true grade_rank from its riskGradeLookupValueId via auth-service's GET /lookups/RISK_GRADE (id -> NORMAL=0..SEVERE=3), correcting any row the blind 0-fill got wrong. Added a spec (backfill-grade-rank-from-lookup.spec.ts) covering corrected/already-correct/unresolved/empty-table cases. Added a doc comment on the migration itself pointing to the script. Confirmed risk_flags is empty on all real environments checked today, so no data has actually been lost yet — but the script is required before this migration is safe to run against any environment with pre-existing graded rows.

apps/visit-form-service/src/forms/form.service.ts
Dharanishwaran-ND: Fixed. Moved the G>4/L<P/abortions>=2/prior-complications threshold math into anc-risk.rulesJson.ts itself; form.service.ts's resolveAncRiskRegistrationAnswers now only fetches and passes through the raw registration fields (age/gravida/livingChildren/abortions/priorComplications) unchanged. Added a dedicated 'Bad Obstetric History threshold (moved into GoRules)' test block in anc-risk.rulesJson.spec.ts covering all four threshold branches plus the not-graded-at-all case, directly exercising this pack's own math instead of relying on the indirect gating test.

apps/risk-referral-service/src/risk-assessments/riskAssessment.repository.ts
Dharanishwaran-ND: Fixed. Rewrote findConsecutiveNoImprovementCount as one batched query (conditionCode: {in: conditionCodes}) instead of one query per code in a loop, grouping/trimming to 4-per-code in JS afterward. Also addressed the unconditional-call issue you raised in the other comment on this file together with this one: the service now only calls this method for NN/INC/CCV phases (the only rule pack that reads consecutiveNoImprovementCount), and runs it after findConditionIdsByPhase/findEverFlaggedConditionCodes resolve concurrently rather than blocking on an unrelated round trip.

apps/rules-service/src/rules/scheduling/anc-risk.rulesJson.ts
Dharanishwaran-ND: Not split in this pass — flagging rather than marking resolved. The other fixes in this review (moving BOH math in, removing the dead hrGateByFirstInstance option, fixing the accompanied-trigger bug) touched this file substantially, and I didn't want to interleave a structural split with those functional changes in the same pass. Splitting out a shared risk-grading.helpers.ts (danger-signs/multiselect helpers, patchAccompaniedTrigger) is still worth doing as a follow-up — happy to do it as a separate small PR once this one's functional fixes are settled, so the diff for each is easy to review independently.

apps/rules-service/src/rules/scheduling/anc-risk.rulesJson.ts
Dharanishwaran-ND: Fixed. Removed the hrGateByFirstInstance option and its one (always-false) call site entirely from anc-risk.rulesJson.ts. Note: infant-risk.rulesJson.ts now has a hrGateByFirstInstance option too (see the isFirstInstance-gating reply above) — that one has 3 real call sites (Hypothermia/Hyperthermia/Danger Signs) that actually exercise the true branch, so it's a live option there, not dead code.

apps/risk-referral-service/prisma/seed.ts
Dharanishwaran-ND: Fixed. Extracted seedRiskConditions(items, entityType, phase, stepName) — seedAncRiskConditions/seedInfantRiskConditions are now one-line callers. Also fixed the doc-comment count typos you didn't flag here but that were adjacent (18->19 ANC, 10->12 Infant) while rewriting these functions.

apps/visit-form-service/src/forms/form.service.spec.ts
Dharanishwaran-ND: Fixed. Added a dedicated 'FORM_CODE_TO_RISK_PHASE mapping (PR #172 review)' test block that asserts the actual riskPhase value (not objectContaining with the field omitted) for every mapped formCode, plus a negative test that an unmapped formCode throws rather than sending an invalid value. This directly covers the CHILD_REGISTRATION scenario at the line you cited.

apps/risk-referral-service/prisma/seed-data/anc-risk-conditions.json
Dharanishwaran-ND: You're right these are two independent sources of truth today, and after digging in I don't think a naive wire-up is safe: the seed fields are static per-condition booleans, but the real trigger logic in both rule packs is grade-/first-instance-/accompanied-by-dependent (e.g. Anemia only refers on MODERATE/SEVERE, Hypotension only refers when accompanied by another flagged condition) — a single static default can't express that without a real regression. Rather than leave the ambiguity implicit, I made the intent explicit: added doc comments on schema.prisma's RiskCondition model, the GET /risk-conditions route schema, the seed.ts interface, and both rule packs' own doc comments, all stating plainly that these fields are reference/display-only master data and are NOT read by the grading logic, which remains each rule pack's sole source of truth.

apps/risk-referral-service/src/risk-assessments/riskAssessment.service.ts
Dharanishwaran-ND: Fixed. findConditionIdsByPhase and findEverFlaggedConditionCodes now run concurrently via Promise.all — findConsecutiveNoImprovementCount is the only one that genuinely needs conditionCodes, so it now runs after that resolves, and only for NN/INC/CCV phases (see the N+1 fix reply on riskAssessment.repository.ts, done together with this one).

apps/risk-referral-service/prisma/seed.ts
Dharanishwaran-ND: Fixed as part of the seed.ts dedup — corrected to 19 ANC / 12 Infant while extracting the shared seedRiskConditions helper.

PR #175, opened 2026-08-20
feat: Beneficiary risk color, CHILD phase-advance, BR-13, last-visit vitals

### LINKED ISSUE #174: Beneficiary detail: risk color, CHILD phase-advance, BR-13 CCV opening risk state, last-visit vitals
## Background
`GET /beneficiaries/:id` was missing several fields needed by the UI and by the CHILD case
lifecycle per the SRS: a derived risk color for badges, CHILD phase-advance beyond NN,
BR-13's CCV opening risk state, and last-visit vitals.

## Scope
- Add `riskColor` (GREEN/YELLOW/RED) derived from `riskLevel` on the beneficiary detail response.
- Extend CHILD case phase-advance to cover NN->INC and INC->CCV, keeping
  `ChildCaseDetails.currentPhase` in sync with `BeneficiaryCase.currentPhase`.
- Implement BR-13: compute and persist `ccvOpeningRiskState` exactly once, at the INC->CCV
  transition, from the last 3 completed INC visits plus a full 0-12-month high-risk scan.
- Add a last-visit-vitals endpoint (weight, BP, temperature, hemoglobin, MUAC, respiratory
  rate) and wire it into the beneficiary detail response.

## Notes
- `riskAssessment.service.listByVisitIds` and `form.service.getLatestVisitVitals` needed new
  ownership checks (SAKHI-own-case / SUPERVISOR-own-roster / MANAGER-ADMIN-unrestricted) — an
  automated security review flagged both as IDOR without them.
- A cross-service infinite request loop was found and fixed during implementation: the vitals
  ownership check must use a new minimal `GET /beneficiaries/:id/ownership` endpoint, not the
  full enriched detail endpoint (which itself calls back into vitals).
- `api-gateway` had no route for the new `latest-visit-vitals` endpoint — added.

## Test plan
- [ ] `beneficiary.service.spec.ts` / `beneficiary.repository.spec.ts` pass (riskColor, phase sync, ownership)
- [ ] `riskAssessment.service.spec.ts` passes (listByVisitIds + IDOR guards)
- [ ] `form.service.spec.ts` / `vitalsExtractor.spec.ts` pass (BR-13 gating, vitals extraction)
- [ ] Live-verified against running dev services: riskColor, CHILD phase-advance, BR-13 end-to-end
  (INC visits -> CCV_VISIT submission -> `ccvOpeningRiskState` populated), and
  `GET /beneficiaries/:id/latest-visit-vitals` through the gateway

### LINKED ISSUE #176: Add Pre-Visit Health History read API (FR-S-4.6)
## Requirement
FR-S-4.6: the Sakhi mobile app needs to show a beneficiary's last 2 completed visits' key vitals
(Hemoglobin, Blood Pressure, Weight, Blood Sugar, Temperature) before she starts her next visit.

## Scope
- Investigate and fix gaps blocking this endpoint:
  - `form_answers` population on submission (confirmed already implemented — no fix needed)
  - Visit completion write path (real gap — `VisitInstance.completedAt` was never set from form
    submission; fixed via a best-effort resolver wired into `createSubmission`)
  - Beneficiary scoping on visit-history reads (confirmed already correct elsewhere; new endpoint
    copies the existing `getLatestVisitVitals` ownership-check pattern)
- Add `GET /beneficiaries/:beneficiaryId/visit-history` (visit-form-service), gateway-routed,
  Swagger-documented, scoped to the caller's own roster.
- Extend `vitalsExtractor.ts` with a unit-wrapped response shape and a `bloodSugarMgDl` field.

## Live-verification finding
Found and fixed a pre-existing bug in `findBeneficiaryOwnership` (`beneficiary.client.ts`): a
caller outside their roster got a 502 instead of a 403, because the client only special-cased 404
and treated the ownership endpoint's own legitimate 403 as an upstream failure. This affected both
the new endpoint and the existing `GET /beneficiaries/:beneficiaryId/latest-visit-vitals`.

Dharanishwaran-ND:
Closes #174

## Summary
- Adds `riskColor` (GREEN/YELLOW/RED) derived from `riskLevel` on the beneficiary detail response
- Extends CHILD case phase-advance to NN->INC and INC->CCV, keeping `ChildCaseDetails.currentPhase`
  in sync with `BeneficiaryCase.currentPhase` via an atomic transaction
- Implements BR-13: computes and persists `ccvOpeningRiskState` exactly once at the INC->CCV
  transition, from the last 3 completed INC visits plus a full 0-12-month high-risk detection scan,
  evaluated through the seeded CCV scheduling rule pack
- Adds `GET /beneficiaries/:beneficiaryId/latest-visit-vitals` (visit-form-service), wired into the
  beneficiary detail response, with a per-formCode question_code mapping for weight/BP/temperature/
  hemoglobin/MUAC/respiratory-rate
- Adds `GET /risk-assessments` (risk-referral-service) so visit-form-service's BR-13 resolver can
  look up risk assessments by visit id without a cross-service DB join
- Adds `GET /beneficiaries/:id/ownership` (beneficiary-service), a minimal ownership-only lookup
  used by the vitals endpoint's ownership check to avoid a cross-service infinite request loop
  that calling the full enriched detail endpoint would otherwise cause
- Fixes an api-gateway routing gap: `latest-visit-vitals` previously 404'd, falling through to the
  generic `/beneficiaries` prefix

## Security
An automated security review flagged 2 IDOR issues introduced mid-implementation
(`RiskAssessmentService.listByVisitIds`, `FormService.getLatestVisitVitals` both took a
caller-supplied `beneficiaryId` with no ownership check) — both fixed with the standard
SAKHI-own-case / SUPERVISOR-own-roster / MANAGER-ADMIN-unrestricted check used elsewhere in
these services.

## Test plan
- [x] `beneficiary.service.spec.ts` / `beneficiary.repository.spec.ts` — riskColor, CHILD phase
  sync, ownership endpoint, vitals wiring
- [x] `riskAssessment.service.spec.ts` — `listByVisitIds` + IDOR guards
- [x] `form.service.spec.ts` / `vitalsExtractor.spec.ts` / `ccvOpeningRiskState.resolver` coverage —
  BR-13 gating (only fires on the actual INC->CCV transition, not repeat CCV_VISIT submissions),
  vitals extraction per form code
- [x] Live-verified end-to-end against the running dev stack: riskColor, CHILD phase-advance,
  BR-13 (real INC visits -> CCV_VISIT submission -> `ccvOpeningRiskState` populated), and
  `GET /beneficiaries/:id/latest-visit-vitals` through the gateway

## Env vars
- `beneficiary-service`: `AUTH_SERVICE_BASE_URL`, `RISK_REFERRAL_SERVICE_BASE_URL`,
  `VISIT_FORM_SERVICE_BASE_URL` (all optional, gateway-routed, default `http://localhost:3000`)
- `visit-form-service`: `CCV_SCHEDULE_RULE_SET_ID` (optional — BR-13 computation is skipped, not
  failed, when unset; look up via `GET /rules` with an ADMIN token)

All new/changed env vars are documented in the respective `.env.example` files.

---

## Update: Pre-visit health history API (FR-S-4.6)

Closes #176

- Adds `GET /beneficiaries/:beneficiaryId/visit-history` (visit-form-service) — a beneficiary's
  last N (default 2) **completed** visits' key vitals (hemoglobin, blood pressure, weight, blood
  sugar, temperature), scoped to the caller's own roster same as `latest-visit-vitals`. Supports
  `formCode`/`visitType` (repeatable) and `limit` query params. Gateway-routed, Swagger-documented.
- Wires visit auto-completion into `createSubmission`: a visit-linked submission against a
  vitals-bearing form (ANC/PP/NN/INC/CCV) now sets `VisitInstance.completedAt` automatically,
  best-effort, instead of requiring a separate `PATCH /visits/:id` call. Confirmed via
  investigation that `VisitSchedule.status` (a separate, unrelated field) is dead code — nothing
  in the codebase writes to it; the real completion signal is `VisitInstance.completedAt`.
- Extends `vitalsExtractor.ts` with a `bloodSugarMgDl` field and a new unit-wrapped extractor
  (`extractVisitHistoryVitals`) reusing the existing per-formCode question_code mapping — no new
  mapping introduced. Temperature is reported in °F (as every seeded form actually stores it), not
  converted to °C.

### Live-verification finding (fixed in this PR)
Found via manual end-to-end testing against the running dev stack: `findBeneficiaryOwnership`
(`beneficiary.client.ts`) only special-cased a `404` from `GET /beneficiaries/:id/ownership` —
any other non-2xx response, including that endpoint's own legitimate `403` for an out-of-roster
caller, was surfaced as a `502` instead. This silently broke the scoping-rejection path on **both**
this new endpoint and the existing `latest-visit-vitals` endpoint. Fixed by forwarding the 403 with
its upstream message; added test coverage in `beneficiary.client.spec.ts`.

### Test plan (this update)
- [x] `visitCompletion.resolver.spec.ts` — auto-completion logic, all 5 form codes, idempotent
  no-op on an already-COMPLETED visit, never throws on a downstream failure
- [x] `form.service.spec.ts` — completion call wired into `createSubmission`, doesn't block the
  submission
- [x] `vitalsExtractor.spec.ts` — new `bloodSugarMgDl` field, unit-wrapped shape, null-with-unit
  behavior for an uncaptured vital
- [x] `visitInstance.repository.spec.ts` / `visitInstance.service.spec.ts` — empty/1/3+ completed
  visits (limit truncation), formCode/visitType filtering, full SAKHI/SUPERVISOR/MANAGER/ADMIN
  scoping matrix, not-found
- [x] `visit-history-query.dto.spec.ts` — limit default/coercion/bounds, repeatable params,
  `.strict()` rejection
- [x] `beneficiary.client.spec.ts` — 403-forwarding fix
- [x] Live-verified end-to-end against the running dev stack with a throwaway Sakhi/beneficiary:
  empty visits, submission → auto-complete → visit-history reflecting real vitals, 3 completed
  visits → only last 2 returned, a visit with no captured vitals → null with unit, formCode/
  visitType filters, 400/401/403/404 error paths, and the `latest-visit-vitals` 403 fix

---

## Update: Fix immunization date fields showing after selecting "No"

Every immunization Yes/No question in INC_VISIT/INFANT_VISIT/CCV_VISIT's shared History section
(22 vaccine/date pairs — BCG through DPT booster1) had its date field's `visibleWhen` tied only to
the outer "have you been able to meet the beneficiary for the visit?" question, not to its own
preceding Yes/No answer. Selecting "No" on e.g. BCG still showed the BCG date field, since the
schema never told it to hide.

- Fixed `infant-visit.json` (the shared seed source for all 3 form codes) — each date field's
  `visibleWhen` now references its own preceding vaccine field (`{field: <vaccine_code>, operator:
  "eq", value: "yes"}`) as a full replacement, not an "and" condition: the vaccine field is itself
  already gated behind the outer question, so an unanswered/hidden governing field already
  evaluates to hidden downstream — no visibility is lost by dropping the outer reference.
- Updated `seed-data.spec.ts` to assert the corrected per-vaccine gating instead of the old
  blanket outer-question assertion for these 22 date fields.
- Live-verified against the running dev stack: published a new version for all 3 form codes
  (`INFANT_VISIT`, `INC_VISIT`, `CCV_VISIT` — confirmed they still share identical schema content)
  via the admin create-draft → patch → publish flow, then fetched each `active-version` and
  field-by-field verified all 22 radio fields still gate on the outer question (unchanged) and all
  22 date fields now gate on their own vaccine field. Caught and corrected a transposed-mapping
  bug in my own verification script before it reached this PR's source file — the committed
  `infant-visit.json` was unaffected and independently re-verified correct.
- Does not fix the separate "OPV-0 date accepts future dates" issue — that's a mobile-app
  client-side date-bound rule, not schema-driven.

### Test plan (this update)
- [x] `seed-data.spec.ts` — all 22 vaccine/date pairs assert correct per-field gating; full suite
  (87 tests) passes
- [x] Full `visit-form-service` suite (519 tests) passes
- [x] Live-verified: all 3 form codes' active versions field-by-field checked for all 22 pairs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,13 +20,5 @@ jobs:
       - name: Derive affected base
         uses: nrwl/nx-set-shas@v4
       - run: npx nx affected -t lint --parallel=3
-      - name: Run tests
-        env:
-          # Unit tests mock Prisma and never touch a real database, but every
-          # service's app-config.ts requires DATABASE_URL/DIRECT_URL with no
-          # default (fail-fast boot check) — importing app.module.ts (even
-          # transitively, e.g. via a controller spec) throws without these.
-          DATABASE_URL: postgresql://user:password@localhost:5432/db
-          DIRECT_URL: postgresql://user:password@localhost:5432/db
-        run: npx nx affected -t test --parallel=3 --coverage --passWithNoTests
+      - run: npx nx affected -t test --parallel=3 --coverage --passWithNoTests
       - run: npx nx affected -t build --parallel=3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,5 +20,13 @@ jobs:
       - name: Derive affected base
         uses: nrwl/nx-set-shas@v4
       - run: npx nx affected -t lint --parallel=3
-      - run: npx nx affected -t test --parallel=3 --coverage --passWithNoTests
+      - name: Run tests
+        env:
+          # Unit tests mock Prisma and never touch a real database, but every
+          # service's app-config.ts requires DATABASE_URL/DIRECT_URL with no
+          # default (fail-fast boot check) — importing app.module.ts (even
+          # transitively, e.g. via a controller spec) throws without these.
+          DATABASE_URL: postgresql://user:password@localhost:5432/db
+          DIRECT_URL: postgresql://user:password@localhost:5432/db
+        run: npx nx affected -t test --parallel=3 --coverage --passWithNoTests
       - run: npx nx affected -t build --parallel=3
--- a/apps/api-gateway/src/proxy/routes.ts
+++ b/apps/api-gateway/src/proxy/routes.ts
@@ -117,6 +117,14 @@ export const SERVICE_ROUTES: readonly ServiceRoute[] = [
     target: appConfig.VISIT_FORM_SERVICE_URL,
     requiresAuth: true,
   },
+  // GET /beneficiaries/:beneficiaryId/visit-history (FR-S-4.6) — same
+  // ownership as latest-visit-vitals above: owned by visit-form-service
+  // (it owns visit_instances/form_submissions), not beneficiary-service.
+  {
+    prefix: '/beneficiaries/:beneficiaryId/visit-history',
+    target: appConfig.VISIT_FORM_SERVICE_URL,
+    requiresAuth: true,
+  },
   { prefix: '/beneficiaries', target: appConfig.BENEFICIARY_SERVICE_URL, requiresAuth: true },
   // UNCONFIRMED alias for GET /beneficiaries/risk-summary — see the route's
   // own doc comment in beneficiary.routes.ts for why this is a best guess.
--- a/apps/api-gateway/src/proxy/routes.ts
+++ b/apps/api-gateway/src/proxy/routes.ts
@@ -108,6 +108,15 @@ export const SERVICE_ROUTES: readonly ServiceRoute[] = [
     target: appConfig.VISIT_FORM_SERVICE_URL,
     requiresAuth: true,
   },
+  // GET /beneficiaries/:beneficiaryId/latest-visit-vitals — owned by
+  // visit-form-service (it owns form_submissions/visit_instances), not
+  // beneficiary-service, which calls this same route itself server-to-
+  // server (via visitVitals.client.ts) to enrich GET /beneficiaries/:id.
+  {
+    prefix: '/beneficiaries/:beneficiaryId/latest-visit-vitals',
+    target: appConfig.VISIT_FORM_SERVICE_URL,
+    requiresAuth: true,
+  },
   { prefix: '/beneficiaries', target: appConfig.BENEFICIARY_SERVICE_URL, requiresAuth: true },
   // UNCONFIRMED alias for GET /beneficiaries/risk-summary — see the route's
   // own doc comment in beneficiary.routes.ts for why this is a best guess.
--- a/apps/auth-service/src/auth/auth.repository.ts
+++ b/apps/auth-service/src/auth/auth.repository.ts
@@ -126,7 +126,16 @@ export class AuthRepository {
     return this.prisma.role.findUnique({ where: { roleCode } });
   }
 
-  /** Creates the user and their initial role assignment atomically. */
+  /**
+   * Creates the user and their initial role assignment atomically. When
+   * `sakhiProfile` is supplied (roleCode SAKHI — see AuthService.createUser),
+   * the sakhi_profiles row is created in the same transaction, so a SAKHI
+   * user can never exist without one: previously POST /users only wrote
+   * users/user_roles, leaving a newly-created SAKHI unable to be found by
+   * any endpoint that resolves identity via sakhi_profiles (e.g.
+   * GET /sakhi/:id/dashboard) until a follow-up PATCH supplied a
+   * profile field. Mirrors updateUserTransaction's `create` branch.
+   */
   createUserWithRole(data: {
     username: string;
     mobileNumber: string;
@@ -136,6 +145,7 @@ export class AuthRepository {
     roleId: string;
     projectId: string | null;
     geographyUnitId: string | null;
+    sakhiProfile?: { primaryProjectId: string; phoneNumber: string };
   }) {
     return this.prisma.$transaction(async (tx) => {
       const user = await tx.user.create({
@@ -156,6 +166,16 @@ export class AuthRepository {
           effectiveFrom: new Date(),
         },
       });
+      if (data.sakhiProfile) {
+        await tx.sakhiProfile.create({
+          data: {
+            userId: user.id,
+            primaryProjectId: data.sakhiProfile.primaryProjectId,
+            phoneNumber: data.sakhiProfile.phoneNumber,
+            activeFrom: new Date(),
+          },
+        });
+      }
       return user;
     });
   }
--- a/apps/auth-service/src/auth/auth.routes.ts
+++ b/apps/auth-service/src/auth/auth.routes.ts
@@ -190,11 +190,14 @@ export function registerAuthRoutes(
   doc.post(
     '/users',
     {
-      summary: 'Create a user (ADMIN creates any role; SUPERVISOR creates SAKHI only)',
+      summary:
+        'Create a user (ADMIN creates any role; SUPERVISOR creates SAKHI only). ' +
+        'roleCode SAKHI additionally requires projectId — a matching sakhi_profiles ' +
+        'row is created in the same transaction, using mobileNumber as its phone number.',
       tags: ['Users'],
       responses: {
         201: { description: 'User created', schema: envelope(createdUserSchema) },
-        400: errorResponse(400, { message: 'password: Required' }),
+        400: errorResponse(400, { message: 'projectId: Required to create a SAKHI user.' }),
         401: errorResponse(401),
         403: errorResponse(403, {
           message: 'You do not have access to this resource.',
--- a/apps/auth-service/src/auth/auth.service.spec.ts
+++ b/apps/auth-service/src/auth/auth.service.spec.ts
@@ -316,7 +316,7 @@ describe('AuthService', () => {
       createdAt: new Date('2026-01-01T00:00:00.000Z'),
     };
 
-    it('hashes the password, creates the user and role assignment', async () => {
+    it('hashes the password, creates the user, role assignment, and sakhi_profiles row', async () => {
       repository.findRoleByCode.mockResolvedValue(ACTIVE_ROLE as never);
       repository.createUserWithRole.mockResolvedValue(CREATED_USER as never);
 
@@ -327,6 +327,7 @@ describe('AuthService', () => {
           password: 'Str0ngPass!',
           displayName: 'New Sakhi',
           roleCode: 'SAKHI',
+          projectId: 'project-1',
         },
         ['ADMIN'],
       );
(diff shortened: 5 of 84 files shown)


bharathrams:
General review — overall this looks good. Clear separation of concerns, the forklift rule is respected throughout, the doc comments are genuinely useful (especially the honest `mostRecentIncVisitHrType` approximation note and the infinite-loop rationale on `findBeneficiaryOwnership`), and the IDOR guards are in the right places. A few things worth a look:

**1. `findRecentCompletedIncVisits` orders by a different column than "completed"** (`visitInstance.repository.ts`)

It filters on `completedAt: { not: null }` but orders by `actualVisitDate desc, nulls last`. `findRecentCompletedVisits` right below it orders by `completedAt desc`. Since BR-13 picks "the last 3 completed INC visits" and then treats `recentIncVisits[0]` as *the most recent* one for `mostRecentIncVisitHrType`, the ordering column matters — a visit conducted earlier but back-dated/submitted later can land in slot 0. Worth making both use `completedAt` unless there's a reason for the split.

**2. Duplicate ownership-scoping block, copy-pasted 4×**

The exact same SAKHI-own-case / SUPERVISOR-own-roster / else-unrestricted block now appears in `form.service.ts#getLatestVisitVitals`, `visitInstance.service.ts#getVisitHistory`, `riskAssessment.service.ts#listByVisitIds`, plus `beneficiary.service.ts#getOwnership`/`setCcvOpeningRiskState`. That's the kind of check where one copy drifting is a security bug, not a style issue. Worth extracting to a shared `assertCallerOwnsBeneficiary(beneficiary, caller, authHeader)` helper.

**3. BR-13 phase read costs an extra round trip on every CCV submission**

In `createSubmission`, `findBeneficiaryById` is called before every `CCV_VISIT` submission just to read `currentPhase`, even though `isFirstCcvTransition` is true at most once per case. Not wrong, but it's a full enriched `GET /beneficiaries/:id` (which itself fans out to risk-conditions + vitals) on the hot submission path. Could `updateBeneficiaryPhase` return the previous phase instead, or use the lighter `/ownership`-style endpoint?

**4. `visitIds` batch cap vs. `findAllCompletedInfantVisitIds` (unbounded)**

`listRiskAssessmentsQuerySchema` caps `visitIds` at 100, but `findAllCompletedInfantVisitIds` returns *every* completed NN/INC/INC_HR visit with no limit and the resolver joins them all into one query string. A beneficiary with >100 infant visits is unlikely in 0–12m, but if it ever happens the DTO rejects the request and `listRiskAssessments` returns `null` → BR-13 silently skips (best-effort, so no error surfaces). Also note the ids go in a comma-joined query string — worth a sanity check against gateway/proxy URL length limits at 100 UUIDs (~3.7 KB).

**5. `resolveVisitCompletion` sets `meetBeneficiaryFlag: existing.meetBeneficiaryFlag ?? true`**

Defaulting an unset "did you meet the beneficiary" to `true` writes a clinical fact the Sakhi never asserted. If the field is nullable in the schema, leaving it null seems safer than assuming yes. (Same for `actualVisitDate ?? new Date()` — that one's more defensible.)

**6. `extractVitals('', {})` as an empty-value sentinel**

Used in three places (`form.service.ts`, `visitInstance.service.ts` ×2) to produce an all-null snapshot by passing a formCode that intentionally misses the lookup. It works, but it reads as accidental. `vitalsExtractor.ts` already has `EMPTY_VITALS` — exporting that (and an `EMPTY_VISIT_HISTORY_VITALS`) would be clearer than relying on a lookup miss.

**7. Unrelated changes in the PR**

The auth-service `sakhiProfile`-in-transaction fix and the `findConsecutiveNoImprovementCount` N+1 batching are both good changes, but they're unrelated to this PR's stated scope (risk color / phase-advance / BR-13 / vitals) and neither is mentioned in the description. Fine if intentional, just flagging that they'll be easy to lose track of in the history.

Minor: `visitVitals.client.ts` uses `console.error` for a degrade-to-null path while the equivalent paths in `riskAssessment.client.ts` and the resolvers use `console.warn` — worth aligning.

Nothing here blocks in my view — 1, 2 and 5 are the ones I'd want addressed before merge.

Dharanishwaran-ND:
Thanks for the thorough review — addressed in 21d2749:

**Fixed:**
1. **Ordering fix** — `findRecentCompletedIncVisits` now orders by `completedAt desc`, matching the "completed" filter it applies and `findRecentCompletedVisits`'s own convention. A visit conducted earlier but completed later can no longer outrank a more recently completed one in `recentIncVisits[0]`.
2. **Shared ownership check** — extracted `assertCallerOwnsBeneficiary` (new `beneficiaries/beneficiaryOwnership.guard.ts`) and wired it into both `form.service.ts#getLatestVisitVitals` and `visitInstance.service.ts#getVisitHistory`, replacing the copy-pasted block. Added dedicated test coverage for the helper itself (8 tests covering SAKHI/SUPERVISOR/MANAGER/ADMIN + edge cases). risk-referral-service keeps its own copy — no shared code across services, per the forklift rule.
5. **meetBeneficiaryFlag** — no longer defaults to `true`; left `null` (the field's own nullable type) when unset, since auto-completion never asked the Sakhi that question. Added a test asserting this explicitly.
6. **Empty-vitals sentinel** — exported `EMPTY_VITALS`/`EMPTY_VISIT_HISTORY_VITALS` from `vitalsExtractor.ts`; both call sites now use them directly instead of relying on `extractVitals('', {})`'s lookup-miss.
Minor: aligned `visitVitals.client.ts` to `console.warn`, matching every other degrade-gracefully resolver.

**Deferred (flagged as non-blocking, and both need an API-surface change beyond this PR's scope):**
3. **BR-13 extra round trip** — confirmed real: `findBeneficiaryById` fires on every `CCV_VISIT` submission, not just the transition. The cheap fix (reusing `findBeneficiaryOwnership`) doesn't carry `currentPhase`/`childDateOfBirth`, so a real fix needs either extending `/ownership`'s response shape or having `updateBeneficiaryPhase` return the previous phase — both are backend contract changes I didn't want to make unreviewed. Happy to take this as a follow-up if you'd like it scoped now instead.
4. **visitIds batch cap vs. unbounded infant-visit scan** — agreed this is a real (if currently unreachable in practice) mismatch; tracking as a follow-up rather than guessing at the right cap/pagination shape here.

Also confirmed: the auth-service `sakhiProfile` fix and the `findConsecutiveNoImprovementCount` batching were both already-merged, pre-existing work (from PR #172, before this branch's work began) — not something this PR introduced, just showing up in the diff because this branch is downstream of that commit.

PR #187, opened 2026-08-22
fix(auth-service): Match supervisor-app seed accounts by username too

### LINKED ISSUE #186: auth-service seed: supervisor-app-account idempotency check misses username collisions
## Problem
\`seedSupervisorAppAccounts\` (apps/auth-service/prisma/seed.ts) checks whether a
supervisor/sakhi seed account already exists by querying \`prisma.user.findUnique({ where: { id: account.id } })\`
— i.e. only by the seed's fixed hardcoded id.

If a row with the same \`username\` already exists under a *different* id (e.g. from
an earlier partial seed run, a manual account creation, or a reset that didn't fully
clear the \`users\` table), this check finds nothing and falls through to
\`tx.user.create()\`, which then fails on the database's real \`username @unique\`
constraint with a P2002 error — instead of skipping gracefully like every other
seed step in this file.

## Repro
Observed on a shared dev-style environment: running the full seed script a second
time (after a partial prior run) failed with:
\`\`\`
Unique constraint failed on the fields: (\`username\`)
\`\`\`
inside \`seedSupervisorAppAccounts\`, blocking every seed step after it.

## Fix
Extend the existence check to also match by \`username\`, matching the idempotent
"skip if already-seeded" pattern used by every other step in this seed file.

Dharanishwaran-ND:
## Summary
- Fixes #186 — `seedSupervisorAppAccounts` only checked for an existing row by
  the seed's fixed hardcoded `id`, so a row with the same `username` under a
  different `id` (e.g. from an earlier partial seed run) wasn't detected, and
  the insert failed on the DB's `username @unique` constraint instead of
  skipping gracefully like every other step in this seed file.
- Extended the existence check to `findFirst` matching on `id` OR `username`,
  matching the idempotent "skip if already-seeded" pattern used elsewhere in
  the file.
- **Second fix bundled in**: `PATCH /users/:id` accepted a SAKHI's
  `roleCode`+`projectId` and updated the `user_roles` scope row, but never
  synced `sakhi_profiles.primaryProjectId` — the field every roster/ACL check
  (`SakhiRepository.findByProject`, `SakhiService.getById`) actually reads.
  Reassigning a Sakhi to a new supervisor's project therefore left her
  invisible to that supervisor even though the API call reported success.
  Fixed by also updating `primaryProjectId` when `roleCode === 'SAKHI'` and
  `projectId` is provided (rejecting `projectId: null` for a SAKHI, since
  `primaryProjectId` is non-nullable).

## Test plan
- [x] `npx tsc --noEmit -p apps/auth-service/tsconfig.json` — clean
- [x] `npx nx test auth-service` — 321/321 passing (pre-existing unrelated
      failure in `ses-email.client.spec.ts` due to a missing `@aws-sdk/client-ses`
      dependency, not touched by this change)
- [x] `auth.service.spec.ts`'s `updateUser` suite: 77/77 passing, including 4
      new tests for the `primaryProjectId` sync behavior
- [ ] Re-run the seed script on the affected environment to confirm it now
      skips existing usernames instead of throwing P2002
- [x] Verified live on dev (api.armman.org): PATCH with roleCode SAKHI +
      projectId now updates primaryProjectId, confirmed via the target
      supervisor's own roster call returning the reassigned Sakhi

### DIFF
--- a/apps/auth-service/prisma/seed.ts
+++ b/apps/auth-service/prisma/seed.ts
@@ -489,7 +489,9 @@ async function seedSupervisorAppAccounts(scope: {
   const results: SeedResult[] = [];
 
   for (const account of SUPERVISOR_APP_ACCOUNTS) {
-    const existing = await prisma.user.findUnique({ where: { id: account.id } });
+    const existing = await prisma.user.findFirst({
+      where: { OR: [{ id: account.id }, { username: account.username }] },
+    });
     if (existing) {
       results.push({
         step: `supervisorAppAccount:${account.username}`,
--- a/apps/auth-service/src/auth/auth.service.spec.ts
+++ b/apps/auth-service/src/auth/auth.service.spec.ts
@@ -771,7 +771,7 @@ describe('AuthService', () => {
     });
 
     describe('role/project/geography scope', () => {
-      it('updates projectId and geographyUnitId on the matching active role row', async () => {
+      it('updates projectId and geographyUnitId on the matching active role row for a non-SAKHI role', async () => {
         repository.findActiveUserRole.mockResolvedValue({ id: 'user-role-1' } as never);
         repository.findProjectById.mockResolvedValue({ projectId: 'project-2' } as never);
         repository.updateUserTransaction.mockResolvedValue(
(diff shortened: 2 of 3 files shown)


Dharanishwaran-ND:
no need this change now

PR #190, opened 2026-08-24
feat(visit-form-service): Generate visit schedules via GoRules

### LINKED ISSUE #189: Wire GoRules into server-side visit-schedule generation (SRS mandate not implemented)
## Problem

The SRS (docs/Arogya_Sakhi_SRS_v3.0.md) mandates that ANC/PP/NN/INC/CCV/HR/
DELIVERY visit-schedule timing be config-driven via GoRules, with no
hardcoded logic (§1081: "gorules is mandatory for ALL workflow rules... This
applies to: ANC/PP/NN/INC/CCV scheduling formulas"; §1866-1897 names "Visit
scheduling parameters" as GoRules-administrator-configurable).

In the live code, rules-service already has real, seeded, PUBLISHED GoRules
decision packs for all 7 journeys (`apps/rules-service/src/rules/scheduling/
*.rulesJson.ts`) — but only CCV's pack is ever actually invoked, and only to
compute a risk-tier flag (`ccvOpeningRiskState`), never a schedule date.
`visit-form-service`'s only schedule-creating endpoint,
`POST /visit-schedules/bulk`, takes caller-supplied dates directly and never
calls rules-service to compute anything — it only validates that
`generatedByRuleVersionId` points to *some* PUBLISHED version, without
invoking it.

So for ANC/PP/NN/INC/HR/DELIVERY, GoRules exists but is unreachable from any
server-side write path — the SRS's "no hardcoded logic" mandate is
unmet for 6 of 7 journeys.

## Fix

Add `POST /visit-schedules/generate`, a new endpoint that calls
rules-service's already-published `evaluate-schedule` for the requested
journey, remaps its output into `VisitSchedule` rows, and persists them
through the same idempotent path `/bulk` already uses — closing the gap
without touching the existing device-upload flow.

Dharanishwaran-ND:
## Summary
- Fixes #189 — the SRS mandates GoRules drive ANC/PP/NN/INC/CCV/HR/DELIVERY
  visit-schedule timing with no hardcoded logic. rules-service already has
  real, seeded, PUBLISHED GoRules packs for all 7 journeys, but only CCV's
  is ever invoked (and only for a risk-tier flag, not dates) —
  `POST /visit-schedules/bulk` just persists caller-supplied dates and never
  calls rules-service to compute anything.
- Adds `POST /visit-schedules/generate`: calls rules-service's
  `evaluate-schedule` for the requested journey (ANC/PP/NN/INC/HR/DELIVERY —
  CCV is untouched, it keeps its existing risk-tier-only wiring), remaps the
  pack's output into `VisitSchedule` rows via `scheduleMapper.ts`, and
  persists them through the exact same idempotent/ownership-checked path
  `/bulk` already uses (extracted into a shared `persistSchedules` method).
- `generate-visit-schedule.dto.ts` is a discriminated union keyed on
  `scheduleKind` — each variant carries exactly the fields that journey's own
  `rulesJson` pack destructures. A first version used one shared "every
  field, optional" shape; live testing against rules-service caught that it
  silently sent the wrong fields for NN/HR/DELIVERY, which 502'd instead of
  giving a clean 400 — the discriminated union makes that impossible at the
  type level now.
- 6 new optional per-journey `*_SCHEDULE_RULE_SET_ID` env vars
  (`app-config.ts`, `.env.example`), same pattern as the existing
  `CCV_SCHEDULE_RULE_SET_ID` — a journey with none configured 400s on
  `/generate`, it doesn't fail the service at boot.
- Also adds three read-only `rules-service` endpoints so a mobile client can
  fetch a rule pack's published content directly, instead of only through
  service-to-service `evaluate-*` calls:
  - `GET /rules/versions/:versionId/content` — the full `rulesJson` JDM
    graph, but only for the currently-PUBLISHED version (a DRAFT/RETIRED id
    404s exactly like an unknown id, so it never reveals an unpublished
    pack's existence to a non-admin caller).
  - `GET /rules/:setId/published-version` — resolves a rule set's current
    published `versionId`, open to any authenticated role (unlike the
    ADMIN-only `GET /admin/rules/:setId`).
  - `GET /rules/published-content?setIds=...` — batch variant of the above
    two combined, for a mobile client's full sync (SCHEDULE ×6 + RISK ×2) in
    one round trip; a `setId` with no published version is simply omitted
    from the result, never errors the whole batch.

## Test plan
- [x] `npx tsc --noEmit -p apps/visit-form-service/tsconfig.json` — clean
- [x] `npx nx test visit-form-service` — 576/576 passing (44 new/updated
      across `visitSchedule.service.spec.ts`, `scheduleMapper.spec.ts`,
      `scheduleRuleSets.spec.ts`)
- [x] Verified live, end-to-end, through a real gateway → visit-form-service
      → rules-service chain for all 6 journeys (ANC 10 rows, PP 5, NN 2,
      INC 11, HR 1, DELIVERY 0 rows + correct dispatch decision), confirming
      dates are genuinely GoRules-computed, not caller-supplied. Re-ran the
      same call twice to confirm idempotency (`created: 0, alreadyExisted: 10`
      on replay).
- [x] `npx nx lint rules-service` — clean
- [x] `npx nx test rules-service` — 179/179 passing
- [x] Verified `GET /rules/versions/:versionId/content` live against
      rules-service directly: 200 with the expected `rulesJson`/`status` for
      a seeded PUBLISHED version, 404 for an unknown/unpublished id.
- [ ] Set the 6 new `*_SCHEDULE_RULE_SET_ID` env vars in each deployed
      environment once this merges — real ruleSetIds are already seeded and
      PUBLISHED on dev (`api.armman.org`), confirmed via `GET /rules` and a
      live `evaluate-schedule` call; other environments need the same seed
      run (`npm run seed` in rules-service) if not already done, then the
      env vars set from the resulting `ruleSetId`s.


apps/visit-form-service/src/visit-schedules/scheduleMapper.ts
```diff
@@ -0,0 +1,146 @@
...
+    case 'HR': {
+      const generateHrVisit = evaluation.generateHrVisit;
+      const hrVisit = evaluation.hrVisit;
+      if (generateHrVisit !== true) return [];
+      if (!isRuleVisitWindow(hrVisit)) {
+        throw badRequest(
+          'HR evaluation hrVisit is malformed or missing despite generateHrVisit=true.',
+        );
+      }
+      return [toRow(beneficiaryId, 'ANC_HR', 'ANC_HR', 1, hrVisit, 'ACTUAL_VISIT')];
```
palani-ND: **Correctness bug**: the HR case always labels the generated row `ANC_HR`/`ANC_HR1`, regardless of the caller-supplied `phase` (`ANC` | `INC` | `CCV`) in `generate-visit-schedule.dto.ts`. `toBulkScheduleRows` never receives `phase` (only `beneficiaryId`, `scheduleKind`, `evaluation`), and `hr.rulesJson.ts`'s output (`generateHrVisit`/`hrVisit`/`cumulative`) doesn't echo it back either. But `visitCodeTypeSchema` in `create-visit-schedule-bulk.dto.ts` explicitly defines distinct `INC_HR`/`CCV_HR` visit types, so this is a real, intended distinction elsewhere in the domain model.

**Failure scenario**: `POST /visit-schedules/generate` with `{ scheduleKind: 'HR', phase: 'INC', hrDetectedThisVisit: true, actualCompletionDate: ... }` persists a row with `visitType: 'ANC_HR'`/`visitCode: 'ANC_HR1'` instead of `INC_HR`/`INC_HR1'`. No validation catches this (the `trailingSequenceNo` check just compares the same wrong prefix against itself), so the mislabeled row silently persists and any downstream logic keyed on visitType (analytics, escalation, dispatch) misclassifies the visit's phase. None of the new tests exercise `phase: 'INC'` for HR generation, so this ships uncaught.

apps/rules-service/src/rules/ruleVersion.repository.ts
```diff
@@ -37,6 +37,25 @@ export class RuleVersionRepository {
...
+  /**
+   * The currently-PUBLISHED version for each of a batch of rule sets — a
+   * plain `findMany`, no transaction/locking (unlike publishNewVersion),
+   * since this is a read-only batch lookup for a mobile client's full sync.
+   * A ruleSetId with no published version (unknown id, or never published)
+   * is simply absent from the result array — the caller doesn't 404 the
+   * whole batch for one missing/unpublished set.
+   */
+  findPublishedManyBySetIds(ruleSetIds: string[]) {
+    return this.prisma.ruleVersion.findMany({
```
palani-ND: **Plausible concurrency bug**: `findPublishedManyBySetIds` runs a plain `findMany` on `{ ruleSetId: in ruleSetIds, status: PUBLISHED, effectiveTo: null, isDeleted: false }` with no `orderBy`/dedup, whereas the sibling single-id lookup `findPublishedBySetId` (just above) defensively adds `orderBy: { effectiveFrom: 'desc' }` plus `findFirst` to pick exactly one row. There's no DB-level constraint enforcing 'at most one PUBLISHED+effectiveTo:null row per ruleSetId' (schema.prisma's `RuleVersion` only has `@@unique([ruleSetId, versionNo])`) — that invariant is upheld purely by `publishNewVersion`'s transaction retiring the old version.

**Failure scenario**: if that transaction is ever raced (e.g., two concurrent `/admin/rules/:setId/publish` calls, or a partial failure) and two rows end up PUBLISHED+effectiveTo:null for the same ruleSetId, `getPublishedContentBatch`/`GET /rules/published-content` returns two conflicting `rulesJson` entries for that one `ruleSetId` in the batch array, and a mobile client doing its full SCHEDULE/RISK sync gets an ambiguous/duplicate result instead of the single deterministic version `findPublishedBySetId` would have returned for the same data.

apps/visit-form-service/src/rules/evaluateSchedulePack.client.ts
```diff
@@ -0,0 +1,46 @@
+import { badGateway } from '@armman/service-commons';
```
palani-ND, line 1: **Reuse/naming collision**: this new file and the pre-existing `apps/visit-form-service/src/rules/scheduleEvaluate.client.ts` both live in the same directory and both export a function named `evaluateSchedule` with the identical 4-arg signature `(ruleSetId, scheduleKind, input, authorizationHeader)`, hitting the exact same URL and building the identical request body — they differ only in error-handling contract (this one throws `badGateway` on any failure; the other degrades to `null` for CCV's best-effort hint).

Since the PR's own comment on this file already has to spell out 'unlike scheduleEvaluate.client.ts's best-effort degrade-to-null', the duplication is clearly a known trade-off, but it's still a maintenance hazard: a future refactor, codemod, or IDE auto-import that resolves `evaluateSchedule` to the wrong file silently swaps a 'throw and fail the request' path for a 'swallow and return null' path (or vice versa) with no compiler error to catch it, since both satisfy a structurally similar signature. Worth extracting the shared fetch/URL/body-building into one low-level helper that each file wraps with its own error semantics, or renaming one of the two exports so they can't be confused by name alone.

apps/rules-service/src/rules/ruleVersion.routes.ts
```diff
@@ -150,6 +187,88 @@ export function registerRuleVersionRoutes(doc: DocumentedRouter, service: RuleVe
...
+      },
+    },
+    trustGatewayIdentity,
+    requireRoles('SAKHI', 'SUPERVISOR', 'MANAGER', 'ADMIN'),
+    validate(versionIdParamsSchema, 'params'),
+    controller.getContentById,
+  );
+
+  doc.get(
+    '/rules/published-content',
```
palani-ND, line 217: **Convention (CLAUDE.md)**: `.claude/CLAUDE.md` §3 states 'Files ≤ ~250 lines — split by responsibility when larger.' This file was already over that (284 lines on `develop`) before this PR; the three new routes added here push it to 403 lines. Similarly `apps/visit-form-service/src/visit-schedules/visitSchedule.service.ts` grows from 271 to 365 lines in this same PR. Rather than growing an already-oversized file further, the three new read-only rules-content endpoints (`getContentById`/`getPublishedVersionId`/`getPublishedContentBatch` + their schemas) could have been split into their own route/schema module, per the 'split by responsibility when larger' guidance.

apps/visit-form-service/src/visit-schedules/scheduleMapper.ts
```diff
@@ -0,0 +1,146 @@
...
+import type { GeneratableScheduleKind } from './dto/generate-visit-schedule.dto';
+
+interface RuleVisitWindow {
+  visitName: string;
+  scheduledDate: string;
+  windowOpen: string;
+  windowClose: string;
+}
+
+function isRuleVisitWindow(v: unknown): v is RuleVisitWindow {
```
palani-ND, line 12: **Altitude**: `isRuleVisitWindow` requires `visitName` to be a string, but no code path ever reads `window.visitName` afterward — `visitCode`/`sequenceNo` for ANC and INC are instead derived purely from the visit's position in the `visits` array (`i + 1` at lines 96/122), not from the rule pack's own validated name for that visit. This happens to line up today only because `anc.rulesJson.ts` and `inc.rulesJson.ts` both only ever drop trailing visits (EDD/cutoff overruns), never a middle one, so the remaining array stays contiguous.

That contiguity is an implementation detail of those two specific rule packs today, not a contract this mapper (or its own type guard) enforces or even checks against the `visitName` it already required and validated. If either pack's logic changes to skip/exclude a visit mid-sequence for any reason (e.g. a referral override, a future business rule), this mapper would silently mislabel the remaining visits' `visitCode`s using the wrong sequence number instead of trusting the rule pack's own already-present, already-validated `visitName`. Deriving `visitCode`/`sequenceNo` from `window.visitName` directly (already validated as a string) would make the mapper correct independent of the source pack's future implementation, and would make the currently-dead `visitName` validation actually meaningful.

apps/visit-form-service/src/visit-schedules/scheduleMapper.ts
Dharanishwaran-ND: Fixed in 45f76fc (well, dc6657d) — `toBulkScheduleRows`'s HR branch now derives `visitCode`/`visitType` from `hrVisit.visitName` ("<phase>-HR" per `hr.rulesJson.ts`) instead of hardcoding `ANC_HR`, with an explicit reject on any unrecognised value. Added parametrized tests covering ANC/INC/CCV plus a rejection case for an unknown `visitName`.

apps/rules-service/src/rules/ruleVersion.repository.ts
Dharanishwaran-ND: Fixed in 45f76fc — `findPublishedManyBySetIds` now orders by `effectiveFrom desc` and dedups to one row per `ruleSetId`, same defensive pattern as `findPublishedBySetId`. Agreed this is currently unreachable given `publishNewVersion`'s transaction, but it's cheap defense-in-depth against exactly the race you describe.

apps/visit-form-service/src/rules/evaluateSchedulePack.client.ts
Dharanishwaran-ND: Fixed in 45f76fc — renamed this file to `evaluateSchedulePack.client.ts` and its export to `evaluateSchedulePack`, so it can't be confused with `scheduleEvaluate.client.ts`'s `evaluateSchedule` by name alone. Went with a rename over extracting a shared fetch helper since the two files' error-handling contracts are genuinely different call sites with different needs, not incidental duplication — happy to revisit if you'd still prefer the extraction.

apps/rules-service/src/rules/ruleVersion.routes.ts
Dharanishwaran-ND: Agreed on the file-length point, but I'm leaving this as-is for this PR rather than splitting `ruleVersion.routes.ts`/`visitSchedule.service.ts` now — both files predate this PR's overage and a split touches routing/DI wiring beyond this PR's scope. Filing it as follow-up debt rather than bundling a structural refactor into this feature PR; let me know if you'd rather block on it.

apps/visit-form-service/src/visit-schedules/scheduleMapper.ts
Dharanishwaran-ND: Good catch on the dead validation — agreed `visitName` is checked but never used for ANC/INC/PP/NN's own visitCode derivation, only for HR now. Leaving this one as-is for this PR since today's rule packs only ever drop trailing visits (never a mid-sequence one), so it's not an active bug — but noted as a real latent risk if that ever changes; would need to switch ANC/INC to deriving visitCode from `visitName` directly rather than array position.

Issue #191, opened 2026-08-25
GoRules risk-grading: clarifications needed for HR thresholds, referral triggers, and form-field gaps


Dharanishwaran-ND:
## Context

Cross-checked against SRS v3.0 (§3A.2.4, §3A.2.5, Appendix D, Appendix E, Appendix J.3), `High Risk Protocols_Developer's Copy` (Appendix D's own source), and `Revised App Form Final 20.3.26` per team process before escalating. Two originally-drafted items were answered by these docs and are dropped below (Infant HR Rules — Appendix D Part 2; Fundal Height expected range — GA ±2cm formula, Appendix D §1.3 / form Q32). One new conflict surfaced during verification is added (item 10, Sickle Cell tier).

## Clarifications needed to finalize GoRules risk-grading logic

1. **PP Phase HR Thresholds** — not available; SRS itself flags PP/NN/INC/CCV threshold sheets as pending from ARMMAN. Please share the HR threshold sheet/rules for the Postpartum (PP) phase.

2. **Hypothermia/Hyperthermia** — Appendix D §2.3 lists a value labeled "Severe hypothermia: <96°F" under the "Mild (At Risk)" row. Please confirm the correct classification.

3. **MUAC** — the protocol states MUAC should be measured only after 6 months, but also includes a 0–6 month MUAC threshold (e.g. Severe <11.5cm for 0–6m). Please confirm the correct age band and applicable threshold.

4. **HR Visit Triggers** — Cord Infection, Respiratory Distress, and Neuro-developmental Status each have an every-instance referral trigger defined, but no HR-visit-trigger value is given. Please confirm whether these should also trigger an HR visit.

5. **Moderate Classification** — the Moderate band is blank across every infant condition table in the source sheet. Please provide the applicable thresholds/classification, or confirm it's intentionally absent.

6. **PPH (Postpartum Hemorrhage)** — both the referral-trigger and HR-visit-trigger columns are blank for PPH in Appendix D, and grade thresholds are incomplete (only Normal/Low and Mild defined). Please confirm whether PPH should trigger a referral, HR visit, or both, and under what conditions/thresholds.

[...]

PR #193, opened 2026-08-26
Learn More placeholder shell + risk-grading fixes (stillbirth guard, Fundal Height GA calc)

### LINKED ISSUE #192: Learn More / contextual health education — build 'Content coming soon' placeholder shell (SRS FR-S-13.1-13.4, FR-S-5.2)
## Context

SRS v3.0 (`docs/Arogya_Sakhi_SRS_v3.0.md`) specifies two related features:

- **FR-S-13.1–13.4** — "Learn More" knowledge base: two-level Sections > Topics, six media types (QnA text/PDF/Infographic/GIF/Video/Audio), offline-cached, Strapi-managed. FR-S-13.4 covers contextual in-form access ("relevant content appears at the bottom of specific form fields").
- **FR-S-5.2** — on HR condition detection, the system "queues a health education message for display."

Per SRS Assumptions (line 166) and Open Item 12 (line 601), confirmed with Niharika Vyas (12 May 2026):

> "Learn More is the last feature to be built after all other features are complete. Initial release will display 'Content coming soon' placeholder until ARMMAN provides final content."

`cms-content-service` (the SRS-designated owner, "wraps Strapi") was previously an empty shell — health/info endpoints only, no domain logic.

## Scope of this issue

Build the placeholder shell only, matching the SRS's own stated interim state — not full Strapi integration, since:
- ARMMAN has not yet delivered the real Learn More content structure (section names, topics, content files) — confirmed genuinely unspecified in the SRS, HLD, and the ERD doc's own "Open Decisions" table (all three mark it PENDING/TBD).
- FR-S-5.2's exact message delivery mechanism (per-condition vs. generic message, queue/notification data shape, condition-to-message mapping) is also confirmed unspecified anywhere in the SRS/HLD/Appendix D.

## What's delivered

- `cms-content-service`: real Prisma schema (`LearnMoreSection`/`LearnMoreTopic`), three endpoints (`GET /learn-more/sections`, `/sections/:code/topics`, `/topics/:code`), seeded with one clearly-marked "Content coming soon" placeholder row.
- `api-gateway`: `/learn-more` route wired, `CMS_CONTENT_SERVICE_URL` added, Swagger aggregator updated.
- `visit-form-service`: optional `learnMoreTopicCode` added to the field-definition schema (FR-S-13.4 hook), unused until real topic codes exist.

## Explicitly out of scope (follow-up needed)

- FR-S-5.2's actual trigger wiring (`isEducationTrigger` on `risk_flags` currently dead-ends — written by the risk rule packs but has no downstream consumer).
- Real Strapi content/taxonomy integration — blocked on ARMMAN providing the content structure.

Dharanishwaran-ND:
## Summary

This branch carries two separate pieces of work, both stacked on the issue #191 doc-comment markers commit:

### 1. Learn More placeholder shell (SRS FR-S-13.1–13.4) — closes #192

Builds the "Content coming soon" placeholder shell the SRS explicitly requires (Assumptions line 166, Open Item 12) until ARMMAN delivers the real Learn More content structure — confirmed genuinely unspecified in the SRS, HLD, and the ERD doc's own "Open Decisions" table (all three mark it PENDING/TBD), so this is a real, working shell rather than a mock.

**Modules touched:**

- **`cms-content-service`** — was a bare health/info-only shell with no domain logic. Now has:
  - Real Prisma schema: `LearnMoreSection` / `LearnMoreTopic` (two-level Sections > Topics per FR-S-13.2, six media types: QNA_TEXT/PDF/INFOGRAPHIC/GIF/VIDEO/AUDIO).
  - Three read-only endpoints, all `requireRoles('SAKHI', 'SUPERVISOR', 'MANAGER', 'ADMIN')`:
    - `GET /learn-more/sections`
    - `GET /learn-more/sections/:sectionCode/topics`
    - `GET /learn-more/topics/:topicCode` (the FR-S-13.4 contextual-lookup route)
  - Seeded with one clearly-marked placeholder section/topic (`COMING_SOON`), fixed UUIDs, idempotent seed.
  - Brought up to the same architecture as every other service in this repo: `DocumentedRouter`/OpenAPI wiring, Prisma-backed `/health/ready`, `main.ts` connect/disconnect lifecycle — none of which existed before.
- **`api-gateway`** — added the `/learn-more` route (was completely unroutable before this PR — verified live), `CMS_CONTENT_SERVICE_URL` config, and registered `cms-content` in the Swagger aggregator (`docs.controller.ts`) — also previously missing.
- **`visit-form-service`** — added an optional, nullable `learnMoreTopicCode` to `formFieldSchema` (FR-S-13.4's contextual in-form hook), so a future form-version PATCH can attach a real Learn More topic to a field without a second schema migration. Confirmed additive: all 91 existing seed-data schema tests still pass unchanged.

**Explicitly out of scope** (tracked as follow-ups in #192): FR-S-5.2's health-education message trigger (`isEducationTrigger` on `risk_flags` is written by the rule packs today but has no downstream consumer — dead-ends), and real Strapi content integration (blocked on ARMMAN providing the content structure).

**Verified live**, not just unit-tested: ran `cms-content-service` and `api-gateway` locally against the dev DB, created a disposable SAKHI account through the real `/users` API, and hit all three endpoints through the actual gateway — 401 unauthenticated, 200 with real seeded rows, 404 on unknown codes, exactly as designed. Migration applied and seed run on dev.

### 2. Risk-grading fixes (pre-existing on this branch, `fce0b5d`/`81cc50f`/`4f72a71`)

- **`fce0b5d`** — comment-only: cross-references GitHub issue #191 (clarifications needed from ARMMAN on HR/risk-grading thresholds) at the exact code locations each open item affects, in `anc-risk.rulesJson.ts`/`infant-risk.rulesJson.ts`. No behavior change.
- **`81cc50f`** — two fixes:
  - Replaced the order-dependent stillbirth guard (`existingChildCount >= liveBirthCount`) with a per-slot check on a new `childDetails.birthOrder` field, so registering a stillborn twin/triplet's slot out of order can no longer block a legitimate live sibling's CHILD case. `GET /beneficiaries/:beneficiaryId/delivery-outcomes` now tags each outcome with its own `birthOrder` instead of a positional array that silently shifted when a slot was empty.
  - Implemented `anc-risk.rulesJson.ts`'s Fundal Height deviation calculation for real — computed from the visit's own gestational age (`visitDate` + `MOTHER_REGISTRATION`'s `lmpDate`) instead of expecting a pre-computed input nothing ever supplied.
- **`4f72a71`** — fixes a regression `81cc50f` would otherwise have shipped: `birthOrder` was being required whenever `motherBeneficiaryId` was set on `POST /beneficiaries`, which would have 422'd the standalone Child Registration screen (no `birthOrder` field, unrelated to any stillbirth) for every mother, including those with zero stillbirths. Now the guard fetches delivery outcomes first and only requires `birthOrder` once an actual stillbirth is found on that mother's record.

## Test plan

- [x] `tsc --noEmit` clean on `cms-content-service`, `api-gateway`, `visit-form-service`
- [x] `nx affected -t lint` — 4/4 projects pass
- [x] `nx affected -t test` — 29 suites / 587 tests pass (includes new `learnMore.service.spec.ts`, 6 tests; `beneficiary.service.spec.ts` regression coverage for the birthOrder fix)
- [x] Live end-to-end verification of all three `/learn-more` endpoints through the real gateway with a real SAKHI JWT (disposable test account, created/deleted via the real API)
- [x] Migration + seed run confirmed on dev — `form_definitions`/`risk_conditions`/Learn More rows all verified present via direct query, not just log output

## Follow-ups

- SIT/UAT need the new env vars (`CMS_CONTENT_SERVICE_URL` on gateway; `DATABASE_URL`/`DIRECT_URL`/`PUBLIC_BASE_URLS` on `cms-content-service`) plus `prisma migrate deploy` + seed run applied.
- FR-S-5.2 (health-education message trigger) and real Strapi taxonomy — tracked in #192, blocked on ARMMAN.
- Remaining #191 items (HR threshold clarifications) unaffected by this PR — still pending ARMMAN sign-off.


palani-ND (commented):
Reviewed both halves of this branch. Part 1 (Learn More CMS shell) is clean and well-tested — role-gated endpoints verified as real middleware (not just doc annotations), idempotent seed, proper Prisma lifecycle/health-check, additive migration. Part 2 (risk-grading fixes) correctly fixes the diagnosed bugs (order-dependent stillbirth guard, missing gateway route for delivery-outcomes, unsupplied Fundal Height input) but introduces a new cross-service dependency in the beneficiary-creation critical path that isn't accounted for by the existing best-effort error handling — flagged inline as the main concern. Also worth discussing: bundling an unrelated feature (CMS) with business-logic bug fixes on one branch means a revert of one reverts both.

apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
```diff
@@ -1155,6 +1159,55 @@ export class BeneficiaryService {
...
+    // disambiguate. This is deliberately NOT "required whenever
+    // motherBeneficiaryId is set": the standalone Child Registration screen
+    // (registering a child born before this app was in use, or outside any
+    // recorded delivery session) has no birthOrder field today and calls
+    // this endpoint with motherBeneficiaryId set for a "registered mother"
+    // path unrelated to any stillbirth — that flow must keep working
+    // unchanged. A mother with all live births, or no DELIVERY_VISIT at
+    // all, never requires birthOrder and this guard never blocks her.
+    if (dto.case.caseType === 'CHILD' && dto.case.motherBeneficiaryId) {
+      const outcomes = await resolveDeliveryOutcomesBySlot(
```
palani-ND, line 1185: **High: this can silently reopen the exact bug class cb197c8 just fixed, one hop further down.**

This now makes a synchronous cross-service call (beneficiary-service → gateway → visit-form-service) for *every* CHILD case with `motherBeneficiaryId` set, not just ones near a stillbirth. `resolveDeliveryOutcomesBySlot` throws `badGateway` on any network failure or non-2xx (deliveryOutcomes.client.ts:56-62) — correct, fail-closed design on its own.

But trace the call chain for the DELIVERY_VISIT auto-child-creation path: this `create()` call is itself invoked from `create-child.client.ts`, whose caller wraps the whole thing in a pre-existing best-effort `catch`/`if (!res.ok)` that just `console.warn`s and returns `null` (create-child.client.ts:105-121). So any transient blip in this *new* extra round-trip — a timeout, a gateway restart, a brief network hiccup — causes this to throw `badGateway`, which propagates out of `create()`, comes back as a non-2xx to `create-child.client.ts`, and gets silently swallowed exactly the way cb197c8's own commit message describes as "the root cause" of live births with no beneficiary record. cb197c8 fixed the deterministic 404 case; this reopens a probabilistic version of the identical failure, unmonitored (console.warn only, no alerting).

Also a separate concern: this makes *every* standalone Child Registration (4f72a71's own target use case — no stillbirth involved at all) newly dependent on visit-form-service being reachable, where before it was a purely local write. That's a new SPOF on a previously-independent path.

Suggest: distinguish "the delivery-outcomes check itself failed" from "an actual stillbirth-slot conflict" in the error/logging path so this doesn't silently recur, and add a test in beneficiary.service.spec.ts that asserts what happens when `resolveDeliveryOutcomesBySlot` rejects (only success-path mocks exist today, e.g. lines 2324-2439).

apps/visit-form-service/src/beneficiaries/create-child.client.ts
```diff
@@ -77,6 +82,7 @@ export async function createChildBeneficiary(
       sex: input.sex,
       birthWeightKg: input.birthWeightKg,
       birthLengthCm: input.birthLengthCm,
+      birthOrder: input.birthOrder,
```
palani-ND, line 85: birthOrder now gets threaded through into the auto-create payload here, but the pre-existing best-effort catch/if (!res.ok) a few lines below (log + return null) is the exact mechanism that will hide the new failure mode introduced at beneficiary.service.ts's stillbirth guard (see comment there) — any transient failure of the new delivery-outcomes round-trip on the beneficiary-service side lands here indistinguishable from a real validation rejection. Worth tagging the warn log with the response body/reason so a 502 from the new dependency is distinguishable from an actual 422.

apps/beneficiary-service/prisma/schema.prisma
```diff
@@ -306,6 +306,13 @@ model ChildCaseDetails {
   birthWeightKg       Decimal?             @map("birth_weight_kg") @db.Decimal(5, 2)
   birthLengthCm       Decimal?             @map("birth_length_cm") @db.Decimal(5, 2)
   prematureFlag       Boolean?             @map("premature_flag")
+  // 1-based DELIVERY_VISIT child slot (child1/child2/child3) this case
+  // corresponds to — required at create-time whenever motherBeneficiaryId
+  // is set (see createBeneficiarySchema), null for a standalone/independent
+  // registration with no DELIVERY_VISIT to have a slot on. NULL on any case
+  // created before this column existed — not backfillable, since no slot
+  // was ever captured for those.
+  birthOrder          Int?                 @map("birth_order")
```
palani-ND, line 316: No unique constraint on `(motherBeneficiaryId, birthOrder)`. Nothing in beneficiary.service.ts checks whether a slot already has a CHILD case before creating another one for the same birthOrder. The DELIVERY_VISIT auto-creation path is protected transitively by `localCaseUuid` idempotency, but a manually-submitted standalone registration with a reused birthOrder (once a stillbirth exists on that mother, forcing birthOrder to be supplied) could create two live CHILD cases claiming the same slot with no error. Low likelihood, but worth a follow-up partial unique index if slot integrity matters.

apps/rules-service/src/rules/scheduling/anc-risk.rulesJson.ts
```diff
@@ -322,18 +342,37 @@ const handler = (input) => {
...
+  // submission yet simply skips this condition, same as Age/BOH above. ---
+  const fundalHeightCm = input.fundal_height_in_cm;
+  const lmpDate = input.lmpDate;
+  const visitDate = input.visitDate;
+  if (
+    typeof fundalHeightCm === 'number' &&
+    typeof lmpDate === 'string' &&
+    typeof visitDate === 'string'
+  ) {
+    const gestationalAgeWeeks = dayjs(visitDate).diff(dayjs(lmpDate), 'day') / 7;
```
palani-ND, line 369: No guard against `visitDate` preceding `lmpDate`. A data-entry error (visit recorded before LMP, or LMP corrected after visits already exist) produces a negative `gestationalAgeWeeks`; `Math.floor` on a negative fraction rounds toward -∞ (e.g. -0.7 → -1), so `fundalHeightDeviationCm` comes out inflated by 1 instead of the calculation being skipped. Doesn't crash, but silently produces a nonsensical GA and an ungrounded deviation grade — could mask a real risk as NORMAL or spuriously flag MILD. `anc-risk.rulesJson.spec.ts` covers the missing-lmpDate skip case but not an inverted-date case.

apps/beneficiary-service/src/beneficiary/beneficiary.service.spec.ts
```diff
@@ -2295,6 +2309,173 @@ describe('BeneficiaryService', () => {
...
+    function childDto(birthOrder: number): CreateBeneficiaryInput {
+      return {
+        ...baseChildInput,
+        case: { ...baseChildInput.case, motherBeneficiaryId },
+        childDetails: { dateOfBirth: new Date('2025-12-01'), birthOrder },
+      };
+    }
+
+    it('blocks creating a CHILD case for a slot whose own recorded outcome is a stillbirth', async () => {
+      resolveDeliveryOutcomesBySlotMock.mockResolvedValue([
```
palani-ND, line 2324: All the delivery-outcomes-related tests in this describe block only use `resolveDeliveryOutcomesBySlotMock.mockResolvedValue(...)` — none exercise `.mockRejectedValue(...)`. Given the finding on beneficiary.service.ts:1185, this is the gap that would have caught it: nothing here asserts what `create()` actually does when the delivery-outcomes fetch itself fails (vs. a real stillbirth-slot conflict).

Dharanishwaran-ND:
### Code review

Found 2 issues:

1. New cross-service dependency in the stillbirth guard is silently swallowed by the auto-create caller's best-effort catch, reopening the exact failure class commit `cb197c8` in this same PR just fixed

`create()` now calls `resolveDeliveryOutcomesBySlot` synchronously for every mother-linked CHILD case, which throws `badGateway` on any network failure or non-2xx response:

https://github.com/ARMMAN-Organization/arogyasakhi-service/blob/396a1b33043ce0797d4e0409cda4f12edc9f0298/apps/beneficiary-service/src/beneficiary/beneficiary.service.ts#L1183-L1188

But its only caller in the auto-child-creation path swallows any non-ok response with `console.warn` + `return null`, so a transient failure in this new round-trip silently drops the child's beneficiary record instead of surfacing an error:

https://github.com/ARMMAN-Organization/arogyasakhi-service/blob/396a1b33043ce0797d4e0409cda4f12edc9f0298/apps/visit-form-service/src/beneficiaries/create-child.client.ts#L104-L111

This is the same failure mode `cb197c8` (in this PR) traced and fixed for the missing gateway route (404 → badGateway → swallowed → live births with no beneficiary record). This guard reopens it one hop further down, and also makes the standalone Child Registration flow — previously a purely local write — newly dependent on visit-form-service being reachable. Already flagged inline by @palani-ND on this PR and unresolved.

2. No DB constraint prevents two children from claiming the same `birthOrder` slot for the same mother

`birthOrder` has no unique constraint scoped to `motherBeneficiaryId`:

https://github.com/ARMMAN-Organization/arogyasakhi-service/blob/396a1b33043ce0797d4e0409cda4f12edc9f0298/apps/beneficiary-service/prisma/schema.prisma#L308-L315

The only check is in `create()`, which validates the submitted `birthOrder` against a stillbirth outcome — it doesn't prevent two standalone/manual registrations from being saved with the same `motherBeneficiaryId` + `birthOrder`. Note the adjacent comment ("required at create-time whenever motherBeneficiaryId is set") is also stale — that DTO-level requirement was removed in commit `4f72a71`, which moved the check into `create()` gated on an actual stillbirth being on record.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
Dharanishwaran-ND: Confirmed independently — verified `create()` calls `resolveDeliveryOutcomesBySlot` synchronously for every mother-linked CHILD case (line 1185), and the auto-create caller's best-effort catch (create-child.client.ts:104-111) swallows the resulting `badGateway` with just a console.warn. Reopens the cb197c8 failure class. Flagged in the summary review: https://github.com/ARMMAN-Organization/arogyasakhi-service/pull/193#issuecomment-5423685081

apps/visit-form-service/src/beneficiaries/create-child.client.ts
Dharanishwaran-ND: Confirmed — this is the other side of the same issue on the beneficiary.service.ts:1185 thread. The catch here (line 104-111) is pre-existing best-effort logging that now hides a new failure mode: a badGateway from the stillbirth guard's cross-service call looks identical to a real 422 validation rejection in the logs. Flagged in the summary review: https://github.com/ARMMAN-Organization/arogyasakhi-service/pull/193#issuecomment-5423685081

apps/beneficiary-service/prisma/schema.prisma
Dharanishwaran-ND: Confirmed — no `@@unique([motherBeneficiaryId, birthOrder])` on ChildCaseDetails, and `create()` only validates a submitted birthOrder against a stillbirth outcome, not against another CHILD case already holding that slot. Also worth noting: the comment on this same field (line 311-312, "required at create-time whenever motherBeneficiaryId is set") is stale — that DTO-level requirement was removed in 4f72a71, which moved the check into create() gated on an actual stillbirth being on record. Flagged in the summary review: https://github.com/ARMMAN-Organization/arogyasakhi-service/pull/193#issuecomment-5423685081

apps/rules-service/src/rules/scheduling/anc-risk.rulesJson.ts
Dharanishwaran-ND: Verified the formula and guard match — only type checks, no chronological ordering check between lmpDate and visitDate. I didn't find upstream validation elsewhere in the diff that rules out visitDate < lmpDate reaching this calc (e.g. after an LMP correction). Assessed independently at moderate confidence: real gap, but it requires a specific data-entry/correction scenario rather than the normal path, and the immediate effect is a silently wrong grade rather than a hard failure — worth fixing but didn't clear the bar for the summary review's headline issues.

apps/beneficiary-service/src/beneficiary/beneficiary.service.spec.ts
Dharanishwaran-ND: Agreed — this is the missing coverage that would have caught the beneficiary.service.ts:1185 issue. All the mocks here only exercise the success path (mockResolvedValue); adding a mockRejectedValue case asserting create() surfaces the failure distinctly (rather than it disappearing into the caller's best-effort catch) would close the loop on that finding.

Dharanishwaran-ND:
Pushed 4546a6d addressing the review findings:

1. **Cross-service silent-swallow** — `create-child.client.ts` now logs 5xx/network failures distinctly from real 422 rejections, so a failed stillbirth-slot check no longer reads identically to a genuine conflict in the logs.
2. **Missing unique constraint** — added `@@unique([motherBeneficiaryId, birthOrder])` on `ChildCaseDetails` plus migration `20260826100000_add_child_case_slot_unique_constraint`.
3. **Fundal Height inverted-date guard** — the deviation grade is now skipped when `visitDate` precedes `lmpDate`, rather than grading off a negative gestational age.
4. **Missing test coverage** — added a `beneficiary.service.spec.ts` case asserting `create()` surfaces a distinct `badGateway` (502) when the delivery-outcomes check itself fails, not a stillbirth conflict.

`nx affected -t lint` and the beneficiary-service/rules-service/visit-form-service test suites all pass (185/185, 39/39, 587/587 respectively).

🤖 Generated with [Claude Code](https://claude.ai/code)

apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
Dharanishwaran-ND: Fixed in 4546a6d: create-child.client.ts now logs 5xx/network failures distinctly from real 422 rejections, so a failed check no longer reads identically to a genuine stillbirth conflict in the logs. (Kept the existing best-effort/non-blocking behavior — did not change this to a hard failure.)

apps/visit-form-service/src/beneficiaries/create-child.client.ts
Dharanishwaran-ND: Fixed in 4546a6d — see the reply on the beneficiary.service.ts:1185 thread for details.

apps/beneficiary-service/prisma/schema.prisma
Dharanishwaran-ND: Fixed in 4546a6d: added @@unique([motherBeneficiaryId, birthOrder]) on ChildCaseDetails, plus migration 20260826100000_add_child_case_slot_unique_constraint.

apps/rules-service/src/rules/scheduling/anc-risk.rulesJson.ts
Dharanishwaran-ND: Fixed in 4546a6d: the Fundal Height condition is now skipped (not graded) when visitDate precedes lmpDate, rather than grading off a negative gestational age. Added a regression test for this case in anc-risk.rulesJson.spec.ts.

apps/beneficiary-service/src/beneficiary/beneficiary.service.spec.ts
Dharanishwaran-ND: Fixed in 4546a6d: added a mockRejectedValue test case asserting create() rejects with a badGateway (502), distinct from the 422 CHILD_ALREADY_STILLBIRTH conflict.

PR #199, opened 2026-08-27
feat(beneficiary-service): Batch risk-condition-summary endpoint + isFirstInstance/consecutiveNoImprovementCount

### LINKED ISSUE #198: Batch risk-condition-summary endpoint + precomputed isFirstInstance/consecutiveNoImprovementCount
## Context

Frontend (Bharath) requested status/ETA on two previously-approved backend items blocking full completion of risk persistence on their side:

1. **Batch risk-condition-summary endpoint** — `GET /beneficiaries/risk-condition-summary?beneficiaryIds=...`, returning `{ beneficiaryId, riskConditionSummaries: [...] }[]`, mirroring the existing `GET /beneficiaries/by-ids-with-risk` batch pattern.
2. **Precomputed `isFirstInstance` / `consecutiveNoImprovementCount`** — added to that same batch response, computed server-side using the exact logic the real grading pipeline (risk-referral-service) already uses internally as rule-pack evaluation inputs, but never previously exposed anywhere.

Neither had been built — item 2's computation logic (`findEverFlaggedConditionCodes`/`findConsecutiveNoImprovementCount`) existed only privately inside risk-referral-service's grading pipeline with no route exposing it.

## Design decision

Since risk-referral-service owns the RiskAssessment/RiskFlag source-of-truth tables and beneficiary-service cannot cross-service-join into them (forklift rule), the two new fields are computed at write-time by risk-referral-service (already computed there for rule-pack evaluation) and pushed via the existing `pushRiskConditionSummary` call into two new columns on beneficiary-service's `BeneficiaryRiskConditionSummary` table. The new batch endpoint then reads its own table with no new outbound cross-service call at read-time.

## Scope

- risk-referral-service: push `isFirstInstance`/`consecutiveNoImprovementCount` alongside the existing per-condition rollup push.
- beneficiary-service: two new columns + migration, extended upsert DTO/repository/service, new `GET /beneficiaries/risk-condition-summary` batch endpoint (route/controller/service/repository), mirroring `by-ids-with-risk`'s ownership-scoping and silent-drop-on-out-of-scope-id semantics.

## Verification

- 34 risk-referral-service tests, 356 beneficiary-service tests (all passing, zero regressions).
- Live end-to-end verification through api-gateway with a real JWT: repeated grading submissions correctly produced `isFirstInstance:false` + growing `consecutiveNoImprovementCount`, condition name/grade-scale resolution confirmed, silent-drop and 400-validation edge cases confirmed.

Dharanishwaran-ND:
## Summary
- Adds `GET /beneficiaries/risk-condition-summary?beneficiaryIds=...` — batch risk-condition-summary lookup, mirroring the existing `by-ids-with-risk` batch pattern (ownership scoping, silent-drop on out-of-scope/nonexistent ids).
- Adds two precomputed fields to each risk condition summary — `isFirstInstance` and `consecutiveNoImprovementCount` — sourced from logic risk-referral-service's grading pipeline already computes internally as rule-pack evaluation inputs, but never previously exposed.
- Persisted at write-time: risk-referral-service pushes both fields via the existing `pushRiskConditionSummary` call into two new columns on `BeneficiaryRiskConditionSummary`, avoiding a new cross-service read-time call (beneficiary-service cannot join into risk-referral-service's source-of-truth tables per the forklift rule).

Closes #198

## Test plan
- [x] `npx nx test risk-referral-service` — 34 tests passing (7 new)
- [x] `npx nx test beneficiary-service` — 356 tests passing (24 new)
- [x] `npx nx lint risk-referral-service && npx nx lint beneficiary-service` — clean
- [x] Live end-to-end verification through api-gateway with a real JWT: repeated grading submissions for the same beneficiary/condition correctly produced `isFirstInstance:false` + `consecutiveNoImprovementCount` incrementing; `conditionCode`/`conditionName`/`gradeScale` resolved correctly; out-of-scope/nonexistent beneficiaryId silently omitted (not 404/403); missing/empty `beneficiaryIds` → 400.

## Notes
- New migration `20260827090000_add_risk_condition_summary_first_instance_streak` — additive-only (2 new columns on `Beneficiary_risk_condition_summary`, one `NOT NULL DEFAULT true`, one nullable `INTEGER`), no data loss.


apps/risk-referral-service/src/referrals/referral.service.ts
```diff
@@ -26,8 +29,37 @@ export class ReferralService {
...
+   * visitId are unrestricted.
+   */
+  async create(dto: CreateReferralInput): Promise<{ referral: Referral; alreadyExisted: boolean }> {
+    const validTill = addDays(dto.referralDate, 7);
+    try {
+      const referral = await this.repository.create({ ...dto, validTill });
+      return { referral, alreadyExisted: false };
+    } catch (err) {
+      if (!isUniqueConstraintViolation(err, 'visit_id')) throw err;
+      const existing = dto.visitId ? await this.repository.findByVisitId(dto.visitId) : null;
```
palani-ND, line 64: On a `visit_id` unique-constraint collision, this returns whatever row already exists (`alreadyExisted: true`, HTTP 200) with no comparison between the submitted `dto` and the existing referral's fields (`beneficiaryId`, `referralTypeLookupValueId`, `status`, etc.). This diverges from this PR's own design doc (`docs/superpowers/plans/2026-08-25-referral-create-409-conflict.md`), which specifies mapping this collision to a `409 Conflict` — the shipped code instead silently returns a 200 success with the *original* (possibly now-wrong) referral. Concretely: a referral resubmitted with a corrected `referralTypeLookupValueId` after a mistake, or a `visitId` genuinely reused for a different beneficiary, returns the stale/wrong row as if the request succeeded, with no way for the caller to detect its second submission didn't take effect. Worth comparing the submitted fields against `existing` and returning a real conflict when they differ, matching the design doc.

apps/risk-referral-service/src/referrals/referralFollowup.repository.ts
```diff
@@ -0,0 +1,32 @@
...
+    followupStatus: 'COMPLETED' | 'INCOMPLETE',
+    referralStatus: 'COMPLETED' | 'PENDING_FOLLOWUP',
+    data: Omit<CreateReferralFollowupInput, 'mediaAssetIds'>,
+    createdByUserId: string,
+  ) {
+    return this.prisma.$transaction(async (tx) => {
+      const followup = await tx.referralFollowup.create({
+        data: { ...data, referralId, followupStatus, createdByUserId },
+      });
+      const referral = await tx.referral.update({
```
palani-ND: This unconditionally overwrites the referral's status (`tx.referral.update` with a plain `where: { id: referralId }`), unlike every other status-mutating method in this service — `ReferralRepository.updateStatus` and the new `updateType` both use `updateMany` guarded by the expected `fromStatus`, treating a 0-count as a conflict. `ReferralFollowupService.create`'s outer `referral.status !== 'PENDING_FOLLOWUP'` check is read once, before either transaction starts, so two concurrent follow-up submissions for the same referral (e.g. an offline-sync retry racing a genuine second submission, or a follow-up racing a `PATCH /referrals/:id/decision`) can both pass that check and both commit — creating two `ReferralFollowup` rows for one referral with no 409, no error, and a final status that silently depends on commit order. Consider guarding this update with `updateMany({ where: { id: referralId, status: 'PENDING_FOLLOWUP' }, ... })` the same way the sibling repository does.

apps/risk-referral-service/src/referrals/dto/create-referral-followup.dto.ts
```diff
@@ -0,0 +1,33 @@
...
+    followupDate: z.coerce.date(),
+    notVisitedReason: z.string().trim().min(1).max(255).optional(),
+    diagnosis: z.string().trim().min(1).optional(),
+    treatmentGiven: z.string().trim().min(1).optional(),
+    outcome: z.string().trim().min(1).max(255).optional(),
+    // Media evidence — case paper, discharge summary, facility photo,
+    // Sakhi-beneficiary photo, investigation reports — is uploaded to
+    // media-service directly (POST /media/upload-url then POST /media,
+    // tagged with this follow-up's real id and the correct assetType per
+    // asset) BEFORE this call; this field only lists the resulting ids so
```
palani-ND: This instructs clients to tag media evidence with "this follow-up's real id" *before* calling this endpoint — but `ReferralFollowup.id` is server-generated only inside this very endpoint's transaction, so the client cannot know the real id in advance. In practice a client uploading evidence beforehand has to tag it with some placeholder/guessed id. `mediaAssetExists` only checks existence/viewability, not that `followupId` matches, so the follow-up is created successfully — but the new `GET /media?followupId=<real-id>` filtering feature added in this same PR returns nothing for that evidence, since it was never actually tagged with the real id. The evidence becomes orphaned and undiscoverable via the new query path this PR introduces.

apps/risk-referral-service/src/referrals/dto/create-referral.dto.ts
```diff
@@ -57,7 +57,8 @@ export const createReferralSchema = z
       'SKIPPED',
       'CANCELLED',
     ]),
-    validTill: z.coerce.date().optional(),
+    // validTill is server-computed (referralDate + 7 days) — see
```
palani-ND, line 60: `validTill` is dropped from the schema, but `.strict()` (further down) is unchanged — so a client that has always included `validTill` in its request body (even if the server never used it) now gets a hard 400 ("Unrecognized key(s): validTill") instead of the field being silently accepted/ignored as before. This is an unannounced, untested breaking API change — no test in `create-referral.dto.spec.ts` covers a client sending `validTill`, and there's no version gate or migration note in the diff. Worth confirming all client versions in the field have already stopped sending this field before this ships, or relaxing `.strict()` to tolerate it a bit longer.

apps/media-service/src/media/dto/list-mediaAssets.dto.ts
```diff
@@ -0,0 +1,15 @@
+import { z } from 'zod';
+
+/**
+ * Query schema for `GET /media?followupId=...` — narrows the asset list to
+ * one referral follow-up's evidence media (case paper, discharge summary,
+ * facility photo, etc.). Omitting followupId preserves the existing
+ * unfiltered "50 most recent" behavior. `.strict()` rejects unknown fields.
```
palani-ND, line 7: `GET /media` previously had no query-schema validation at all, so any extra/legacy/typo'd query parameter was silently ignored. This new `.strict()` schema now rejects any request with an unrecognized query param with a 400 — a breaking change for any existing caller (internal or external) that was relying on the previous permissive behavior, with no test asserting backward compatibility for unknown params.

apps/risk-referral-service/src/referrals/referral.prisma-errors.ts
```diff
@@ -0,0 +1,21 @@
...
+ * index on this table doesn't get misreported as a visit collision.
+ *
+ * Prisma's P2002 `meta.target` for Postgres reports the underlying
+ * column name(s) the unique index covers (e.g. `["visit_id"]`), NOT the
+ * index's own name (`visit_referral_once`) — verified against a live
+ * collision on this service (2026-08-27); the constraint-name string never
+ * actually appears in `meta.target` at runtime. Callers must pass the
+ * column name, not the index name.
+ */
+export function isUniqueConstraintViolation(err: unknown, columnName: string): boolean {
```
palani-ND, line 14: This P2002-narrowing shape (`typeof err === 'object' && 'code' in err && err.code === 'P2002'`) is now duplicated in at least 8 places across the codebase: `visitSchedule.service.ts`, `auth.service.ts`/`lookup.service.ts`/`project.service.ts`/`geography.service.ts` in auth-service, `reopen-request.service.ts`/`closure.service.ts` in closure-reopen-service, and `operations.service.ts`. This copy is actually the most sophisticated of the bunch (it also disambiguates by `meta.target` column, and its doc comment even documents an empirically-verified Prisma behavior quirk), but that improvement stays local to this service instead of upgrading the shared behavior everywhere. A future Prisma version change to the P2002 error shape means this verification has to be independently rediscovered and reapplied in 8+ places. Worth promoting this into a shared `@armman/service-commons` helper (e.g. `isUniqueConstraintViolation(err, columnName)`) that all these call sites import — `form.service.ts`'s existing generic `isPrismaErrorCode(err, code)` is a natural base to build it on.

apps/risk-referral-service/src/referrals/referralConversion.service.ts
```diff
@@ -0,0 +1,85 @@
...
+   */
+  async convertToAccompanied(
+    referralId: string,
+    caller: AuthenticatedUser,
+    authorizationHeader: string,
+  ) {
+    const referral = await this.repository.findById(referralId);
+    if (!referral) throw notFound('Referral not found.');
+
+    const beneficiary = await this.beneficiaryClient.getById(
```
palani-ND: This ownership-check sequence (`findById` → not-found, `beneficiaryClient.getById` → `sakhiId !== caller.id` → forbidden, then a status check) is copy-pasted verbatim into `referralFollowup.service.ts`'s `create` method — both files' own doc comments even say "Same SAKHI-owns-beneficiary ownership check as ReferralFollowupService.create" / the equivalent, acknowledging the duplication without extracting it. Any future change to this authorization rule (e.g. adding SUPERVISOR roster support, or changing the required status) has to be found and edited in two places with nothing enforcing they stay in sync. Consider extracting a shared helper (e.g. `assertOwnedPendingFollowupReferral(...)`) that both `convertToAccompanied` and `create` call.

apps/risk-referral-service/src/referrals/referralFollowup.service.ts
```diff
@@ -0,0 +1,77 @@
...
+    );
+    if (!beneficiary || beneficiary.sakhiId !== caller.id) {
+      throw forbidden('This referral does not belong to your own roster.');
+    }
+
+    if (referral.status !== 'PENDING_FOLLOWUP') {
+      throw conflict(`Cannot submit a follow-up for a referral with status ${referral.status}.`);
+    }
+
+    for (const mediaAssetId of dto.mediaAssetIds) {
```
palani-ND: This loop `await`s `mediaAssetExists` one id at a time; each call is an independent HTTP round-trip to media-service with no data dependency on the others. With up to 10 `mediaAssetIds` (the DTO's max), a submission with several attachments pays for up to 10x a single round-trip's latency serially before the request can proceed. Consider `const results = await Promise.all(dto.mediaAssetIds.map(id => mediaAssetExists(id, authorizationHeader)))` instead.

apps/risk-referral-service/src/risk-assessments/riskAssessment.service.ts
```diff
@@ -207,6 +213,15 @@ export class RiskAssessmentService {
           isReferralTrigger: condition.isReferralTrigger,
           isHrVisitTrigger: condition.isHrVisitTrigger,
           ruleVersionId: evaluation.ruleVersionId,
+          // conditionCode is always resolvable here (evaluation.conditions'
+          // riskConditionIds all originate from conditionIdsByCode, which
+          // conditionCodeById is the exact reverse of) — the `?? true`/`?? null`
+          // fallbacks exist only to satisfy the type checker, not because this
+          // path is expected to be hit.
+          isFirstInstance: conditionCode ? isFirstInstance[conditionCode] : true,
```
palani-ND: This falls back to `isFirstInstance: true` whenever `conditionCodeById.get(condition.riskConditionId)` misses, and the surrounding comment states this path "is not expected to be hit." If a stale/mismatched rule-pack version ever returns a `riskConditionId` not present in the current phase's active condition set, this would silently report a long-standing, previously-flagged chronic condition as a first instance — with no error, warning, or log to catch it. Given `isFirstInstance` can gate downstream first-instance-only logic (referral/education triggers), a silent miss here has real behavioral consequences; consider logging or throwing on the miss rather than defaulting to `true`. Separately, worth noting the accompanying migration (`add_risk_condition_summary_first_instance_streak`) backfills every pre-existing `BeneficiaryRiskConditionSummary` row to `isFirstInstance: true` unconditionally — any beneficiary whose condition hasn't been re-graded since this migration deploys will show as "first instance" for an already-established condition until the next grading pass touches that row.

palani-ND (commented):
Automated review across 8 independent passes (correctness, removed-behavior, cross-file, reuse, simplification, efficiency, altitude, CLAUDE.md/STANDARDS.md conventions). Two findings below (idempotent-create and the missing follow-up concurrency guard) are corroborated by multiple independent passes and are worth treating as blocking. Findings ranked roughly by severity.

apps/risk-referral-service/src/referrals/referral.service.ts
Dharanishwaran-ND: Fixed in 222443e: the idempotent-retry path now compares `beneficiaryId`, `referralTypeLookupValueId`, `facilityType`, and `facilityName` against the existing row — a mismatch on any of these now 409s instead of silently returning the stale referral as a 200. Added 3 new tests covering each mismatch case.

apps/risk-referral-service/src/referrals/referralFollowup.repository.ts
Dharanishwaran-ND: Fixed in 3b2297e: switched to `updateMany({ where: { id, status: 'PENDING_FOLLOWUP' } })` matching the sibling pattern — a 0-count now throws `ReferralNoLongerPendingFollowupError`, mapped to a 409 by the service. Added a test covering the concurrent-submission case.

apps/risk-referral-service/src/referrals/dto/create-referral-followup.dto.ts
Dharanishwaran-ND: Fixed in fff7966: corrected the comment to describe the only actually-possible flow — media is finalized AFTER this call succeeds, using the server-generated followupId this endpoint returns, not before it. Good catch — the old comment described an impossible client sequence.

apps/risk-referral-service/src/referrals/dto/create-referral.dto.ts
Dharanishwaran-ND: Agreed this is a real behavior change, but it's the intentional/correct enforcement here: validTill is deliberately made server-computed and non-caller-settable per this PR's own design doc (docs/superpowers/plans/2026-08-25-referral-create-409-conflict.md), and .strict() rejecting it is this repo's global convention (.claude/CLAUDE.md §8) applied consistently rather than carving out an exception for one field. A client still sending validTill getting a loud 400 instead of a silent no-op is the safer failure mode for a field that's being made non-authoritative — better to force a stale client to notice and update than to let it believe its submitted validTill was honored. Not changing this, but flagging for visibility: if there's a known client version still sending this field in production, worth confirming that's been checked before this ships.

apps/media-service/src/media/dto/list-mediaAssets.dto.ts
Dharanishwaran-ND: Agreed this is a breaking change for any caller passing unknown query params, but it's intentional hardening consistent with this repo's global .strict() convention (.claude/CLAUDE.md §8), now extended to a route that was previously ungated. Not changing this — GET /media has no known external consumers today beyond this repo's own services, and the new followupId filter is the only query param this route is meant to accept. Flagging for visibility rather than reverting.

apps/risk-referral-service/src/referrals/referral.prisma-errors.ts
Dharanishwaran-ND: Confirmed — verified this pattern actually appears in 13 files, not just the ones named (also form.service.ts, visitInstance.service.ts, ruleVersion.repository.ts/service.ts). Agreed a shared @armman/service-commons helper is the right end state, but promoting it is a larger cross-service refactor (13 call sites across 5+ services) outside this PR's scope. Deferring — will track as a follow-up rather than bundling into this PR.

apps/risk-referral-service/src/referrals/referralConversion.service.ts
Dharanishwaran-ND: Fixed in a446882: extracted assertSakhiOwnsReferral(referral, caller, beneficiaryClient, authorizationHeader) — both convertToAccompanied and ReferralFollowupService.create now call the shared helper instead of duplicating the findById/getById/forbidden sequence.

apps/risk-referral-service/src/referrals/referralFollowup.service.ts
Dharanishwaran-ND: Fixed in 3b2297e: switched to Promise.all over the mediaAssetIds existence checks instead of a sequential for...of.

apps/risk-referral-service/src/risk-assessments/riskAssessment.service.ts
Dharanishwaran-ND: Fixed the silent-default half in 1f550cc: a conditionCode miss now logs and skips the risk-condition-summary push for that condition entirely (matching the unresolvable-phase branch directly above it in the same loop), rather than guessing isFirstInstance:true. On the migration backfill: confirmed accurate — 20260827090000_add_risk_condition_summary_first_instance_streak does unconditionally default every existing row to isFirstInstance:true with no backfill logic, so an already-chronic condition that hasn't been re-graded since this deploys will misreport as "first instance" until its next grading pass. Not fixing in this PR — a real backfill would need to reconstruct historical first-instance state from risk_flags history, which is a bigger, separate migration. Flagging as a known, accepted limitation rather than silently leaving it undocumented.

Issue #200, opened 2026-08-28
Contextual health-education content — mapping, stage vocabulary, precedence, and delivery mechanism — clarifications needed


Dharanishwaran-ND:
## Type

Cross-service architecture / product decision (mirrors issue #191, but separate from it — #191 is GoRules clinical-threshold gaps; these are product/architecture/content-ops questions one layer above that).

## Affects

risk-referral-service, cms-content-service, sakhi-mobile-app

## Raised by

Bharath (Sakhi App — Frontend), 2026-08-28, following backend's review of CR-M3-06.

## Summary

CR-M3-06 (contextual health education / "Learn More") is buildable today against cms-content-service's real, live, flat sections/topics API, and will correctly show a non-blocking "coming soon" placeholder for every risk-triggered condition. It cannot go further than that placeholder until four open items below are decided. None of these are code questions — each needs a decision from whoever owns the relevant service boundary or from ARMMAN/product.

### 1. riskConditionId → topicCode mapping — architecture decision needed

No mapping table exists in either risk-referral-service or cms-content-service. `educationContent.client.ts`'s comment referencing "this feature's implementation plan doc" is stale — no such doc exists.

Three options, undecided:

- (a) New join table in cms-content-service, keyed by riskConditionId
- (b) New column on risk_conditions in risk-referral-service, pointing at a topicCode
- (c) Extend the risk-assessment response itself to embed topicCode directly, removing the client-side lookup entirely

Needs: a decision from whoever owns the risk-referral-service ↔ cms-content-service boundary.

### 2. stage enum vocabulary — undefined

Closest existing thing is `RiskCondition.phase` (`REGISTRATION|ANC|DELIVERY|PP|NN|INC|CCV`) — but that's visit phase, not the health-education CSV's free-text Stage column ("as soon as detected", "1st-2nd trimester", "primigravida", etc.), and they don't map 1:1 ("primigravida" is a parity condition, not a phase).

Needs: ARMMAN/product to fix the vocabulary — and a decision on whether stage becomes a field on the content model at all, or whether the risk rule pack's own phase-gating already makes a separate stage field unnecessary (depends on #1).

### 3. Risk-based vs. gestational-age-based message precedence — no rule defined

PRD states "Risk-based messages < Gestational-age based message" but doesn't specify which one displays when both are queued for the same beneficiary/visit/stage. No gestational-age-based message path exists in the codebase yet to even compare against.

[...]

Issue #23, opened 2026-07-16
JWT + Role middleware


Dharanishwaran-ND:
| Field | Value |
|---|---|
| Category | Backend |
| Dependency | Users tables |
| Estimate (days) | 6 |
| Priority | High |
| Deliverable | Authentication |
| Success Criteria | Protected APIs work |

Issue #22, opened 2026-07-16
Prettier,Husky,Es-lint


Dharanishwaran-ND:
| Field | Value |
|---|---|
| Category | Backend |
| Dependency | 0 |
| Estimate (days) | High |
| Priority | Prettier,Husky,Es-lint |
| Deliverable | Prettier,Husky,Es-lint |
| Success Criteria | Completed |

Issue #21, opened 2026-07-16
Configure Logger, CORS, Helmet


Dharanishwaran-ND:
| Field | Value |
|---|---|
| Category | Backend |
| Dependency | Express Setup |
| Estimate (days) | 3 |
| Priority | High |
| Deliverable | Middleware |
| Success Criteria | Middleware active |

PR #201, opened 2026-08-28
feat: Contextual health education content + Learn More placeholder wiring

Dharanishwaran-ND:
## Summary

Wires SRS FR-S-5.2(c) (risk-triggered health education messages) and FR-S-13.4 (contextual in-form Learn More) through to real, tested, live endpoints — using the existing `COMING_SOON` Learn More placeholder where real per-condition/per-field content is still blocked on ARMMAN, rather than leaving both mechanisms as inert, never-exercised schema hooks.

- **`resolveEducationContent` client** (risk-referral-service) — resolves a Learn More topic by code via cms-content-service's `GET /learn-more/topics/:topicCode`. Degrades to `null` on any failure (this is placeholder content, not critical data) rather than throwing.
- **`GET /beneficiaries/:beneficiaryId/risk` now returns `educationContent`** on every risk flag — resolved once per call, attached to every `isEducationTrigger:true` flag, `null` otherwise.
- **`HealthEducationMessage` model + `GET /health-education/messages`** (cms-content-service) — ingests ARMMAN's delivered 32-row health-education CSV (real English content), filterable by `riskConditionId` and/or `stage`. Every row's `bodyMarathi` is a literal "Marathi content coming soon" placeholder (source CSV's own Marathi column is entirely empty) and every row's `riskConditionId` is `null` (manual matching against the real 48-row `risk_conditions` table found no unambiguous match for any of the CSV's ~21 condition names — confirmed, not guessed).
- **Contextual in-form Learn More** (visit-form-service) — `learnMoreTopicCode` (an existing but previously unused schema field) is now set to `COMING_SOON` on 8 curated, clinically-relevant fields across 5 forms (LMP date, BMI, gestational weight gain, feeding concerns, deformity, danger signs), making the FR-S-13.4 contextual-help mechanism actually exercisable for the first time.
- **`/health-education` gateway route** added — found missing during live verification (the endpoint worked directly on the service but 404'd through the real gateway).
- Fixed a stale doc-comment reference and filed issue #200 tracking the remaining open product/architecture decisions (riskConditionId→topicCode mapping strategy, stage vocabulary, message precedence, Marathi/media delivery mechanism) that block the *next* increment (real per-condition content) — not this one.

## Test plan

- [x] `npx tsc --noEmit` clean on cms-content-service, risk-referral-service, visit-form-service
- [x] `npx nx lint` clean on cms-content-service, risk-referral-service, visit-form-service, api-gateway
- [x] `npx nx test` — cms-content-service (9/9), risk-referral-service (231/231), visit-form-service (587/587, incl. 91 seed-data schema-validation tests), api-gateway (67/67)
- [x] Live end-to-end verification through the real gateway with a real JWT: a real JAUNDICE-triggered risk flag correctly resolved the real seeded `COMING_SOON` topic; `GET /health-education/messages?stage=...` returned real, accurate ARMMAN content with the Marathi placeholder; no-filter and `riskConditionId`-filter paths both verified

## Notes

- No migration risk beyond `cms-content-service`'s new `health_education_messages` table (additive-only, applied and verified live this session).
- `visit-form-service`'s seed-data changes are NOT live-published to any environment — this codebase's form-versioning model requires a deliberate publish action; a fresh environment seed (or the admin publish flow) will pick these up.
- Issue #200 tracks what's still needed before real (non-placeholder) per-condition/per-field content can ship.

---

## Appendix — Referral Follow-up Evidence Media: API contract Q&A with frontend (unrelated to this PR's diff)

*The section below is a record of a separate cross-team Q&A with the Sakhi mobile frontend team (Bharath) about the "Referral Follow-up Evidence Media Uploads" feature (health facility photo, Sakhi-beneficiary photo, case paper, discharge summary, investigation report). It does not correspond to any code change in this PR — `media-service` and the referral-evidence-media endpoints already existed prior to this PR and are untouched here. Included per explicit request; tracking for this work should live on its own issue/PR, not this one.*

### Endpoint contracts confirmed to frontend

**Base URL:**
```
<gateway-base-url>/api/v1/media/upload-url    POST
<gateway-base-url>/api/v1/media               POST, GET
<gateway-base-url>/api/v1/media/:id           GET
```

**Step 1 — presigned upload URL:**
```
POST /api/v1/media/upload-url
{ "assetType": "REFERRAL_CASE_PAPER", "mimeType": "image/jpeg", "sizeBytes": 204800 }
```
→ `200`, `data.uploadUrl` (expires in 900s), `data.s3Key`, `data.maxSizeBytes` (25 MB cap).

**Step 2 — direct S3 PUT** of raw file bytes to `uploadUrl`.

**Step 3 — finalize:**
```
POST /api/v1/media
{ "assetType": "REFERRAL_CASE_PAPER", "s3Key": "...", "expectedSizeBytes": 204800, "referralId"/"followupId": "optional-uuid" }
```
→ `201`, narrow response: `{ id, assetType, storageUri, mimeType, sizeBytes, uploadedByUserId, uploadedAt, encryptedFlag, createdAt }`.

**`assetType` enum (verbatim, case-sensitive):** `CONSENT_PHOTO`, `REFERRAL_CASE_PAPER`, `REFERRAL_DISCHARGE_SUMMARY`, `REFERRAL_HEALTH_FACILITY_PHOTO`, `REFERRAL_SAKHI_BENEFICIARY_PHOTO`, `REFERRAL_INVESTIGATION_REPORT`, `LMP_SONOGRAPHY_REPORT`, `TRAINING_PHOTO`, `HEALTH_EDUCATION`, `FAQ`, `REPORT_EXPORT`, `OTHER`.

**Error envelope (real runtime shape):**
```json
{ "success": false, "message": "...", "errorCode": "VALIDATION_ERROR", "traceId": "...", "fieldErrors": {...} }
```

**Attaching media to a referral follow-up:** `POST /referrals/{referralId}/follow-up` accepts `mediaAssetIds: string[]` (max 10) but only at follow-up-creation time — confirmed no route exists to attach media to an already-created follow-up. Frontend's normal flow (finalize media after follow-up creation, using the real `followupId`) is fully supported; no new endpoint needed unless their offline queue requires uploading before the follow-up record exists.

### Evidence-type scope — resolved against the SRS

Checked SRS data points #21/#23 (`docs/Revised_App_Form_Final_20.3.26.xlsx.md`) and FR-S-6.4/Appendix E.5 (`docs/Arogya_Sakhi_SRS_v3.0.md`):

- **#21** is a single bundled **Optional** Image field: "Photo of case paper/discharge summary, health facility, and Sakhi photo with the beneficiary" — one upload slot, not four.
- **FR-S-6.4** (accompanied-referral facility photo tied to Supervisor approval/incentive) is **not a separate requirement** — Appendix E.5 confirms the photo is uploaded "in the follow-up form" itself, i.e. it's the Accompanied-case instance of #21.
- **#23** — investigation report — is the only genuinely separate field: **Mandatory**, conditional on #22 = Yes.

**Decision: build 2 frontend capture slots, not 3 or 5** — one multi-file "Evidence documents" slot (case paper/discharge/facility/Sakhi photo, tagged with the appropriate `assetType` per file) and one conditional "Investigation report" slot. No backend/enum changes required — the 5-value enum remains for backend/reporting granularity; this was purely a frontend UI scoping decision.


apps/risk-referral-service/src/beneficiary-risk/educationContent.client.ts
```diff
@@ -0,0 +1,45 @@
...
+  authorizationHeader: string,
+): Promise<EducationContent | null> {
+  try {
+    const res = await fetch(`${API_GATEWAY_BASE_URL}/api/v1/learn-more/topics/${topicCode}`, {
+      headers: { Authorization: authorizationHeader },
+    });
+    if (!res.ok) return null;
+
+    const body = (await res.json()) as { data: EducationContent };
+    return body.data;
```
palani-ND: **Correctness/contract leak**: `return body.data;` returns the raw cms-content-service response unfiltered and cast to `EducationContent`. `GET /learn-more/topics/:topicCode` actually returns `{id, topicCode, topicName, mediaType, contentUrl, sortOrder}` (see `learnMoreTopicSchema` in `learnMore.routes.ts`), but this file's `EducationContent` interface (and the OpenAPI-documented `educationContent` shape in `beneficiaryRisk.routes.ts`) only declares 4 fields. The extra `id`/`sortOrder` will silently leak into `GET /beneficiaries/:beneficiaryId/risk`'s response, undermining the documented/typed contract.

Suggested fix: destructure only the declared fields before returning, e.g. `const { topicCode, topicName, mediaType, contentUrl } = body.data; return { topicCode, topicName, mediaType, contentUrl };`

apps/risk-referral-service/src/beneficiary-risk/educationContent.client.ts
```diff
@@ -0,0 +1,45 @@
...
+ * beneficiary.client.ts/riskGrade.client.ts in this same service, which
+ * throw badGateway on failure because their data IS load-bearing for the
+ * response (ownership scoping, grade codes).
+ */
+export async function resolveEducationContent(
+  topicCode: string,
+  authorizationHeader: string,
+): Promise<EducationContent | null> {
+  try {
+    const res = await fetch(`${API_GATEWAY_BASE_URL}/api/v1/learn-more/topics/${topicCode}`, {
```
palani-ND, line 41: **Reliability**: no timeout/`AbortController` on this `fetch`. The docstring explicitly promises "a beneficiary's risk profile must stay readable even if cms-content-service is unreachable," but that's only handled for hard failures (network error, non-2xx, bad JSON) via the try/catch. A *slow* (not down) gateway or cms-content-service — under load, GC pause, etc. — will leave this `await` pending indefinitely, hanging `GET /beneficiaries/:beneficiaryId/risk` for every beneficiary with an `isEducationTrigger` flag, which contradicts the stated resilience goal.

(Sibling clients `beneficiary.client.ts`/`riskGrade.client.ts` in this same directory have the same gap, but their data is load-bearing so a hang there is arguably expected; this call is supposed to be the one that degrades gracefully, so the missing timeout undercuts its whole reason for existing.)

apps/risk-referral-service/src/beneficiary-risk/beneficiaryRisk.routes.ts
```diff
@@ -33,6 +42,14 @@ const riskFlagViewSchema = z.object({
   isReferralTrigger: z.boolean(),
   isEducationTrigger: z.boolean(),
   isHrVisitTrigger: z.boolean(),
+  educationContent: educationContentSchema.openapi({
```
palani-ND, line 45: **Altitude**: the documented `educationContent` contract here is locked to `LearnMoreTopic`'s shape — a single nullable `{topicCode, topicName, mediaType, contentUrl}` object. But this same PR's own plan doc (`docs/superpowers/plans/2026-08-28-health-education-content-ingestion.md`) states the real future resolution source for `isEducationTrigger` is the new `HealthEducationMessage` model, which is fundamentally different: an ordered one-to-many sequence per condition (e.g. Anemia's "message 1" then "message 2" at postpartum) with a different field set (`conditionLabel`/`stage`/`messageOrder`/`bodyEn`/`bodyMarathi`/`mediaFile`) that a single nullable object can't represent.

When issue #200's mapping work lands and rewires resolution to `HealthEducationMessage`, this already-published API contract will need a breaking change (singular→array, different fields), not just a data backfill — any consumer (the reference Android app) built against today's shape will break.

apps/risk-referral-service/src/beneficiary-risk/educationContent.client.ts
palani-ND: **Correctness / API contract leak**: this returns `body.data` verbatim and types it as `EducationContent` (only `topicCode`/`topicName`/`mediaType`/`contentUrl`), but the actual JSON from `GET /learn-more/topics/:topicCode` also includes `id` and `sortOrder` (see `learnMoreTopicSchema` in `learnMore.routes.ts`). Since response schemas here are documentation-only (no runtime stripping — see `documented-router.ts`), those two extra fields leak straight through into `GET /beneficiaries/:beneficiaryId/risk`'s `educationContent`, which the OpenAPI doc (`educationContentSchema` in `beneficiaryRisk.routes.ts`) declares as exactly 4 fields. A client that trusts the documented/typed shape (or does a strict equality/snapshot check) will see unexpected extra keys. Consider picking only the 4 fields before returning, e.g. `const { topicCode, topicName, mediaType, contentUrl } = body.data; return { topicCode, topicName, mediaType, contentUrl };`.

apps/risk-referral-service/src/beneficiary-risk/educationContent.client.ts
```diff
@@ -0,0 +1,45 @@
...
+  topicCode: string;
+  topicName: string;
+  mediaType: string;
+  contentUrl: string | null;
+}
+
+/**
+ * Resolves a Learn More topic by its stable code, via cms-content-service's
+ * existing GET /learn-more/topics/:topicCode (through the gateway) — SRS
+ * FR-S-5.2(c)'s "queues a health education message for display", wired to
```
palani-ND, line 20: **Altitude/Reuse**: this comment says the risk-flag integration is "wired to the FR-S-13 Learn More placeholder shell rather than building a second content model." But this same PR *does* build a second content model for the identical SRS requirement (FR-S-5.2(c)): the new `HealthEducationMessage` table/module/seed-data/gateway-route (`apps/cms-content-service/src/health-education/*`). Neither this file nor that module cross-references the other, and issue #200 is cited only from this side.

A future engineer picking up #200's mapping work has no signal that two systems need reconciling. They could plausibly add per-condition `topicCode`s to `LearnMoreTopic` alone, leaving the 32 already-ingested real-content `HealthEducationMessage` rows orphaned and unused — duplicated authoring effort and two half-finished pipelines instead of one converged design.

apps/cms-content-service/src/health-education/healthEducation.routes.ts
```diff
@@ -0,0 +1,77 @@
...
+  messageOrder: z.number().int(),
+  titleEn: z.string().nullable(),
+  bodyEn: z.string(),
+  bodyMarathi: z.string(),
+  mediaType: z.enum(['TEXT', 'IMAGE', 'AUDIO', 'VIDEO']),
+  mediaFile: z.string().nullable(),
+  sortOrder: z.number().int(),
+});
+
+const apiErrorSchema = z.object({
```
palani-ND: **Reuse**: `apiErrorSchema`/`envelope()` are re-declared here (identical shape to `learnMore.routes.ts`'s own copy in the same service) instead of reusing `errorResponse()` from `libs/service-commons/src/docs/error-responses.ts`, which is already used by other services for exactly this purpose and also documents `traceId`/`fieldErrors` — fields missing from this local copy.

Not introduced by this PR (the pattern already exists in `learnMore.routes.ts`), but this PR had the opportunity to consolidate onto the shared helper instead of adding a third copy. A future change to the real error envelope now means hunting down another duplicated schema literal.

apps/cms-content-service/prisma/seed.ts
```diff
@@ -46,9 +47,55 @@ async function seedPlaceholderContent(): Promise<void> {
   });
 }
 
+/**
+ * Ingests ARMMAN's delivered health-education content (2026-08-28,
+ * "Revised App Form Final 20.3.26 - Health education message.csv", 34
```
palani-ND: **Minor/cleanup**: this comment says the source CSV was "34 rows after dropping one genuinely blank spacer row," but `seed-data/health-education-messages.json` actually has 32 rows, and every other reference in this PR (the `HealthEducationMessage` model's doc comments in `schema.prisma`, both plan docs under `docs/superpowers/plans/`) consistently states the source CSV was 32 rows. The "34 rows" figure here is also internally inconsistent (34 − 1 ≠ 32). Likely just a stale/incorrect comment, but worth fixing so a future re-ingestion pass isn't chasing a row-count discrepancy that doesn't exist.

apps/risk-referral-service/src/beneficiary-risk/educationContent.client.ts
palani-ND, line 41: **Reliability**: this `fetch` has no timeout/`AbortSignal` — the surrounding docstring explicitly promises `GET /beneficiaries/:beneficiaryId/risk` "must stay readable even if cms-content-service is unreachable," but that's only true for hard failures (connection refused, non-2xx, malformed body — all handled). A *slow* cms-content-service or gateway hop won't throw or reject; it'll just hang this `await`, and since `getRiskProfile` `await`s this before returning, the whole read blocks indefinitely. Sibling clients in this same directory (`beneficiary.client.ts`, `riskGrade.client.ts`) have the same gap, but their data is load-bearing so a hang there is arguably 'the request should wait'; here the call is explicitly best-effort/non-critical, which is exactly the case that calls for a short timeout (e.g. `AbortSignal.timeout(...)`) so a slow dependency degrades to `null` the same way an unreachable one does.

apps/cms-content-service/src/health-education/healthEducation.repository.ts
```diff
@@ -0,0 +1,23 @@
...
+   * everything" rather than an error.
+   */
+  findMany(filters: { riskConditionId?: string; stage?: string }) {
+    return this.prisma.healthEducationMessage.findMany({
+      where: {
+        isDeleted: false,
+        ...(filters.riskConditionId ? { riskConditionId: filters.riskConditionId } : {}),
+        ...(filters.stage ? { stage: filters.stage } : {}),
+      },
+      orderBy: [{ conditionLabel: 'asc' }, { messageOrder: 'asc' }],
```
palani-ND: **Dead field**: `HealthEducationMessage.sortOrder` is added to the schema (`prisma/schema.prisma`), populated from the CSV in the seed data, and exposed in the API response (`sortOrder: z.number().int()` in `healthEducation.routes.ts`) — but this `orderBy` never references it, only `conditionLabel`/`messageOrder`. Since `messageOrder` already fully determines the intra-condition sequence, `sortOrder` is currently a column + migration + seed value + response field that does nothing. Either wire it into `orderBy` (if cross-condition CSV ordering is meant to matter) or drop it — carrying it unused invites a future reader to assume it controls display order when it doesn't.

apps/cms-content-service/prisma/seed.ts
palani-ND: **Inaccurate comment**: this says the source CSV is "34 rows after dropping one genuinely blank spacer row" (i.e. 33 rows ingested), but `seed-data/health-education-messages.json` actually contains 32 objects, and every other reference to this same ingestion (`schema.prisma`'s `HealthEducationMessage` doc comment, both files under `docs/superpowers/plans/2026-08-28-health-education-content-ingestion.md`) consistently says "32 rows." This comment is the one outlier and looks like a leftover from an earlier draft/count — worth fixing so the ingestion provenance story is consistent.

apps/risk-referral-service/src/beneficiary-risk/educationContent.client.ts
Dharanishwaran-ND: Confirmed valid — verified against learnMoreTopicSchema (extra id/sortOrder fields exist on the real endpoint response). Fixed: resolveEducationContent now destructures only the 4 declared fields before returning.

apps/risk-referral-service/src/beneficiary-risk/educationContent.client.ts
Dharanishwaran-ND: Confirmed valid — no timeout existed. Fixed: added AbortSignal.timeout(3000) to the fetch call, matching your suggested approach. The existing try/catch already handles the resulting AbortError the same way as a network failure, so no other change needed.

apps/risk-referral-service/src/beneficiary-risk/beneficiaryRisk.routes.ts
Dharanishwaran-ND: This was accurate as of the commit this review was against (f20fdc6), but has since been resolved on this same branch by commit d73eb52 ("Resolve real per-condition health-education content", 2026-08-31), which landed before this review was submitted. beneficiaryRisk.service.ts now bridges RiskCondition.conditionCode -> HealthEducationMessage.conditionLabel via CONDITION_CODE_TO_LABEL, calling both resolveEducationContent (COMING_SOON fallback) and the new resolveHealthEducationMessages, and educationContentSchema/riskFlagViewSchema already carry the optional bodyEn field for mapped conditions. No code change needed here — confirming so this doesn't get lost as an open item.

apps/risk-referral-service/src/beneficiary-risk/educationContent.client.ts
Dharanishwaran-ND: Confirmed valid — same root cause as your comment on educationContent.client.ts:41 (this is the identical fix, posted on the same line in a duplicate review pass). Fixed: destructured only the 4 declared EducationContent fields before returning.

apps/risk-referral-service/src/beneficiary-risk/educationContent.client.ts
Dharanishwaran-ND: Same as my reply on beneficiaryRisk.routes.ts:45 — this was accurate as of commit f20fdc6 (what this review was submitted against) but has since been resolved on this branch by d73eb52 (2026-08-31), which landed before the review. beneficiaryRisk.service.ts now explicitly bridges the two systems via CONDITION_CODE_TO_LABEL and calls resolveHealthEducationMessages alongside resolveEducationContent. No code change needed here.

apps/cms-content-service/src/health-education/healthEducation.routes.ts
Dharanishwaran-ND: Confirmed valid. Fixed: added errorResponse re-export to cms-content-service's app.module.ts (matching the convention other services already use), and switched healthEducation.routes.ts's 400/401/403 responses to it, removing the local apiErrorSchema. Left learnMore.routes.ts's own copy untouched since it's a pre-existing, unrelated file outside this PR's diff — happy to file a follow-up to consolidate it too if useful.

apps/cms-content-service/prisma/seed.ts
Dharanishwaran-ND: Confirmed valid — verified the seed data file directly: 32 rows, matching every other reference in this PR. Fixed the comment to say 32.

apps/risk-referral-service/src/beneficiary-risk/educationContent.client.ts
Dharanishwaran-ND: Confirmed valid — same fix as your other comment on this line (duplicate review pass). Fixed to say 32 rows.

Dharanishwaran-ND: Correction to my reply above (posted in error, referencing the wrong thread): this is the timeout comment (duplicate of your comment on this same line in the other review pass). Confirmed valid, fixed the same way — added AbortSignal.timeout(3000) to the fetch call. The wrong text above was meant for the seed.ts row-count thread; apologies for the confusion.

apps/cms-content-service/src/health-education/healthEducation.repository.ts
Dharanishwaran-ND: Confirmed valid — sortOrder does hold meaningful data (the real CSV row order, verified against seed-data/health-education-messages.json) but was never referenced anywhere. Fixed: wired it into orderBy as the outermost tiebreaker, so conditions keep the CSV's authored cross-condition sequence when no conditionLabel filter narrows the result, while conditionLabel/messageOrder still decide grouping and intra-condition order.

apps/cms-content-service/prisma/seed.ts
Dharanishwaran-ND: Confirmed valid — same fix as my other reply on this exact comment/line (duplicate review pass). Fixed the comment to say 32 rows, matching the seed data and every other reference in this PR.

PR #205, opened 2026-09-01
feat(notification-escalation-service,rules-service): system-raised escalations

logeshn-1:
## Summary
- Adds support for system-raised and sync-delay escalations, supervisor fanout, and missing notification types in notification-escalation-service
- Allows the system role to call the escalation rule-set evaluate endpoint in rules-service
- The feature/transfer-escalation split (module: notification-escalation-service + rules-service)

## Test plan
- [x] `nx run notification-escalation-service:test` — 9 suites / 157 tests passing
- [x] `nx run rules-service:test` — 9 suites / 183 tests passing

### DIFF
--- a/apps/notification-escalation-service/src/escalations/dto/create-escalation-event.dto.ts
+++ b/apps/notification-escalation-service/src/escalations/dto/create-escalation-event.dto.ts
@@ -14,7 +14,13 @@ import { z } from 'zod';
  */
 export const createEscalationEventSchema = z
   .object({
-    beneficiaryId: z.string().uuid(),
+    // Exactly one of beneficiaryId/sakhiUserId must be set — enforced by the
+    // .refine below, not per-field .optional() semantics alone. SYNC_DELAY
+    // is the one escalationType that uses sakhiUserId instead of
+    // beneficiaryId (a sync delay is about a Sakhi/device, not a specific
+    // beneficiary); every other type still requires beneficiaryId.
+    beneficiaryId: z.string().uuid().optional(),
+    sakhiUserId: z.string().uuid().optional(),
     escalationType: z.enum([
       'ANC_2_MISSED',
       'ANC_HR_MISSED',
@@ -42,6 +48,9 @@ export const createEscalationEventSchema = z
     visitsMissedCount: z.number().int().optional(),
     assignedSupervisorId: z.string().uuid().optional(),
   })
-  .strict();
+  .strict()
+  .refine((v) => Boolean(v.beneficiaryId) !== Boolean(v.sakhiUserId), {
+    message: 'Exactly one of beneficiaryId or sakhiUserId must be provided.',
+  });
 
 export type CreateEscalationEventInput = z.infer<typeof createEscalationEventSchema>;
--- a/apps/notification-escalation-service/src/escalations/dto/create-escalation-event.dto.ts
+++ b/apps/notification-escalation-service/src/escalations/dto/create-escalation-event.dto.ts
@@ -46,11 +46,22 @@ export const createEscalationEventSchema = z
     visitId: z.string().uuid().optional(),
     referralId: z.string().uuid().optional(),
     visitsMissedCount: z.number().int().optional(),
-    assignedSupervisorId: z.string().uuid().optional(),
+    // Required, not derived server-side: this service has no roster lookup of
+    // its own, so the raiser (cron/rules process) must resolve and supply the
+    // owning Supervisor. Every row list() serves is scoped by this field, so
+    // an omitted value silently orphans the row from every Supervisor's feed.
+    assignedSupervisorId: z.string().uuid(),
   })
   .strict()
   .refine((v) => Boolean(v.beneficiaryId) !== Boolean(v.sakhiUserId), {
     message: 'Exactly one of beneficiaryId or sakhiUserId must be provided.',
-  });
+  })
+  .refine(
+    (v) => (v.escalationType === 'SYNC_DELAY' ? Boolean(v.sakhiUserId) : Boolean(v.beneficiaryId)),
+    {
+      message:
+        'SYNC_DELAY escalations must provide sakhiUserId; every other escalationType must provide beneficiaryId.',
+    },
+  );
 
 export type CreateEscalationEventInput = z.infer<typeof createEscalationEventSchema>;
--- a/apps/notification-escalation-service/src/escalations/escalation.repository.ts
+++ b/apps/notification-escalation-service/src/escalations/escalation.repository.ts
@@ -49,47 +49,72 @@ export class EscalationRepository {
   }
 
   /**
-   * Raises a new escalation event, always OPEN — status is never a create
-   * input (see createEscalationEventSchema's doc comment).
+   * Raises a new escalation event, always OPEN (status is never a create
+   * input — see createEscalationEventSchema's doc comment) — or, if an OPEN
+   * escalation already exists for the same natural key, returns that row
+   * instead (self-healing its assignedSupervisorId to the incoming value if
+   * it's changed, e.g. after a beneficiary/Sakhi reassignment) so an
+   * automated job re-raising the same missed visit/follow-up/sync-delay on
+   * its next tick no-ops instead of piling up duplicate OPEN rows.
+   *
+   * The natural key varies by how the event was raised (visitId for a
+   * missed visit, referralId for a follow-up, sakhiUserId for a sync
+   * delay) — only the fields actually present on `input` narrow the match,
+   * so e.g. a visit-driven escalation is never matched against one raised
+   * for a different visitId on the same beneficiary.
+   *
+   * The find-then-write is wrapped in a transaction holding a Postgres
+   * advisory lock keyed on the natural key, so two concurrent calls for the
+   * same key serialize instead of both passing the "no existing row" check
+   * and inserting duplicates — there's no DB-level unique constraint
+   * backing this natural key (it's conditional on which fields are
+   * present), so the lock is what actually closes the race.
    */
-  create(input: CreateEscalationEventInput, createdByUserId: string) {
-    return this.prisma.escalationEvent.create({
-      data: {
-        beneficiaryId: input.beneficiaryId ?? null,
-        sakhiUserId: input.sakhiUserId ?? null,
-        escalationType: input.escalationType,
-        visitId: input.visitId ?? null,
-        referralId: input.referralId ?? null,
-        visitsMissedCount: input.visitsMissedCount ?? null,
-        assignedSupervisorId: input.assignedSupervisorId ?? null,
-        status: 'OPEN',
-        createdByUserId,
-      },
-    });
-  }
+  createOrReuseOpen(input: CreateEscalationEventInput, createdByUserId: string) {
+    return this.prisma.$transaction(async (tx) => {
+      const lockKey = [
+        input.escalationType,
+        input.beneficiaryId ?? '',
+        input.sakhiUserId ?? '',
+        input.visitId ?? '',
+        input.referralId ?? '',
+      ].join('|');
+      await tx.$executeRaw`SELECT pg_advisory_xact_lock(hashtext(${lockKey}))`;
 
-  /**
-   * Finds an existing OPEN escalation sharing the same natural key as
-   * `input` — the idempotency guard EscalationService.create checks before
-   * inserting, so an automated job re-raising the same missed
-   * visit/follow-up/sync-delay on its next tick no-ops instead of piling up
-   * duplicate OPEN rows. The natural key varies by how the event was raised
-   * (visitId for a missed visit, referralId for a follow-up, sakhiUserId
-   * for a sync delay) — only the fields actually present on `input` narrow
-   * the match, so e.g. a visit-driven escalation is never matched against
-   * one raised for a different visitId on the same beneficiary.
-   */
-  findOpenDuplicate(input: CreateEscalationEventInput) {
-    return this.prisma.escalationEvent.findFirst({
-      where: {
-        status: 'OPEN',
-        isDeleted: false,
-        escalationType: input.escalationType,
-        ...(input.beneficiaryId ? { beneficiaryId: input.beneficiaryId } : {}),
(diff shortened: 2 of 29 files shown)


palani-ND (commented):
Automated code review. Findings below as inline comments; one more that GitHub can't anchor inline since it points at an unchanged line outside this diff's hunks:

**`missedVisitDetailSchema.visitType` enum (`escalation.routes.ts:89`, unchanged by this PR) is missing `'ANC_POST_EDD'`.** `missed-visit-types.ts` (touched in this PR) newly maps `POST_EDD_MISSED → 'ANC_POST_EDD'` and adds it to `MISSED_VISIT_TYPES`, so `GET /missed-visit-escalations/:id` for one of these rows now actually returns `visitType: 'ANC_POST_EDD'` at runtime — but the documented response enum still only lists `['ANC', 'PP', 'NN', 'INC', 'INC-HR', 'CCV', 'CCV-HR']`. Response schemas here are documentation-only (not runtime-validated), so this won't 500, but any generated/strict client built off the published OpenAPI contract will reject or mishandle the new value. Worth adding `'ANC_POST_EDD'` to that enum in this same PR.

apps/notification-escalation-service/src/escalations/escalation.service.ts
```diff
@@ -61,67 +109,165 @@ export class EscalationService {
...
-   * null if it doesn't exist (or isn't one of the 8 supported card types). */
-  async findById(id: string) {
+  /**
+   * Fetches a single escalation event shaped as an enriched Quick Response
+   * card, or null if it doesn't exist (or isn't one of the 8 supported card
+   * types). Unlike list()'s best-effort enrichment, a beneficiary/Sakhi
+   * lookup failure here propagates — for a single card the enrichment IS
+   * the payload, same philosophy as getEddNearingDetail.
+   */
+  async findById(id: string, authorizationHeader: string) {
```
palani-ND: **IDOR: `findById` has no roster/ownership scoping, unlike `list()`.**

`list()` (line 204) scopes results to `assignedSupervisorId: caller.id` for a non-privileged caller, but `findById` takes only `(id, authorizationHeader)` — no `caller` param at all — and the route (`escalation.routes.ts:186`) gates on `requireRoles('SUPERVISOR', 'MANAGER')` with no ownership check in the controller either.

Any authenticated SUPERVISOR can call `GET /escalation-events/:id` with an id from another supervisor's roster (guessed, enumerated, or observed elsewhere) and get back the full enriched card — beneficiary full name, phone, risk level, Sakhi name/phone — bypassing the exact roster restriction `list()` enforces for the same data. Worth adding the same `assignedSupervisorId` check here before returning the card.

apps/notification-escalation-service/src/escalations/dto/create-escalation-event.dto.ts
```diff
@@ -42,6 +48,9 @@ export const createEscalationEventSchema = z
     visitsMissedCount: z.number().int().optional(),
     assignedSupervisorId: z.string().uuid().optional(),
```
palani-ND: **`assignedSupervisorId` is optional with no writer anywhere in the repo.** I grepped the whole monorepo — no code (cron job, other service, this PR) ever sets this field when creating an escalation; the doc comment in `escalation.repository.ts` attributes population to `missedVisit.job.ts`, which doesn't exist in this codebase.

Since `list()` scopes non-privileged callers by `assignedSupervisorId: caller.id` (an exact match against a column that's `NULL` on every real row today, with no backfill), every SUPERVISOR's `GET /escalation-events` will return an empty page for real once this ships — silently breaking the Quick Response feed for every Supervisor, not just new rows. Consider making this field required on the DTO (or deriving it server-side from the beneficiary/Sakhi's roster) rather than leaving it as an easily-forgotten optional.

apps/notification-escalation-service/src/escalations/dto/create-escalation-event.dto.ts
```diff
@@ -42,6 +48,9 @@ export const createEscalationEventSchema = z
...
   })
-  .strict();
+  .strict()
+  .refine((v) => Boolean(v.beneficiaryId) !== Boolean(v.sakhiUserId), {
```
palani-ND, line 56: **This only checks that exactly one of `beneficiaryId`/`sakhiUserId` is set — not that it's the *right* one for `escalationType`.** Per the comment above, SYNC_DELAY should carry `sakhiUserId` and every other type should carry `beneficiaryId`, but nothing here enforces that correlation — `{escalationType: 'ANC_2_MISSED', sakhiUserId: '<uuid>'}` passes.

That row then persists with `beneficiaryId: null`. `requireBeneficiaryId()` in `escalation.service.ts` is called unguarded (not inside the best-effort try/catch) from `list()`'s `enrichRows`/card-assembly and from `findById()` — so this one bad row throws synchronously and 500s the *entire* paginated list for every Supervisor, not just that row, until it's manually cleaned up. Suggest a second `.refine` tying `escalationType === 'SYNC_DELAY'` to `sakhiUserId` and everything else to `beneficiaryId`.

apps/notification-escalation-service/src/escalations/escalation.service.ts
```diff
@@ -61,67 +109,165 @@ export class EscalationService {
...
+   * here, unlike closures/reopen-requests/referrals.
+   *
+   * Idempotent: if an OPEN escalation already exists for the same natural
+   * key (see findOpenDuplicate), that existing row is returned instead of
+   * inserting a duplicate — a cron job re-raising the same missed
+   * visit/follow-up/sync-delay on its next tick (before a Supervisor has
+   * resolved the first one) must not pile up repeat OPEN escalations.
    */
-  create(input: CreateEscalationEventInput, createdByUserId: string) {
+  async create(input: CreateEscalationEventInput, createdByUserId: string) {
```
palani-ND: **Check-then-insert race: `findOpenDuplicate` + `create` aren't atomic, and there's no DB unique constraint backing the natural key** (confirmed — no unique index on `(escalation_type, beneficiary_id, visit_id, ...)` in the schema/migrations). The docstring above says this guard exists specifically so a retried/racing cron tick doesn't pile up duplicate OPEN escalations, but two concurrent `create()` calls (e.g. a client-side timeout + retry while the first request is still in flight) can both pass `findOpenDuplicate` before either insert lands, producing exactly the duplicate rows this is meant to prevent. Wrapping both calls in a transaction, or adding a partial unique index and using it via upsert/`ON CONFLICT DO NOTHING`, would close this.

apps/notification-escalation-service/src/escalations/escalation.repository.ts
```diff
@@ -54,6 +68,31 @@ export class EscalationRepository {
...
+   * `input` — the idempotency guard EscalationService.create checks before
+   * inserting, so an automated job re-raising the same missed
+   * visit/follow-up/sync-delay on its next tick no-ops instead of piling up
+   * duplicate OPEN rows. The natural key varies by how the event was raised
+   * (visitId for a missed visit, referralId for a follow-up, sakhiUserId
+   * for a sync delay) — only the fields actually present on `input` narrow
+   * the match, so e.g. a visit-driven escalation is never matched against
+   * one raised for a different visitId on the same beneficiary.
+   */
+  findOpenDuplicate(input: CreateEscalationEventInput) {
```
palani-ND: **`findOpenDuplicate`'s natural key doesn't include `assignedSupervisorId`, so a reassignment can strand the escalation with the old supervisor.** If a beneficiary/Sakhi is reassigned to a new supervisor after an OPEN escalation was first raised, and the upstream raiser re-triggers the same escalation on its next tick with the *new* `assignedSupervisorId`, this match finds the existing row (same type/beneficiaryId/etc., `assignedSupervisorId` not part of the key) and returns it unchanged — still carrying the old supervisor's id. Combined with `list()`'s scoping, the card becomes invisible to the new supervisor and stuck under the old one until manually resolved.

apps/notification-escalation-service/src/escalations/escalation.service.ts
```diff
@@ -61,67 +109,165 @@ export class EscalationService {
...
+    rows: EscalationEventRow[],
+    authorizationHeader: string,
+  ): Promise<Map<string, RowEnrichment>> {
+    const beneficiaryIds = [...new Set(rows.map((row) => requireBeneficiaryId(row)))];
+    const beneficiaries = new Map(
+      await Promise.all(
+        beneficiaryIds.map(async (id) => {
+          try {
+            return [id, await this.beneficiaryClient.getById(id, authorizationHeader)] as const;
+          } catch {
```
palani-ND: **Best-effort lookup failures here (and the mirrored `catch` at line 151 for Sakhi lookups) are swallowed with zero logging**, unlike every other best-effort failure path added elsewhere in this PR (the CLOSE/TRANSFER notify failures and `fanOutToSupervisor` all `console.error` before swallowing). If beneficiary-service or auth-service starts erroring, every Supervisor's `GET /escalation-events` page silently degrades to blank names/phones/risk-levels with no trace in this service's logs — the outage is invisible to on-call even though it's actively affecting every list request. Worth a `console.error` (or the shared logger) in both catches before returning `null`.

apps/notification-escalation-service/src/escalations/escalation.service.ts
```diff
@@ -61,67 +109,165 @@ export class EscalationService {
...
+    return {
+      cardId: row.id,
+      cardType,
+      cardSource: 'escalation_events' as const,
+      beneficiaryId: row.beneficiaryId,
+      beneficiaryName: enrichment?.beneficiary?.pii.fullName ?? null,
+      beneficiaryPhone: enrichment?.beneficiary?.pii.mobileNumber ?? null,
+      riskLevel: enrichment?.beneficiary?.riskLevel ?? null,
+      assignedSupervisorId: row.assignedSupervisorId,
+      sakhiId: enrichment?.beneficiary?.sakhiId ?? null,
```
palani-ND: **`??` doesn't normalize the empty-string `sakhiId` case.** Per the test suite, beneficiary-service can return `sakhiId: ''` to mean "unassigned" — the rest of this file treats that correctly as falsy (`enrichRows`'s `sakhiIds` filter uses `Boolean(id)`, and the sakhi-lookup gate above uses `beneficiary?.sakhiId ? ... : null`). But here `??` only replaces `null`/`undefined`, so an unassigned beneficiary ships `"sakhiId": ""` in the response while `sakhiName`/`sakhiContact` correctly come back `null` — an inconsistent representation of "no Sakhi" for a field the route schema documents as nullable. Consider `beneficiary?.sakhiId || null` (or reuse the same truthiness check used elsewhere in this file) here too.

apps/notification-escalation-service/src/escalations/escalation.service.ts
logeshn-1: Fixed in 77b876a — findById now takes the caller and applies the same assignedSupervisorId roster check list() uses, returning null (404) for a non-owning SUPERVISOR. MANAGER/ADMIN remain unrestricted.

apps/notification-escalation-service/src/escalations/dto/create-escalation-event.dto.ts
logeshn-1: Fixed in 77b876a — assignedSupervisorId is now required on the create DTO rather than optional, since nothing derives it server-side and every list() call scopes on it.

apps/notification-escalation-service/src/escalations/dto/create-escalation-event.dto.ts
logeshn-1: Fixed in 77b876a — added a second .refine correlating escalationType with the right id field: SYNC_DELAY requires sakhiUserId, every other type requires beneficiaryId.

apps/notification-escalation-service/src/escalations/escalation.service.ts
logeshn-1: Fixed in 77b876a — replaced the separate findOpenDuplicate + create calls with EscalationRepository.createOrReuseOpen, which wraps the find-then-write in a transaction holding a Postgres advisory lock keyed on the natural key, closing the race.

apps/notification-escalation-service/src/escalations/escalation.repository.ts
logeshn-1: Fixed in 77b876a — createOrReuseOpen now compares the existing row's assignedSupervisorId to the incoming value and updates it in place when it's changed, so a reassignment no longer strands the escalation with the old supervisor.

apps/notification-escalation-service/src/escalations/escalation.service.ts
logeshn-1: Fixed in 77b876a — both catches (beneficiary and sakhi lookups) now console.error the id and error before returning null.

apps/notification-escalation-service/src/escalations/escalation.service.ts
logeshn-1: Fixed in 77b876a — switched from ?? to || for sakhiId so an empty-string 'no Sakhi' value from beneficiary-service normalizes to null like the rest of this file already treats it.

logeshn-1:
Addressed all review findings in 77b876a, including the unanchored one on `missedVisitDetailSchema.visitType` (`escalation.routes.ts:89`) — added `'ANC_POST_EDD'` to the documented response enum so it matches what `missed-visit-types.ts`'s new mapping actually returns.

Summary of fixes:
- IDOR on `findById` — now roster-scoped like `list()`
- `assignedSupervisorId` made required on the create DTO (was optional with no writer anywhere)
- Added a `.refine` correlating `escalationType` with the correct id field (SYNC_DELAY → sakhiUserId, else → beneficiaryId)
- Replaced the check-then-insert race with a transaction + advisory lock (`createOrReuseOpen`)
- Reassignment self-heal: a duplicate match now updates `assignedSupervisorId` if it's changed instead of stranding the row
- Logged the two previously-silent beneficiary/Sakhi lookup failures
- Normalized empty-string `sakhiId` to `null` (`||` instead of `??`)
- Added `ANC_POST_EDD` to the documented `visitType` response enum

`nx run notification-escalation-service:test` — 9 suites / 165 tests passing.

PR #206, opened 2026-09-01
feat(approval-service,service-commons): quick response enrichment and m2m auth

logeshn-1:
## Summary
- Enriches quick response cards in approval-service with beneficiary/sakhi names and notifies a sakhi's supervisor on approval request creation
- Adds a machine-to-machine auth client and a DB-backed job lock to service-commons, used by approval-service's new enrichment calls
- Adds a fetch timeout wrapper for outbound quick-response client calls
- PR2 of the feature/transfer-escalation split (module: approval-service + libs/service-commons)
- package-lock.json intentionally excluded — its only change (node-cron) belongs to risk-referral-service/visit-form-service, not this module

## Test plan
- [x] `nx run approval-service:test` — 3 suites / 145 tests passing
- [x] `nx run service-commons:test` — 9 suites / 54 tests passing


palani-ND (commented):
Automated code review. Findings below as inline comments.

Two lower-severity process notes that didn't get their own inline comment: `libs/service-commons/src/auth/service-token-client.ts` and `libs/service-commons/src/jobs/job-lock.ts` are ~125 lines of new shared infra (m2m auth token client, DB-backed job lock) with zero consumers anywhere in this repo — reasonable code, but untested by real usage until something actually calls it. Also, `service-token-client.ts:6` reads `process.env.API_GATEWAY_BASE_URL` directly with a silent fallback instead of going through the repo's shared fail-fast `loadConfig`/Zod-validated `appConfig` pattern every other service uses — a malformed value here won't surface until the first token-mint attempt instead of at boot.

apps/approval-service/src/quick-response/quick-response.controller.ts
```diff
@@ -10,16 +10,18 @@ import type { z } from 'zod';
...
 
     getCardDetail: asyncHandler(async (req, res, next) => {
+      if (!req.user) return next(unauthorized());
       const authorizationHeader = req.header('authorization');
       if (!authorizationHeader) return next(unauthorized());
-      res.json(ok(await service.getCardDetail(req.params.cardId, authorizationHeader)));
+      res.json(ok(await service.getCardDetail(req.params.cardId, req.user, authorizationHeader)));
     }),
 
     decide: asyncHandler(async (req, res, next) => {
```
palani-ND, line 27: **`decide()` never checks that the calling SUPERVISOR owns the target card's Sakhi — unlike `list()`/`getCardDetail()` right above, which this same PR just scoped.**

`list`/`getCardDetail` (lines 13-24) now pass the full `req.user` (`CallerScope`) into the service so it can call `resolveOwnSakhiIds`/enforce ownership. `decide` only forwards `req.user.id` (line 34 in `quick-response.service.ts`'s controller call) — `decideApprovalRequestCard` then loads the row via `repository.findById(cardId)` with no ownership check at all, and the route only requires `requireRoles('SUPERVISOR')`.

Concretely: a SUPERVISOR scoped to Sakhis X/Y who gets a 403 from `getCardDetail` on a card raised by Sakhi Z (a different supervisor's roster) can still `POST /quick-response/:cardId/decision` and approve/reject it — mutating real state (LMP write, referral completion, account reactivation, incentive trigger) on a beneficiary/Sakhi outside their scope. This is the write-path counterpart of the read-path restriction this PR just added, and it's the significant gap in an otherwise-correct scoping change.

apps/approval-service/src/quick-response/notification.client.ts
```diff
@@ -14,6 +14,7 @@ export class NotificationClient {
     title: string,
     body: string,
     authorizationHeader: string,
+    linkedEntity?: { linkedEntityType: string; linkedEntityId: string },
   ): Promise<void> {
     let res: Response;
     try {
```
palani-ND, line 21: **This `fetch` has no `AbortSignal` timeout, unlike every other client this PR touched** (`beneficiary.client.ts`, `closure.client.ts`, `escalation.client.ts`, `geography.client.ts`, `referral.client.ts`, `sakhi.client.ts`, `visit.client.ts`, `reopen-request.client.ts` — all now use the new `DOWNSTREAM_FETCH_TIMEOUT_MS` from `fetch-timeout.ts`, added in this same PR specifically because "a saturated downstream service leaves the caller hanging... instead of failing fast with a 502").

This file was touched in this PR (the `linkedEntity` param), so it's the one call in the directory that's inconsistent with the new convention. If notification-escalation-service hangs, every `decide()` call and the new `notifySupervisor()` path (`approvalRequest.service.ts`) awaits this indefinitely instead of failing fast.

apps/approval-service/src/quick-response/lookup.client.ts
```diff
@@ -11,37 +11,69 @@ interface LookupCategory {
...
+  private async fetchCategory(authorizationHeader: string): Promise<LookupCategory | null> {
+    if (this.cached && this.cached.expiresAt > Date.now()) {
+      return this.cached.category;
     }
+    if (this.inflight) return this.inflight;
+
+    this.inflight = (async () => {
+      let res: Response;
+      try {
+        res = await fetch(`${appConfig.API_GATEWAY_BASE_URL}/api/v1/lookups/APPROVAL_STATUS`, {
```
palani-ND: **Same gap as `notification.client.ts`: no `AbortSignal` timeout on this fetch, even though the doc comment above (lines 22-28) specifically calls out that this call "was...sitting on the critical path before the DB query could even start" for every `list()`/`decide()`.** This is the one client in the directory without the new `DOWNSTREAM_FETCH_TIMEOUT_MS` bound — if auth-service hangs on `GET /lookups/APPROVAL_STATUS`, every Quick Response `list()`/`decide()` call blocks indefinitely instead of failing fast with a 502.

apps/approval-service/src/quick-response/lookup.client.ts
```diff
@@ -11,37 +11,69 @@ interface LookupCategory {
...
-    authorizationHeader: string,
-  ): Promise<string | null> {
-    let res: Response;
-    try {
-      res = await fetch(`${appConfig.API_GATEWAY_BASE_URL}/api/v1/lookups/APPROVAL_STATUS`, {
-        headers: { Authorization: authorizationHeader },
-      });
-    } catch {
-      throw badGateway('Unable to resolve APPROVAL_STATUS — auth-service is unreachable.');
+  private cached: { category: LookupCategory; expiresAt: number } | null = null;
```
palani-ND: **This in-process cache conflicts with the repo's own architecture standard.** Root `CLAUDE.md` §15 (Performance & scalability) says: "Stateless services (sessions/cache in Redis, not memory) → horizontal scaling" and "Cache hot, rarely-changing data (rule packs, master data) in Redis with TTL + event-driven invalidation." `APPROVAL_STATUS` is exactly that kind of data, but it's cached here in a private instance field, not Redis. With N replicas behind the load balancer, each holds its own independently-stale copy for up to 5 minutes instead of one shared, consistently-invalidated cache — reintroducing per-instance state the standard explicitly calls out.

apps/approval-service/src/quick-response/quick-response.service.ts
```diff
@@ -104,46 +202,81 @@ export class QuickResponseService {
...
+   * a SUPERVISOR. A SUPERVISOR with no projectId on their scope fails
+   * closed to zero accessible Sakhis rather than making an unscoped call.
+   */
+  private async resolveOwnSakhiIds(
+    caller: CallerScope,
+    authorizationHeader: string,
+  ): Promise<string[] | null> {
+    if (isPrivileged(caller)) return null;
+    if (!caller.projectId) return [];
+    return this.sakhiClient.getOwnSakhiIds(caller.projectId, authorizationHeader);
```
palani-ND: **`resolveOwnSakhiIds`'s call to `sakhiClient.getOwnSakhiIds` is unguarded, so it now hard-fails `list()`/`getCardDetail()` for every SUPERVISOR on any auth-service error** — unlike the best-effort name-lookup helpers added elsewhere in this same PR (`resolveSakhiName`/`resolveBeneficiaryName`, which catch and log). Since this runs inside `list()`'s `Promise.all` (line 224) alongside the status lookup, a transient 500 from `GET /projects/:projectId/sakhis` rejects the whole call and 500s the entire Quick Response list for every SUPERVISOR — a new hard dependency on auth-service availability that didn't exist before this PR's scoping change. Worth deciding deliberately whether a SUPERVISOR should fail-open, fail-closed-to-empty, or 502 here, rather than an unguarded throw.

apps/approval-service/src/quick-response/quick-response.service.ts
```diff
@@ -1028,19 +1205,22 @@ export class QuickResponseService {
...
       }
-      const beneficiary = await this.beneficiaryClient.getById(
-        existing.beneficiaryId,
-        authorizationHeader,
-      );
+      // The beneficiary lookup only needs the already-known beneficiaryId —
+      // it has no dependency on the referral decision's result — so run it
+      // concurrently with referralClient.decide() instead of after it.
+      const [, beneficiary] = await Promise.all([
+        this.referralClient.decide(existing.referralId, 'COMPLETE', authorizationHeader),
```
palani-ND: **Parallelizing this breaks the sequential guarantee the surrounding code relies on.** The comment above says the beneficiary lookup "has no dependency on the referral decision's result," which is true for the beneficiary fetch's *inputs* — but the incentive-trigger catch block just below (a few lines down) is explicitly written on the assumption that by the time it runs, `referralClient.decide(..., 'COMPLETE')` has *already succeeded* ("the referral marked COMPLETED, but the incentive trigger failed... needs manual follow-up").

With `Promise.all`, if `beneficiaryClient.getById` rejects (network error, not just a `null`/not-found return) while `decide()` is still in flight, `Promise.all` rejects immediately with the beneficiary error — but `decide()` was already fired and may still complete successfully in the background, unobserved. The caller gets an error response implying the whole decision failed, while the referral is silently marked COMPLETE with no incentive trigger, no Sakhi notification, and — unlike the incentive-failure path just below — no log line flagging it for manual follow-up. Consider awaiting `decide()` first (or at least logging if the beneficiary fetch fails after `decide()` has already resolved) before treating this as safe to parallelize.

apps/approval-service/src/approvals/approvalRequest.service.ts
```diff
@@ -1,15 +1,100 @@
...
+  /** Same as findByClosureId, for REOPEN cards. */
+  async findByReopenRequestId(reopenRequestId: string) {
+    const row = await this.repository.findByReopenRequestId(reopenRequestId);
+    if (!row) throw notFound('No approval request found for this reopen request.');
+    return row;
+  }
+
+  async create(dto: CreateApprovalRequestInput, authorizationHeader: string) {
+    const created = await this.repository.create(dto);
+    await this.notifySupervisor(created.id, dto, authorizationHeader);
```
palani-ND: **This synchronously blocks the HTTP response on `notifySupervisor`, which contradicts the doc comment on that method** ("the approval request row is already committed by the time this runs and must not be rolled back or have its creation blocked by this"). The row isn't *rolled back* by a `notifySupervisor` failure (the try/catch swallows it), but `create()` still fully `await`s it before returning — so `POST /approvals` doesn't respond until up to 3 chained downstream calls (Sakhi lookup → beneficiary lookup → notify) finish or fail, combined with `notification.client.ts`'s missing timeout (see other comment) this can hang the request well past the row actually being written. If a client/gateway times out and retries on a slow `notifySupervisor`, the retry creates a second `approval_requests` row for the same submission. Consider firing `notifySupervisor` without awaiting it (or via a background/queue mechanism) so the response returns as soon as the row is committed, matching the doc comment's stated intent.

libs/service-commons/src/jobs/job-lock.ts
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * `jobRun` has no lock/free state of its own — "free" is simply "lockedUntil
+ * is in the past" (or the row doesn't exist yet). Acquiring means either
+ * updating a past-due row to a new future `lockedUntil` (`updateMany`'s
+ * matched-row count is the atomicity guard — 0 rows updated means someone
+ * else already holds it or it isn't due yet) or, on the very first run for a
+ * `jobName`, creating the row — racing that create against another replica
+ * relies on the column's own unique constraint (P2002) to decide the loser.
+ */
+export async function acquireJobLock(
```
palani-ND, line 52: **`acquireJobLock` grants a fixed-duration lease with no renewal/heartbeat**, so a job run that takes longer than `lockDurationMs` lets a second replica re-acquire the same lock while the first is still working. Concretely: replica A is still processing an unusually large batch when `lockedUntil` passes; replica B's next tick sees `lockedUntil < now`, acquires the lock via `updateMany`, and both replicas process the same batch concurrently — the exact double-processing this helper exists to prevent. Currently dormant (no consumer anywhere in the repo yet), but worth flagging now since the fix (extend the lock periodically while the job runs, or size `lockDurationMs` generously and document that job duration must stay well under it) is easier to bake into the API before a first caller shows up and hard-codes an assumption around the current behavior.

apps/approval-service/src/quick-response/quick-response.controller.ts
logeshn-1: Fixed in ec8c3ee — decide() now takes the caller and applies the same resolveOwnSakhiIds roster check getCardDetail() uses, 403ing decideApprovalRequestCard for a non-owning SUPERVISOR (including through the dedicated /lmp-change-requests/:id/decision route, which shares the same code path and had the same gap).

apps/approval-service/src/quick-response/notification.client.ts
logeshn-1: Fixed in ec8c3ee — added signal: AbortSignal.timeout(DOWNSTREAM_FETCH_TIMEOUT_MS).

apps/approval-service/src/quick-response/lookup.client.ts
logeshn-1: Fixed in ec8c3ee — added the same AbortSignal timeout to fetchCategory().

apps/approval-service/src/quick-response/lookup.client.ts
logeshn-1: Fixed in ec8c3ee — removed the in-process cache entirely; fetchCategory() now hits auth-service fresh on every call. This service has no Redis wiring today, so a Redis-backed cache was out of scope for this PR — flagged as the correct long-term fix if this latency becomes a real problem.

apps/approval-service/src/quick-response/quick-response.service.ts
logeshn-1: Fixed in ec8c3ee — wrapped the sakhiClient.getOwnSakhiIds call in try/catch, failing closed to an empty roster (matching the existing !caller.projectId convention) and logging the error, instead of hard-failing the whole list()/getCardDetail() call.

apps/approval-service/src/quick-response/quick-response.service.ts
logeshn-1: Fixed in ec8c3ee — reverted to sequential: await referralClient.decide() first, then beneficiaryClient.getById(). Also added a test asserting beneficiaryClient.getById is never called if referralClient.decide() rejects.

apps/approval-service/src/approvals/approvalRequest.service.ts
logeshn-1: Fixed in ec8c3ee — create() now fires notifySupervisor without awaiting it (void this.notifySupervisor(...)). Safe since the whole method body is already wrapped in try/catch and never rejects.

libs/service-commons/src/jobs/job-lock.ts
logeshn-1: Acknowledged — no real consumer exists yet to build proper heartbeat/renewal against, so building one now would be speculative. Strengthened the JSDoc in ec8c3ee into an explicit, load-bearing caller contract: lockDurationMs must exceed the job's real worst-case runtime, and a future caller processing a highly variable-sized batch should build its own renewal on top rather than assume a fixed duration is safe.

logeshn-1:
Addressed all 8 review findings in ec8c3ee.

On the two process notes from the review body (not inline comments):
- **service-commons auth/job-lock, zero consumers** — acknowledged, no code change (your own framing: reasonable code, just untested by real usage yet).
- **service-token-client.ts reads process.env directly** — this is a shared lib and can't depend on any one service's Zod-validated appConfig without breaking the forklift rule. Added eager validation in the constructor instead (throws immediately when a consuming service instantiates the client, rather than deferring to the first getToken() call) — closer to fail-fast without coupling the lib to a service-specific config schema.

Summary of the 8 fixes:
- Write-path IDOR: `decide()` now roster-scoped like `getCardDetail()` (also closes the same gap on the dedicated `/lmp-change-requests/:id/decision` route, which shares the code path)
- Added missing `AbortSignal` timeouts to `notification.client.ts` and `lookup.client.ts`
- Removed the in-process `LookupClient` cache (no Redis wiring exists yet in this service — flagged Redis as the correct long-term fix)
- `resolveOwnSakhiIds` now fails closed to an empty roster on an auth-service error instead of hard-failing `list()`/`getCardDetail()`
- Reverted `decideAccompaniedReferralCard` to sequential referral-decide → beneficiary-fetch, restoring the guarantee the incentive-trigger catch depends on
- `notifySupervisor` is now fire-and-forget instead of blocking the HTTP response
- Documented `job-lock.ts`'s fixed-lease contract explicitly

`nx run approval-service:test` — 5 suites / 162 tests passing.
`nx run service-commons:test` — 10 suites / 58 tests passing.

PR #208, opened 2026-09-01
fix: Seed ANC/Infant RISK rule packs and grade Sickle Cell Disease

### LINKED ISSUE #207: ANC/Infant RISK rule packs and FormDefinition.riskRuleSetId are never seeded
## Problem

\`rules-service/prisma/seed.ts\` seeds SCHEDULE and ESCALATION rule packs (fixed UUIDs, upsert-once pattern) but never seeds the two RISK-category clinical grading packs — \`anc-risk.rulesJson.ts\` and \`infant-risk.rulesJson.ts\`. These packs exist only in source code; nothing creates their \`RuleSet\`/\`RuleVersion\` rows on a fresh environment.

Separately, \`visit-form-service/prisma/seed.ts\`'s \`FormDefinition\` rows never set \`riskRuleSetId\` — even if a RISK RuleSet existed, no form would be wired to it.

Net effect: a fresh environment has **no working ANC or Infant clinical risk grading at all**. \`POST /forms/ANC_VISIT/submissions\` silently skips the risk-assessment trigger (\`form.service.ts\`'s \`if (dto.visitId && version.formDefinition.riskRuleSetId)\` guard is never true), with no error — `triggerRiskAssessment` is best-effort and only \`console.warn\`s on failure, so this is invisible without directly inspecting the DB or manually calling \`POST /risk-assessments\`.

Discovered while investigating a "High Risk Visit schedule not generated" bug report: the SCD (Sickle Cell Disease) grading fix in \`anc-risk.rulesJson.ts\` had zero effect in production until this gap was found — the ANC risk RuleSet there exists only because it was bootstrapped ad-hoc via admin API calls in the past, under an untracked UUID, disconnected from the seed script entirely.

## Fix

- Add \`RISK_RULE_PACKS\` (ANC + Infant) to \`rules-service/prisma/seed.ts\`, fixed UUIDs, same upsert-once pattern as \`SCHEDULING_RULE_PACKS\`/\`ESCALATION_RULE_PACKS\`.
- Wire \`riskRuleSetId\` on \`ANC_VISIT\`, \`NEONATAL_VISIT\`, \`INC_VISIT\`, \`CCV_VISIT\` in \`visit-form-service/prisma/seed.ts\` to those fixed ids.
- Does not retroactively fix already-seeded environments (dev/prod) — those need a manual one-time backfill.

Dharanishwaran-ND:
## Summary

- rules-service's `prisma/seed.ts` never created the ANC/Infant RISK RuleSets — only SCHEDULE and ESCALATION packs were seeded. A fresh environment had **no working clinical risk grading at all**; the RuleSets in dev/prod only existed because someone bootstrapped them ad-hoc via the admin API, under untracked UUIDs, disconnected from the seed script.
- visit-form-service's `prisma/seed.ts` never wired `FormDefinition.riskRuleSetId` on any form — even with a RuleSet seeded, nothing pointed at it.
- Adds fixed-UUID `RISK_RULE_PACKS` (ANC + Infant) to rules-service's seed, matching the existing `SCHEDULING_RULE_PACKS`/`ESCALATION_RULE_PACKS` upsert-once pattern, and wires `ANC_VISIT`/`NEONATAL_VISIT`/`INC_VISIT`/`CCV_VISIT` to them. `MOTHER_REGISTRATION`/`CHILD_REGISTRATION`/`DELIVERY_VISIT`/`POSTPARTUM_VISIT` are deliberately left unwired — no REGISTRATION/DELIVERY/PP rule pack exists in code yet.
- Also grades **Sickle Cell Disease** (SEVERE, interim measure — see `anc-risk.rulesJson.ts`'s doc comment, pending #191's tier confirmation) at the beneficiary's first ANC_VISIT, merged in from the MOTHER_REGISTRATION answer via `form.service.ts`. Moves the `SICKLE_CELL_DISEASE` risk-referral-service seed row from `REGISTRATION` to `ANC` phase to match — it was previously seeded under a phase no risk-grading evaluation ever runs against, so it was silently never graded, which is the original bug this was found while investigating ("High Risk Visit schedule not generated after SCD detection at enrollment").

Discovered while fixing that bug: the SCD grading fix alone had zero effect until this seeding gap was found and fixed — publishing new rule content is meaningless if nothing wires the RuleSet to the form that submits against it.

Closes #207

## Test plan

- [x] `nx test rules-service` — 190/190 pass
- [x] `nx test visit-form-service` — 589/589 pass
- [x] `nx test risk-referral-service` — 234/234 pass
- [x] `nx lint` clean on all three services
- [x] `tsc --noEmit` clean on rules-service and visit-form-service
- [ ] Seed scripts not run against a live fresh DB in this PR (only the shared dev environment was available, which already has ad-hoc RISK RuleSets under different UUIDs from prior manual bootstrapping — running seed there would add parallel rows rather than cleanly validate the fresh-environment path). Recommend validating against a genuinely fresh DB before/at merge.
- [ ] Does **not** retroactively fix already-seeded dev/prod databases (seed is idempotent-skip) — those need a one-time manual backfill of `RiskCondition.phase` and `FormDefinition.riskRuleSetId`, tracked separately.

### DIFF
--- a/apps/rules-service/prisma/seed.ts
+++ b/apps/rules-service/prisma/seed.ts
@@ -8,6 +8,8 @@ import { ccvRulesJson } from '../src/rules/scheduling/ccv.rulesJson';
 import { hrRulesJson } from '../src/rules/scheduling/hr.rulesJson';
 import { deliveryRulesJson } from '../src/rules/scheduling/delivery.rulesJson';
 import { escalationRulesJson } from '../src/rules/scheduling/escalation.rulesJson';
+import { ancRiskRulesJson } from '../src/rules/scheduling/anc-risk.rulesJson';
+import { infantRiskRulesJson } from '../src/rules/scheduling/infant-risk.rulesJson';
 
 const prisma = new PrismaClient();
 
@@ -147,6 +149,65 @@ async function seedEscalationRulePacks(): Promise<void> {
   }
 }
 
+// The two RISK-category clinical grading packs (ANC High-Risk grading /
+// Infant High-Risk grading — see anc-risk.rulesJson.ts and
+// infant-risk.rulesJson.ts's own doc comments). Previously created only via
+// out-of-band POST /admin/rules calls in every environment, never by this
+// seed script — meaning a fresh environment's ANC_VISIT/NEONATAL_VISIT/
+// INC_VISIT/CCV_VISIT submissions had no risk-grading RuleSet to point at
+// (visit-form-service's FormDefinition.riskRuleSetId had nothing to
+// reference) until someone manually bootstrapped one. Fixed UUIDs (distinct
+// '55555555...' prefix from SCHEDULE's '33333333...' / ESCALATION's
+// '44444444...') so visit-form-service's seed (below) can wire
+// FormDefinition.riskRuleSetId to a known id rather than a value discovered
+// after the fact. Same upsert-once pattern as the packs above: `update: {}`
+// only ever creates these on a fresh environment — an existing environment's
+// already-published RuleVersion is never silently rewritten by re-seeding;
+// a real content change is a new versionNo via POST /admin/rules/:setId/publish.
+const RISK_RULE_PACKS = [
+  {
+    ruleSetId: '55555555-5555-4555-8555-555555555551',
+    ruleVersionId: '55555555-5555-4555-8555-555555555552',
+    ruleSetName: 'Arogya Sakhi ANC Clinical Risk Grading',
+    rulesJson: ancRiskRulesJson,
+  },
+  {
+    ruleSetId: '55555555-5555-4555-8555-555555555561',
+    ruleVersionId: '55555555-5555-4555-8555-555555555562',
+    ruleSetName: 'Arogya Sakhi Infant Clinical Risk Grading',
+    rulesJson: infantRiskRulesJson,
+  },
+] as const;
+
+async function seedRiskRulePacks(): Promise<void> {
+  for (const pack of RISK_RULE_PACKS) {
+    await prisma.ruleSet.upsert({
+      where: { id: pack.ruleSetId },
+      create: {
+        id: pack.ruleSetId,
+        ruleCategory: 'RISK',
+        ruleSetName: pack.ruleSetName,
+        status: 'ACTIVE',
+      },
+      update: {},
+    });
+
+    await prisma.ruleVersion.upsert({
+      where: { id: pack.ruleVersionId },
+      create: {
+        id: pack.ruleVersionId,
+        ruleSetId: pack.ruleSetId,
+        versionNo: 'v1',
+        rulesJson: pack.rulesJson,
+        effectiveFrom: new Date('2026-08-10'),
+        checksum: createHash('sha256').update(JSON.stringify(pack.rulesJson)).digest(),
+        status: 'PUBLISHED',
+      },
+      update: {},
+    });
+  }
+}
+
 /**
  * Seeds the SCHEDULE rule set + its v1-hardcoded published version, so
  * VisitSchedule.generatedByRuleVersionId (NOT NULL, no default) has a real
@@ -204,6 +265,9 @@ async function main(): Promise<void> {
 
   await seedEscalationRulePacks();
   console.log(`Seeded ${ESCALATION_RULE_PACKS.length} escalation rule pack(s).`);
+
+  await seedRiskRulePacks();
+  console.log(`Seeded ${RISK_RULE_PACKS.length} clinical risk-grading rule pack(s) (ANC/Infant).`);
 }
 
 main()
--- a/apps/rules-service/src/rules/scheduling/anc-risk.rulesJson.ts
+++ b/apps/rules-service/src/rules/scheduling/anc-risk.rulesJson.ts
@@ -74,11 +74,30 @@
  * them; not implemented pending ARMMAN confirming whether new form fields
  * are needed (issue #191).
  *
- * Sickle Cell is not read or graded by this pack at all. Appendix D's
- * Anemia table ties Sickle Cell Disease to the Moderate tier (confirmed,
- * issue #191) — not yet implemented since MOTHER_REGISTRATION has no
- * sickle-cell field wired into this pack's input today. Sickle Cell
- * Trait's effect on grading, if any, is unconfirmed.
+ * Sickle Cell Disease (SCD) is graded as an INTERIM MEASURE, pending GitHub
+ * issue #191 (item 10): Appendix D's Anemia table ties SCD to the Moderate
+ * tier, but the app-form spec (Q60) says SCD selected -> Severe risk +
+ * referral — these two sources conflict and ARMMAN has not yet confirmed
+ * which is correct. This pack currently grades SCD as SEVERE, matching the
+ * app-form spec, since that's the more specific source for this exact
+ * question and matches the field's own stated risk action. If issue #191
+ * is answered as Moderate instead, this grading must be revisited. Sickle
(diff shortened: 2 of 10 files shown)


apps/rules-service/src/rules/scheduling/anc-risk.rulesJson.ts
```diff
@@ -221,6 +240,27 @@ const handler = (input, { dayjs }) => {
...
+  // issue #191 (item 10). Sickle Cell Trait (SCT) is deliberately left
+  // ungraded — its impact, if any, is also unconfirmed by #191.
+  // sickleCellStatus is merged in by form.service.ts from
+  // MOTHER_REGISTRATION's SCD/SCT question, read at the beneficiary's
+  // first ANC_VISIT (not at registration itself — see this pack's doc
+  // comment for why registration-time grading isn't wired up).
+  const sickleCellStatus = input.sickleCellStatus;
+  if (typeof sickleCellStatus === 'string') {
+    const grade = sickleCellStatus === 'sickle_cell_disease_scd' ? 'SEVERE' : 'NORMAL';
+    record('SICKLE_CELL_DISEASE', grade, { sickleCellStatus }, {
```
palani-ND, line 265: **This is a realistic, not theoretical, path into `requireConditionId`'s throw for any already-seeded environment (dev/prod) — and the blast radius is bigger than "SCD doesn't get graded."**

I checked `mother-registration.json`'s Q60 (`have_you_been_detected_with_sickle_cell_disease_or_sickle_cell_trait_sct`): it's `"required": true"` with no neutral/empty option — every value (`tested_and_result_is_normal`, `not_tested_yet`, `don_t_know_not_aware`, etc.) is a non-empty string. So `sickleCellStatus` will be present as a string on essentially every first ANC_VISIT once a beneficiary has completed registration under the current form — this `record()` call will fire for nearly all of them, not just SCD-positive cases.

`requireConditionId` (line 138) throws synchronously if `conditionIds['SICKLE_CELL_DISEASE']` is missing, and there's no per-condition try/catch in this `handler` — a throw here aborts the entire function, so every condition that would otherwise be evaluated after this block (Anemia, danger signs, everything below) never gets recorded either. I confirmed `visit-form-service`'s `triggerRiskAssessment` caller only `console.warn`s on a non-2xx response and lets the visit submission itself succeed — so this failure mode is a fully silent "no risk assessment happened at all for this visit," not a visible error.

The PR description already flags that `RiskCondition.phase` needs a manual backfill for already-seeded environments and that this PR's new fixed-UUID `RuleSet` sits unused there until someone repoints `FormDefinition.riskRuleSetId` (see the other comment). What's not called out: **those two steps have to happen in the right order.** If the new rule content (this file, with the SCD block) goes live in an environment *before* the `RiskCondition.phase` backfill runs, every first ANC_VISIT for a registered beneficiary hits this throw — not a rare edge case, but the common case. Worth either (a) making the deploy runbook for the backfill explicit about this ordering, or (b) making `requireConditionId` fail soft (log + skip that one condition) instead of aborting the whole evaluation, so a single missing mapping degrades gracefully instead of losing every condition in the same pack.

apps/rules-service/prisma/seed.ts
```diff
@@ -147,6 +149,65 @@ async function seedEscalationRulePacks(): Promise<void> {
...
+// (visit-form-service's FormDefinition.riskRuleSetId had nothing to
+// reference) until someone manually bootstrapped one. Fixed UUIDs (distinct
+// '55555555...' prefix from SCHEDULE's '33333333...' / ESCALATION's
+// '44444444...') so visit-form-service's seed (below) can wire
+// FormDefinition.riskRuleSetId to a known id rather than a value discovered
+// after the fact. Same upsert-once pattern as the packs above: `update: {}`
+// only ever creates these on a fresh environment — an existing environment's
+// already-published RuleVersion is never silently rewritten by re-seeding;
+// a real content change is a new versionNo via POST /admin/rules/:setId/publish.
+const RISK_RULE_PACKS = [
```
palani-ND, line 167: **These fixed UUIDs won't converge with the ad-hoc `RuleSet`s already live in dev/prod, since `upsert` matches by id.** Per the PR description, dev/prod's current ANC/Infant RISK `RuleSet`s were bootstrapped manually via the admin API under different (untracked) UUIDs. Since `seedRiskRulePacks()` upserts by these new `55555555...` ids, running this seed against dev/prod won't update or replace the existing live RuleSets — it creates a second, parallel, currently-unreferenced `RuleSet`/`RuleVersion` pair alongside them. `visit-form-service`'s seed also only wires `FormDefinition.riskRuleSetId` on `create` (skipped entirely if the form already exists — confirmed via its `if (existing) continue;` guard), so in any already-seeded environment nothing will actually point at this new content until a manual cutover happens.

That's consistent with the PR description's disclosed limitation, but combined with the other comment on `anc-risk.rulesJson.ts`, it means there are two manual steps (backfill `RiskCondition.phase`, then cut over to this new `RuleSet`/republish its content into the existing live one) that must land in that specific order — nothing in code enforces or checks it, so it's easy for whoever does the manual follow-up to do the cutover first and trigger the crash described in the other comment.

palani-ND (commented):
Automated code review. This is a small, well-documented PR (the doc comments already disclose most of the nuance), so just two findings below, both about deployment sequencing risk rather than unit-level correctness — the new SCD-grading logic itself and its tests look correct.

apps/rules-service/src/rules/scheduling/anc-risk.rulesJson.ts
Dharanishwaran-ND: Fixed in 3d0d621 — option (b), as you recommended. `record()` (in both `anc-risk.rulesJson.ts` and `infant-risk.rulesJson.ts`, which had the identical pattern) now returns early when `conditionIds[code]` is missing, instead of throwing via `requireConditionId`. Every other condition's `record()` call is unaffected — confirmed with two new tests: one generic (missing ANEMIA's mapping still grades AGE/HYPERTENSION/etc.), and one reproducing your exact scenario (missing SICKLE_CELL_DISEASE's mapping with an SCD-positive answer still grades everything else).

Also confirmed no console/logging is available in this GoRules sandbox (grepped — no other rule pack in this repo calls `console.*`), so the skip is silent at this layer; visibility into a missing mapping would need to be added at the caller (rules-service's evaluate endpoint) if that's wanted later — flagged in the new code comment but not implemented here, since it's outside this PR's scope.

This doesn't remove the need for getting `RiskCondition.phase` backfill and the new `RuleSet` cutover in the right order (your other comment) — it just means getting that order wrong now degrades to "this one condition isn't graded yet" instead of "no risk assessment happens at all for this visit."

apps/rules-service/prisma/seed.ts
Dharanishwaran-ND: This limitation stands and isn't something I'm trying to fully solve in this PR — it's the same disclosed limitation from the PR description ("Does not retroactively fix already-seeded dev/prod databases... tracked separately"). Actually reconciling the fixed-UUID RuleSet with dev/prod's existing ad-hoc ones would mean either a real migration/cutover mechanism or a manual admin action, and that's a materially bigger scope than this seed-gap fix.

What changes the calculus here is the fix I just pushed (3d0d621) for your other comment: the specific *crash* this ordering hazard could trigger — "cutover happens before the RiskCondition.phase backfill, `requireConditionId` throws, entire grading aborts" — is gone. Doing the two manual steps in the wrong order now means SCD (or whichever condition is unmapped at the time) just doesn't get graded yet, not that no risk assessment happens at all for the visit. So the ordering is still worth getting right, but getting it wrong is no longer silently catastrophic.

I'll leave the actual dev/prod cutover mechanism as a separate tracked follow-up rather than scope-creep this PR into it — happy to open an issue for it if that's useful.

PR #209, opened 2026-09-02
feat(visit-form-service): cron jobs, m2m auth, and job-lock support

logeshn-1:
## Summary
- Adds missed-visit and post-edd visit-generation cron jobs to visit-form-service
- Both jobs are guarded by a DB-backed job lock so multiple replicas never double-process the same tick, and authenticate downstream calls via the new machine-to-machine service token client
- Fixes the shared bootstrap error-swallowing bug across wrapper-api-service, supervisor-operations-service, reporting-etl-service, media-service, and incentive-wages-service
- Includes package-lock.json's node-cron/@types/node-cron entries — this module owns that dependency
- Also carries libs/service-commons/src/jobs/job-lock.ts and auth/service-token-client.ts (review fixes already applied) since these cron jobs depend on both; the approval-service PR carries the same two files, so whichever merges first lands them, the other is a no-op there
- No new review comments expected — this branch already applies the fixes from the equivalent findings raised on the approval-service PR

## Test plan
- [x] `nx run visit-form-service:test` — 30 suites / 607 tests passing (8 pre-existing skips)
- [x] `nx run-many -t test` across wrapper-api-service, supervisor-operations-service, reporting-etl-service, media-service, incentive-wages-service — all passing (`--passWithNoTests` for the one with no test files)
- [x] `nx run-many -t lint` and `-t build` across all 6 modules — clean
- [x] CI `verify` workflow passing on this branch


apps/visit-form-service/src/jobs/missedVisit.job.ts
```diff
@@ -0,0 +1,252 @@
...
+    try {
+      const transitioned = await scheduleRepo.markMissed(schedule.id);
+      if (!transitioned) continue; // raced with another run/manual PATCH — already handled
+
+      if (missedStatusLookupValueId) {
+        await instanceRepo.markMissedByScheduleId(schedule.id, missedStatusLookupValueId);
+      }
+
+      if (!authorizationHeader) continue;
+      await evaluateAndEscalate(schedule, scheduleRepo, deps, authorizationHeader);
```
palani-ND: **[Correctness] Escalation can be silently and permanently lost**

`markMissed()` (line 112) flips the schedule OPEN→MISSED *before* `evaluateAndEscalate` runs here. If `evaluateEscalation`, `findBeneficiaryOwnership`, `findSakhiById`, or `createEscalationEvent` throws (e.g. a transient rules-service/beneficiary-service error), the surrounding try/catch (lines 111-125) only logs it. Since `findOverdueOpenSchedules` only ever selects `status='OPEN'` schedules, this schedule can never be revisited by a future tick — the required HR-missed-visit escalation (FR-S-7.1) for that beneficiary is dropped with no retry and no record beyond a `console.error` line.

Worth considering: don't flip the schedule to MISSED until escalation has succeeded (or persist a durable retry marker), so a transient downstream failure doesn't permanently forfeit the escalation.

apps/visit-form-service/src/jobs/missedVisit.job.ts
```diff
@@ -0,0 +1,252 @@
...
+      `[${JOB_NAME}] Unable to mint a service token — will still transition schedules, ` +
+        'but skip escalation/notification calls this tick:',
+      err,
+    );
+  }
+
+  let missedStatusLookupValueId: string | null = null;
+  if (authorizationHeader) {
+    try {
+      missedStatusLookupValueId = await resolveVisitStatusIdByCode('MISSED', authorizationHeader);
```
palani-ND: **[Correctness] One lookup failure permanently desyncs the whole tick's VisitInstance rows**

`resolveVisitStatusIdByCode` here is called once per tick, outside the per-schedule loop. If it throws (e.g. auth-service/gateway blip), `missedStatusLookupValueId` stays `null`, so the `if (missedStatusLookupValueId)` guard at line 115 skips `instanceRepo.markMissedByScheduleId` for *every one* of the up to 200 schedules processed that tick — even though `scheduleRepo.markMissed` (line 112) already committed each schedule's status to MISSED. Because MISSED schedules are never selected by `findOverdueOpenSchedules` again, those VisitInstance rows are permanently stuck at their prior status with no future tick able to correct them.

Suggest retrying/backing off this one lookup rather than letting a single transient failure silently affect the whole batch permanently.

apps/visit-form-service/src/visits/visitInstance.repository.ts
```diff
@@ -36,6 +36,22 @@ export class VisitInstanceRepository {
...
+   * updateStatus()'s human-caller ownership checks (this is a system
+   * transition, not a user PATCH). `completedAt: null` is the existing
+   * "not completed" proxy this repo already relies on elsewhere (see
+   * findManyByBeneficiaryId's doc comment) — a COMPLETED instance's
+   * completedAt is always set, so this WHERE clause can never overwrite one.
+   */
+  markMissedByScheduleId(scheduleId: string, missedStatusLookupValueId: string) {
+    return this.prisma.visitInstance.updateMany({
+      where: { scheduleId, isDeleted: false, completedAt: null },
+      data: { statusLookupValueId: missedStatusLookupValueId, statusCode: null },
```
palani-ND: **[Correctness] No VisitStatusHistory row written for job-driven MISSED transitions**

`markMissedByScheduleId` flips `visit_instances` to MISSED without writing the `VisitStatusHistory` row that every other transition creates. The existing `updateStatus()` in this same file (further down) creates a `VisitStatusHistory` row in the same transaction for every status change, and `visitInstance.service.ts`'s own doc comment states *'VisitStatusHistory is the record of when the MISSED transition itself happened.'*

Because this new job-only write path bypasses that, any visit auto-transitioned to MISSED by the cron job leaves a permanent gap in the audit trail — a screen or report that relies on VisitStatusHistory to show when/why a visit became MISSED will show nothing for cron-driven transitions.

libs/service-commons/src/jobs/job-lock.ts
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * `jobRun` has no lock/free state of its own — "free" is simply "lockedUntil
+ * is in the past" (or the row doesn't exist yet). Acquiring means either
+ * updating a past-due row to a new future `lockedUntil` (`updateMany`'s
+ * matched-row count is the atomicity guard — 0 rows updated means someone
+ * else already holds it or it isn't due yet) or, on the very first run for a
+ * `jobName`, creating the row — racing that create against another replica
+ * relies on the column's own unique constraint (P2002) to decide the loser.
+ *
+ * The lease is fixed-duration with no renewal/heartbeat: `lockedUntil` is
```
palani-ND, line 40: **[Correctness/Concurrency] Fixed-duration lock lease has no renewal, and this PR's own post-EDD job can plausibly outrun it**

This doc comment warns: *'If the job is still running when `lockedUntil` passes, a second replica's next tick will re-acquire the same lock and run concurrently.'* `postEddVisitGeneration.job.ts` sets `LOCK_DURATION_MS` to 20 minutes but bounds its own work at `MAX_PAGES_PER_TICK=1000` pages of `PAGE_SIZE=200` (up to 200k candidates), each requiring at least one DB existence check plus a full `generateSchedule` call chain. A large-enough backlog (first deploy, or recovery after an outage) can exceed 20 minutes of real runtime, so the lease expires mid-run and a subsequent tick double-processes the same beneficiaries concurrently — the exact race this helper exists to prevent.

Also, this comment's claim *'There is currently no consumer of this function in the repo'* is now false — this very PR wires up two consumers (missedVisit.job.ts, postEddVisitGeneration.job.ts) — looks like a stale comment carried over from wherever this file originated.

apps/visit-form-service/src/jobs/postEddVisitGeneration.job.ts
```diff
@@ -0,0 +1,151 @@
...
+  const scheduleRepo = new VisitScheduleRepository(deps.prisma);
+  const scheduleService = new VisitScheduleService(scheduleRepo);
+
+  const cutoffDate = new Date(Date.now() - EDD_GRACE_DAYS * 24 * 60 * 60 * 1000)
+    .toISOString()
+    .slice(0, 10);
+
+  let cursor: string | undefined;
+  for (let page = 0; page < MAX_PAGES_PER_TICK; page++) {
+    const result = await findPostEddPendingBeneficiaries(
```
palani-ND: **[Correctness] Long-running paginated fetch is unguarded and the token is minted only once up front**

`authorizationHeader` is minted once (further up in this function) and reused for up to 1000 sequential page fetches. If the run takes long enough to exceed the service token's TTL, this call to `findPostEddPendingBeneficiaries` throws (401 → `badGateway`) with no local try/catch around it — only the per-candidate call inside the loop below is guarded. That propagates out of `runPostEddVisitGenerationJob` entirely, and since `cursor` is local to this function call, the next tick restarts pagination from the very first page instead of resuming — any partial progress from the aborted run is lost.

Worth wrapping this call too (to at least log/return instead of throwing), and/or refreshing the token per-page for long runs.

apps/visit-form-service/src/jobs/missedVisit.job.ts
```diff
@@ -0,0 +1,252 @@
...
+      assignedSupervisorId: supervisorId ?? undefined,
+    },
+    authorizationHeader.replace(/^Bearer /, ''),
+  );
+
+  // Best-effort — the escalation event above is the durable record; a failed
+  // push here is logged, not retried, and never fails the job run.
+  if (supervisorId && event.status === 'OPEN') {
+    try {
+      if (mapping.isHrVisit) {
```
palani-ND: **[Simplification] Duplicated notification branches + repeated Bearer-stripping**

The HR vs non-HR branches here duplicate the entire `createNotification` call shape to vary ~5 fields, and `authorizationHeader.replace(/^Bearer /, '')` is recomputed 3 times (here, and at the `createEscalationEvent` call above) for the same value.

This could be table-driven off `VISIT_TYPE_ESCALATION` (notificationType/title/body-template/linked-entity per family) to build one payload and call `createNotification` once, and the Bearer-stripped token computed once and reused. As-is, a future shared-field change (or a new HR family) has to be edited in two near-identical places, and it's easy for one branch to drift from the other without either branch's own tests catching it.

apps/visit-form-service/src/lookups/lookup.client.ts
```diff
@@ -65,3 +65,22 @@ export async function resolveVisitStatusCodes(
...
+
+/**
+ * The reverse of {@link resolveVisitStatusCode} — resolves a VISIT_STATUS
+ * valueCode (e.g. "MISSED") to its lookup_value_id. Needed by the
+ * missed-visit job (missedVisit.job.ts), which must WRITE a
+ * statusLookupValueId, not just read one. Throws if the code isn't a known
+ * VISIT_STATUS value — a misconfigured/renamed lookup category must fail the
+ * job run loudly, not silently skip every visit.
+ */
+export async function resolveVisitStatusIdByCode(
```
palani-ND, line 77: **[Efficiency] No caching of the (effectively static) VISIT_STATUS lookup category**

`resolveVisitStatusIdByCode` fetches the *entire* VISIT_STATUS category from auth-service (through the gateway) via `fetchVisitStatusCategory`, just to find one of 5 static rows. `missedVisit.job.ts` calls this on every tick (default every 30 minutes), so this is a recurring, avoidable network round trip for data that essentially never changes. A short in-process cache/TTL would eliminate nearly all of it.

apps/visit-form-service/src/visit-schedules/visitSchedule.repository.ts
```diff
@@ -58,6 +58,50 @@ export class VisitScheduleRepository {
...
+   * first — the missed-visit job walks this list counting the unbroken
+   * trailing run of MISSED to derive `consecutiveMissedCount` (scoped to
+   * the exact visitType, e.g. ANC vs ANC_HR are counted separately — see
+   * missedVisit.job.ts for why). Bounded to a small window since only the
+   * leading run matters, not the full history.
+   */
+  findRecentByBeneficiaryAndVisitType(beneficiaryId: string, visitType: VisitCodeType) {
+    return this.prisma.visitSchedule.findMany({
+      where: { beneficiaryId, visitType, isDeleted: false },
+      orderBy: { scheduledDate: 'desc' },
```
palani-ND, line 130: **Correctness (high severity):** `findRecentByBeneficiaryAndVisitType` orders purely by `scheduledDate: desc` with no filter excluding not-yet-due future schedules, so `countConsecutiveMissed` in `missedVisit.job.ts` almost always sees future OPEN rows ahead of the just-missed one and returns 0.

**Failure scenario:** ANC schedules are generated as a full batch up front (scheduleMapper.ts's ANC case pushes ANC1..ANCn plus ANC_POST_EDD in one call), so a beneficiary typically has ANC2/ANC3/ANC4 sitting OPEN with a future scheduledDate while ANC1's window has already closed. When missedVisit.job.ts marks ANC1 MISSED and calls findRecentByBeneficiaryAndVisitType(beneficiaryId, 'ANC') (ordered scheduledDate desc, take 20), it gets ANC4, ANC3, ANC2 (still OPEN) before ANC1 (MISSED). countConsecutiveMissed (missedVisit.job.ts:51-58) breaks on the first non-MISSED row and returns 0 for a schedule that just became MISSED. Since evaluateEscalation is driven by consecutiveMissedCount, HR/consecutive-count escalation logic silently never fires as designed for any beneficiary with pre-generated future visits — the normal case for ANC/PP/INC.

apps/visit-form-service/src/visit-schedules/visitSchedule.service.ts
```diff
@@ -19,9 +19,20 @@ export interface CallerIdentity {
...
+ * identity, which has no Sakhi/Supervisor roster of its own to check
+ * against, same reasoning as escalation.routes.ts/notification.routes.ts
+ * already granting SYSTEM their ADMIN-only endpoints.
+ */
 function isPrivileged(caller: CallerIdentity): boolean {
-  return caller.roles.includes('MANAGER') || caller.roles.includes('ADMIN');
+  return (
+    caller.roles.includes('MANAGER') ||
+    caller.roles.includes('ADMIN') ||
+    caller.roles.includes('SYSTEM')
```
palani-ND, line 34: **Altitude (medium-low severity):** `isPrivileged()` special-cases the SYSTEM role only in this one service's copy of the check; at least 7 other services define their own independent `isPrivileged`/authorization function that doesn't yet include SYSTEM.

**Failure scenario:** a future job/feature elsewhere (beneficiary-service, auth-service, supervisor-operations-service, sync-service, etc.) needing the same service-token/SYSTEM bypass must hand-copy this exact change into each service's own isPrivileged; missing one silently denies a legitimate SYSTEM caller, while an inconsistent copy elsewhere risks an unintended bypass, with no shared mechanism enforcing consistency across services.

apps/visit-form-service/src/jobs/missedVisit.job.ts
```diff
@@ -0,0 +1,252 @@
...
+      err,
+    );
+  }
+
+  let missedStatusLookupValueId: string | null = null;
+  if (authorizationHeader) {
+    try {
+      missedStatusLookupValueId = await resolveVisitStatusIdByCode('MISSED', authorizationHeader);
+    } catch (err) {
+      console.error(`[${JOB_NAME}] Unable to resolve the MISSED lookup value id:`, err);
```
palani-ND: **Conventions:** this file (and `postEddVisitGeneration.job.ts`) logs exclusively via raw `console.error`/`console.log` instead of the shared pino logger, violating `.claude/CLAUDE.md` §7: "Structured JSON via the shared logger (pino). Levels: error/warn/info/debug."

Every operational log line from these two new, recurring cron jobs (lock-skip notices, token-mint failures, per-schedule failures, escalation/notification failures) is unstructured console output with no requestId/JSON structure, unlike the rest of the codebase's request-path logging.

apps/visit-form-service/src/jobs/missedVisit.job.ts
```diff
@@ -0,0 +1,252 @@
...
+
+  const event = await createEscalationEvent(
+    {
+      beneficiaryId: schedule.beneficiaryId,
+      escalationType: mapping.escalationType,
+      visitId: schedule.id,
+      visitsMissedCount: consecutiveMissedCount,
+      assignedSupervisorId: supervisorId ?? undefined,
+    },
+    authorizationHeader.replace(/^Bearer /, ''),
```
palani-ND: **Simplification:** `authorizationHeader.replace(/^Bearer /, '')` is repeated three times in `evaluateAndEscalate` (lines 164, 192, 203) to undo the "Bearer " prefix for each call to `createEscalationEvent`/`createNotification`.

Simpler form: compute `const rawToken = authorizationHeader.replace(/^Bearer /, '')` once at the top of evaluateAndEscalate (or have those two clients accept the full Authorization header like every other *.client.ts in this codebase) instead of re-deriving the same value three times.

apps/visit-form-service/src/jobs/missedVisit.job.ts
```diff
@@ -0,0 +1,252 @@
...
+
+/**
+ * `VisitCodeType` -> the ESCALATION rule pack's (visitFamily, isHrVisit)
+ * input shape (see escalation.rulesJson.ts) and the escalationType code
+ * POST /escalation-events expects for a missed visit in that family. DELIVERY
+ * has neither an escalation-rule branch nor a "missed" escalationType — it's
+ * the ARMMAN open question in the build plan (whether DELIVERY_FORM_PENDING
+ * needs its own earlier threshold) and is deliberately left unhandled here.
+ */
+const VISIT_TYPE_ESCALATION: Record<
```
palani-ND, line 29: **Altitude:** the `escalationType`/`visitFamily` mapping per `VisitCodeType` is hardcoded here in application code, duplicating logic that conceptually belongs alongside rules-service's SCHEDULE rule pack, which already owns the analogous per-visit-type behavior.

This table's own comment admits PP_HR/NN_HR entries are "ready" for a rule pack that doesn't emit them yet — a rules-service change (e.g. renaming a visitName or adding a new HR family) requires a matching hand-edit + deploy of this file, and the two can silently drift with no compile-time or runtime check tying them together.

logeshn-1:
Addressed in a8f3f0c:

**Correctness fixes:**
- `missedVisit.job.ts`: the service token and MISSED lookup id are now resolved once up front, aborting the whole tick on failure instead of transitioning every schedule to MISSED while silently skipping their side effects. Any failure after a schedule is flipped to MISSED (VisitInstance write, escalation lookup/evaluation) now reverts it back to OPEN so the next tick retries it — no more permanently-lost FR-S-7.1 escalations.
- `visitSchedule.repository.ts`: added `revertToOpen()`; `findRecentByBeneficiaryAndVisitType` now excludes not-yet-due schedules (`windowEndDate <= asOf`) so pre-generated future OPEN rows no longer outrank the just-missed row and zero out `consecutiveMissedCount`.
- `visitInstance.repository.ts`: `markMissedByScheduleId` now writes a `VisitStatusHistory` row per transitioned instance in the same transaction, matching `updateStatus()`'s existing convention.
- `postEddVisitGeneration.job.ts`: the service token is now minted fresh per page (the client caches/refreshes internally) instead of once for the whole run, and a page-fetch failure now stops the tick cleanly instead of throwing and losing all prior progress.
- `job-lock.ts`: fixed the stale "no consumer" doc comment.

**Simplification:**
- `missedVisit.job.ts`: the three repeated `Bearer`-strips are now one `rawToken`; the HR/non-HR notification branches now build one payload and call `createNotification` once.

**Not addressed in this pass** (flagged, not fixed):
- pino-logger conversion for these two jobs — no standalone (non-HTTP-request-scoped) pino logger exists in `service-commons` yet; adding one is a bigger shared-infra decision than this PR's scope.
- Cross-service `isPrivileged`/SYSTEM-role duplication — real observation, but a shared-helper consolidation across 7+ services is its own follow-up, not a blocking fix here.

All 5 affected services + `service-commons` re-tested (visit-form-service: 620/620, others: 221/221), lint/build clean.

PR #210, opened 2026-09-02
feat(auth-service,sync-service): service accounts and sync-delay escalations

logeshn-1:
## Summary
- Adds service accounts and POST /auth/service-token to auth-service for the m2m client-credentials exchange
- Raises sync-delay escalations from sync-service's supervisor roster dashboard via the new escalation client
- Fixes the shared bootstrap error-swallowing bug in cms-content-service's main.ts
- Depends only on service-commons content already merged via #205/#206 — no cross-PR file duplication needed here

## Test plan
- [x] `nx run-many -t test -p auth-service sync-service cms-content-service --passWithNoTests` — 26 suites / 306 tests passing (pre-existing skips)
- [x] `nx run-many -t lint -p auth-service sync-service cms-content-service` — clean
- [x] `nx run-many -t build -p auth-service sync-service cms-content-service` — clean

### DIFF
--- a/apps/auth-service/src/auth/auth.service.ts
+++ b/apps/auth-service/src/auth/auth.service.ts
@@ -112,15 +112,23 @@ export class AuthService {
     const account = await this.repository.findServiceAccountByClientId(input.clientId);
 
     // Same generic failure for "no such client" and "wrong secret" as login
-    // — never reveal which one it was.
+    // — never reveal which one it was. Also same failed-attempt tracking as
+    // login: clientIds aren't secret (shipped in .env.example as
+    // descriptive strings like "sync-service"), so without this an
+    // attacker who learns one could guess clientSecret unlimited times with
+    // no record kept anywhere and no lockout.
     if (!account || !account.isActive) {
+      if (account) await this.repository.incrementServiceAccountFailedAuthCount(account.id);
       throw unauthorized('Invalid credentials.');
     }
 
     const secretMatches = await verifyPassword(account.clientSecretHash, input.clientSecret);
     if (!secretMatches) {
+      await this.repository.incrementServiceAccountFailedAuthCount(account.id);
       throw unauthorized('Invalid credentials.');
     }
+
+    await this.repository.recordSuccessfulServiceAuth(account.id);
 
     const accessToken = await this.signer.sign(
       { sub: account.id, roles: [account.role], typ: 'service' },
--- a/apps/auth-service/src/auth/auth.service.ts
+++ b/apps/auth-service/src/auth/auth.service.ts
@@ -15,6 +15,7 @@ import { toUserProfile, type AuthTokens, type CreatedUser, type UserProfile } fr
 import type { LoginInput } from './dto/login.dto';
 import type { CreateUserInput } from './dto/create-user.dto';
 import type { UpdateUserInput } from './dto/update-user.dto';
+import type { ServiceTokenInput } from './dto/service-token.dto';
 
 // Re-exported so importers of `./auth.service` keep resolving these types;
 // their definitions live in auth.mapper.ts.
@@ -24,6 +25,13 @@ export type { AuthTokens, CreatedUser, UserProfile } from './auth.mapper';
 export interface CallerScope {
   readonly id: string;
   readonly roles: string[];
+}
+
+/** Result of a client-credentials exchange — no refresh token; jobs re-mint per run. */
+export interface ServiceTokenResult {
+  accessToken: string;
+  expiresIn: number;
+  roles: string[];
 }
 
 /** Prisma unique-constraint violation code (mobileNumber/email/roleCode etc). */
@@ -91,6 +99,39 @@ export class AuthService {
     // Idempotent by design — revoking an already-revoked/non-existent session
     // is a no-op (updateMany matches zero rows), not an error.
     await this.repository.revokeSessionByRefreshTokenHash(tokenHash);
+  }
+
+  /**
+   * Client-credentials exchange for a machine identity — the real
+   * service-to-service auth this platform was missing (automated jobs
+   * previously had no way to call an ADMIN-only endpoint without forwarding
+   * a human's own token). No session/refresh token is issued: callers are
+   * short-lived jobs that just re-mint a token per run.
+   */
+  async issueServiceToken(input: ServiceTokenInput): Promise<ServiceTokenResult> {
+    const account = await this.repository.findServiceAccountByClientId(input.clientId);
+
+    // Same generic failure for "no such client" and "wrong secret" as login
+    // — never reveal which one it was.
+    if (!account || !account.isActive) {
+      throw unauthorized('Invalid credentials.');
+    }
+
+    const secretMatches = await verifyPassword(account.clientSecretHash, input.clientSecret);
+    if (!secretMatches) {
+      throw unauthorized('Invalid credentials.');
+    }
+
+    const accessToken = await this.signer.sign(
+      { sub: account.id, roles: [account.role], typ: 'service' },
+      this.adminAccessTokenTtl,
+    );
+
+    return {
+      accessToken,
+      expiresIn: Math.floor(parseDurationMs(this.adminAccessTokenTtl) / 1000),
+      roles: [account.role],
+    };
   }
 
   /**
--- a/apps/sync-service/.env.example
+++ b/apps/sync-service/.env.example
@@ -8,3 +8,12 @@ CORS_ORIGINS=http://localhost:5173
 # https://staging-api.example.com,https://api.example.com. Leave empty in
 # local dev to fall back to http://localhost:$PORT.
 PUBLIC_BASE_URLS=
+# Client-credentials identity (see auth-service's SERVICE_ACCOUNTS var) this
+# service authenticates as, to raise SYNC_DELAY escalations from
+# GET /sync/last-synced/by-roster. Optional: that step is skipped (roster
+# data is still returned) when unset.
+SERVICE_ACCOUNT_CLIENT_ID=sync-service
+SERVICE_ACCOUNT_CLIENT_SECRET=
+# Hours since a Sakhi's last COMPLETED sync before she's flagged SYNC_DELAY
+# on the Supervisor roster dashboard.
+SYNC_DELAY_THRESHOLD_HOURS=48
--- a/apps/sync-service/src/config/app-config.ts
+++ b/apps/sync-service/src/config/app-config.ts
@@ -21,9 +21,20 @@ const schema = z.object({
   // from GET /sync/last-synced/by-roster's read-time SYNC_DELAY check.
   // Optional: that check is skipped (roster data is still returned) when
   // unset, rather than failing the whole read over a not-yet-provisioned
-  // credential.
-  SERVICE_ACCOUNT_CLIENT_ID: z.string().min(1).optional(),
-  SERVICE_ACCOUNT_CLIENT_SECRET: z.string().min(1).optional(),
+  // credential. The `.transform` treats a present-but-empty value (the
+  // .env.example convention of leaving the key present with no value) the
+  // same as the key being absent — `.optional()` alone only tolerates
+  // `undefined`, so an empty string would otherwise still fail `.min(1)`
+  // and crash the whole service at boot (loadConfig calls process.exit(1)
+  // on any schema validation failure) instead of just skipping this check.
+  SERVICE_ACCOUNT_CLIENT_ID: z
+    .string()
+    .optional()
+    .transform((v) => v || undefined),
+  SERVICE_ACCOUNT_CLIENT_SECRET: z
+    .string()
+    .optional()
+    .transform((v) => v || undefined),
   // How many hours since the last COMPLETED sync before a roster member is
   // flagged SYNC_DELAY. A plain env var for now, not a GoRules rule pack —
   // unlike the clinical/visit-scheduling thresholds this repo's other rule
--- a/apps/sync-service/src/config/app-config.ts
+++ b/apps/sync-service/src/config/app-config.ts
@@ -16,6 +16,22 @@ const schema = z.object({
         .filter(Boolean),
     ),
   PUBLIC_BASE_URLS: publicBaseUrlsSchema,
+  // Client-credentials identity (POST /auth/service-token) this service
+  // authenticates as, to call the ADMIN/SYSTEM-only POST /escalation-events
+  // from GET /sync/last-synced/by-roster's read-time SYNC_DELAY check.
+  // Optional: that check is skipped (roster data is still returned) when
+  // unset, rather than failing the whole read over a not-yet-provisioned
+  // credential.
+  SERVICE_ACCOUNT_CLIENT_ID: z.string().min(1).optional(),
+  SERVICE_ACCOUNT_CLIENT_SECRET: z.string().min(1).optional(),
+  // How many hours since the last COMPLETED sync before a roster member is
+  // flagged SYNC_DELAY. A plain env var for now, not a GoRules rule pack —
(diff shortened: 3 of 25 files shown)


apps/sync-service/src/sync/systemEscalation.client.ts
```diff
@@ -0,0 +1,55 @@
...
+): Promise<EscalationEvent> {
+  let res: Response;
+  try {
+    res = await fetch(`${API_GATEWAY_BASE_URL}/api/v1/escalation-events`, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${systemAccessToken}`,
+      },
+      body: JSON.stringify({ sakhiUserId, escalationType: 'SYNC_DELAY' }),
```
palani-ND: **This POST body is missing `assignedSupervisorId`, which the receiving DTO requires (not optional) — every one of these calls will 400.**

`create-escalation-event.dto.ts`'s `createEscalationEventSchema` declares `assignedSupervisorId: z.string().uuid()` with no `.optional()`, and its own comment explains why: "this service has no roster lookup of its own, so the raiser... must resolve and supply the owning Supervisor." This function's only two params are `sakhiUserId`/`systemAccessToken` — there's no way to supply it.

The caller (`syncBatch.service.ts`'s `getLastSyncedAtByRoster`) already has exactly the right value in scope: `caller.id` (the Supervisor making the dashboard request) is precisely the owning Supervisor for every Sakhi in their own roster. Since the call is wrapped in `.catch(err => console.error(...))` at the call site, this failure is completely silent — the dashboard read still returns correct `isDelayed` flags, and the unit tests pass because `createSyncDelayEscalationEvent` is `jest.mock`'d at the module level (never exercising the real HTTP body against the real Zod schema). In production, no SYNC_DELAY escalation is ever actually created. Fix: add an `assignedSupervisorId` param here and thread `caller.id` through from the call site.

apps/sync-service/.env.example
```diff
@@ -8,3 +8,12 @@ CORS_ORIGINS=http://localhost:5173
 # https://staging-api.example.com,https://api.example.com. Leave empty in
 # local dev to fall back to http://localhost:$PORT.
 PUBLIC_BASE_URLS=
+# Client-credentials identity (see auth-service's SERVICE_ACCOUNTS var) this
+# service authenticates as, to raise SYNC_DELAY escalations from
+# GET /sync/last-synced/by-roster. Optional: that step is skipped (roster
+# data is still returned) when unset.
+SERVICE_ACCOUNT_CLIENT_ID=sync-service
+SERVICE_ACCOUNT_CLIENT_SECRET=
```
palani-ND, line 16: **This blank-but-present value will crash the entire service at boot, not just skip the SYNC_DELAY escalation feature as the code comments claim.**

`app-config.ts`'s schema is `SERVICE_ACCOUNT_CLIENT_SECRET: z.string().min(1).optional()`. `.optional()` only tolerates the *key being absent* (`undefined`) — a present-but-empty string (exactly what's on this line) still fails `.min(1)`. `loadConfig()` (`libs/service-commons/src/config/load-config.ts:9-14`) calls `process.exit(1)` on any schema validation failure.

So a developer who copies this `.env.example` to `.env` as-is (the normal setup flow) and doesn't think to either delete this line or fill in a value gets the *whole sync-service* refusing to start — contradicting `getLastSyncedAtByRoster`'s own doc comment, which frames this credential as optional and says the SYNC_DELAY check is simply "skipped" when unset. Either blank out the line entirely (so the key is truly absent) or change the schema to also accept an empty string as "not configured" (e.g. `.transform(v => v || undefined)`).

apps/sync-service/src/sync/systemEscalation.client.ts
```diff
@@ -0,0 +1,55 @@
...
+  id: string;
+  sakhiUserId: string | null;
+  escalationType: string;
+  status: string;
+}
+
+/**
+ * Raises a SYNC_DELAY escalation via `POST /escalation-events` (through the
+ * gateway), authenticating as this service's machine identity. Called at
+ * Supervisor dashboard read time (getLastSyncedAtByRoster), not from a cron
```
palani-ND, line 17: **This is the only place SYNC_DELAY escalations are ever raised, and it only runs when a Supervisor happens to open their dashboard.** The codebase already has a proven, replica-safe pattern for exactly this kind of "raise once, independent of who's looking" problem — `libs/service-commons/src/jobs/job-lock.ts`, actively used by `notification-escalation-service` for its own overdue-notification sweep — but this feature bypasses it in favor of a request-coupled side effect.

Combined with the missing-`assignedSupervisorId` bug (other comment) and the boot-crash-on-blank-secret bug (`.env.example`), there are now three independent ways this feature can end up doing nothing: a Supervisor who doesn't check their dashboard for weeks means their whole roster's sync delays never generate an escalation regardless of the other two bugs being fixed — escalation coverage becomes a function of Supervisor UI usage rather than actual sync health, which seems to undercut the point of an automated escalation. Worth confirming this passive-list design (mentioned in the doc comment as a deliberate build-plan choice) is really the intended coverage model, versus wiring this into a scheduled job like the other services do.

apps/sync-service/src/sync/syncBatch.repository.ts
```diff
@@ -27,4 +27,29 @@ export class SyncBatchRepository {
...
+   * The most recent COMPLETED batch's completedAt per userId, for the
+   * Supervisor roster dashboard — the batch counterpart of
+   * findLastSyncedAt. A single query (not one per userId): every COMPLETED
+   * batch for the roster, grouped in application code to the max
+   * completedAt per user, since Prisma's `groupBy` can't express "max of a
+   * field, but return that whole row" directly. A userId with zero
+   * COMPLETED batches simply has no entry in the returned Map — the caller
+   * treats a missing entry the same as `findLastSyncedAt` returning null.
+   */
+  async findLastSyncedAtByUserIds(userIds: string[]): Promise<Map<string, Date>> {
```
palani-ND, line 42: **This fetches every COMPLETED sync batch for the whole roster and dedupes in application code, instead of asking Postgres for one row per user.** `findMany` has no `take`/`distinct` — the max-per-user reduction happens in a JS loop after the full result set is transferred. Over time (a Supervisor's roster syncing regularly for months), this grows unbounded: a 20-Sakhi roster with ~2 syncs/day easily accumulates thousands of rows fetched to extract 20 useful values, on every dashboard load.

Prisma's `groupBy` with `_max: { completedAt: true }` returns exactly what's needed in one aggregate query:
```ts
const rows = await this.prisma.syncBatch.groupBy({
  by: ['userId'],
  where: { userId: { in: userIds }, status: 'COMPLETED' },
  _max: { completedAt: true },
});
```
Also worth adding a supporting `@@index([userId, status, completedAt])` on `SyncBatch` in `schema.prisma` — there isn't one today, and this is the first caller to run this query repeatedly for a whole roster (not just one user) on a live dashboard hot path.

palani-ND (commented):
Automated code review. Two independent findings below point at the same root cause from different angles (the missing `assignedSupervisorId` and the escalation feature's read-triggered-only design), so read them together — the net effect is that this PR's core feature (SYNC_DELAY escalations) has a good chance of never actually firing in production even though all tests pass.

apps/auth-service/src/auth/auth.service.ts
```diff
@@ -93,6 +101,39 @@ export class AuthService {
...
   }
 
+  /**
+   * Client-credentials exchange for a machine identity — the real
+   * service-to-service auth this platform was missing (automated jobs
+   * previously had no way to call an ADMIN-only endpoint without forwarding
+   * a human's own token). No session/refresh token is issued: callers are
+   * short-lived jobs that just re-mint a token per run.
+   */
+  async issueServiceToken(input: ServiceTokenInput): Promise<ServiceTokenResult> {
```
palani-ND, line 111: **No failed-attempt tracking for service-token minting, unlike the human `login()` path just above.** `login()` calls `repository.incrementFailedLoginCount(user.id)` on both the not-found and wrong-password branches, giving ops a per-account counter to detect/rate-limit credential stuffing. `issueServiceToken` copies `login`'s "generic 401, don't reveal which check failed" comment and shape but records nothing on either failure branch, and `ServiceAccount` has no equivalent counter column.

`clientId`s aren't secret — they're descriptive strings shipped in `.env.example` (e.g. `sync-service`, `visit-form-service`). An attacker who learns one can attempt unlimited `clientSecret` guesses against `POST /auth/service-token` with zero record kept anywhere and no lockout, unlike the equivalent attack against a human account.

apps/sync-service/src/config/app-config.ts
```diff
@@ -16,6 +16,22 @@ const schema = z.object({
...
+  SERVICE_ACCOUNT_CLIENT_ID: z.string().min(1).optional(),
+  SERVICE_ACCOUNT_CLIENT_SECRET: z.string().min(1).optional(),
+  // How many hours since the last COMPLETED sync before a roster member is
+  // flagged SYNC_DELAY. A plain env var for now, not a GoRules rule pack —
+  // unlike the clinical/visit-scheduling thresholds this repo's other rule
+  // packs encode, this is a single operational number with no
+  // visitFamily-style branching, and evaluating it via a rules-service round
+  // trip on every dashboard read adds latency this check doesn't need.
+  // Revisit if this ever needs per-project/per-geography variation.
+  SYNC_DELAY_THRESHOLD_HOURS: z.coerce.number().positive().default(48),
```
palani-ND, line 45: **This bakes a business threshold into a plain env var, which the root `.claude/CLAUDE.md` explicitly says not to do.** §1 ("Purpose & boundaries") states: *"What must NEVER be added: ...business thresholds/rates (those live in GoRules)."* A 48-hour sync-delay cutoff is exactly that kind of value — the adjacent doc comment even argues around this rule directly ("unlike the clinical/visit-scheduling thresholds this repo's other rule packs encode... revisit if this ever needs per-project/per-geography variation"), but the rule doesn't carve out an exception for single global numbers.

Practically: every other threshold like this in the repo lives in a GoRules `RuleSet`/`ruleVersion` an admin can update via the existing rule-versioning API with no deploy. This one requires a code change + full sync-service redeploy to adjust — which becomes a real constraint the first time a program manager wants a different SLA for one project/geography.

logeshn-1:
Addressed in 4652465:

**Fixed:**
- `systemEscalation.client.ts`/`syncBatch.service.ts`: threaded `assignedSupervisorId` (the caller's own id) through to `createSyncDelayEscalationEvent` — confirmed this was exactly the bug described: `createEscalationEventSchema` requires this field, so every call was silently 400ing and getting swallowed by the best-effort `.catch()`, meaning no SYNC_DELAY escalation was ever actually created in production.
- `app-config.ts`: `SERVICE_ACCOUNT_CLIENT_ID`/`SECRET` now transform a present-but-empty value to `undefined` before the `.optional()` check, so copying `.env.example` as-is no longer crashes the whole service at boot.
- `syncBatch.repository.ts`: `findLastSyncedAtByUserIds` now uses `groupBy` + `_max(completedAt)` instead of fetching every COMPLETED batch and reducing in application code. Added the suggested `@@index([userId, status, completedAt])` on `SyncBatch` (migration included — there was no index on this table at all before).
- `auth.service.ts`/`auth.repository.ts`: `issueServiceToken` now records failed/successful attempts on `ServiceAccount` (`failedAuthCount`/`lastAuthAt`, new columns + migration), mirroring `login()`'s existing `incrementFailedLoginCount`/`recordSuccessfulLogin`.

**Not addressed in this pass** (flagged, not silently skipped):
- Moving `SYNC_DELAY_THRESHOLD_HOURS` into a GoRules RuleSet — agreed this is the CLAUDE.md-correct approach long-term, but wiring a new rule pack through rules-service's `escalationEvaluator` is a bigger scope than this PR; worth its own follow-up.
- The read-triggered-only (dashboard-load) escalation design vs. a scheduled job like the other escalation types use (`job-lock.ts`) — good question on the coverage model. Left as-is for now since the PR description frames it as a deliberate build-plan choice, but flagging that it's worth an explicit confirmation from whoever owns that decision, especially now that the `assignedSupervisorId` fix means this path will actually start firing.

All 3 affected services re-tested (306/324, pre-existing skips), lint/build clean.

apps/sync-service/src/sync/systemEscalation.client.ts
logeshn-1: Fixed in 4652465 — createSyncDelayEscalationEvent now takes assignedSupervisorId and the call site threads through caller.id (the Supervisor is the owning roster). Confirmed this was previously silently 400ing every time in production, as you diagnosed — tests were mocking the client so this never surfaced.

apps/sync-service/.env.example
logeshn-1: Fixed in 4652465 — SERVICE_ACCOUNT_CLIENT_ID/SECRET now transform a present-but-empty value to undefined before the .optional() check, so copying .env.example as-is no longer crashes the whole service at boot.

apps/sync-service/src/sync/syncBatch.repository.ts
logeshn-1: Fixed in 4652465 — findLastSyncedAtByUserIds now uses Prisma's groupBy with _max(completedAt) instead of fetching every COMPLETED batch and reducing in application code, plus a supporting @@index([userId, status, completedAt]) on SyncBatch.

apps/auth-service/src/auth/auth.service.ts
logeshn-1: Fixed in 4652465 — issueServiceToken now records failed/successful attempts on ServiceAccount (failedAuthCount/lastAuthAt), mirroring login()'s incrementFailedLoginCount/recordSuccessfulLogin.

apps/sync-service/src/sync/systemEscalation.client.ts
logeshn-1: Not addressed in this pass — this is a coverage-model design question, not a mechanical bug, so leaving it open for a maintainer call rather than silently reworking it. Agreed the job-lock pattern (as notification-escalation-service already uses) is the right long-term fix if the read-triggered design isn't intentional. Happy to wire it into a scheduled job on confirmation.

apps/sync-service/src/config/app-config.ts
logeshn-1: Not addressed in this pass — agreed this is exactly the kind of business threshold CLAUDE.md says belongs in GoRules, not an env var. Migrating it is a larger rules-service wiring change than this PR's scope (new RuleSet, rule-version seed, evaluation call from sync-service). Left open pending a decision on scope/priority for that follow-up rather than doing a partial migration here.

logeshn-1:
4 of 6 findings fixed in 4652465, 2 left open as design questions needing a maintainer call rather than a silent rework:

**Fixed:**
- Missing `assignedSupervisorId` — every SYNC_DELAY escalation was silently 400ing in production despite tests passing (tests mocked the client)
- `.env.example`'s blank `SERVICE_ACCOUNT_CLIENT_SECRET` crashing the whole service at boot
- Inefficient roster sync-status query — replaced with Prisma `groupBy`/`_max` + a supporting index
- No failed-attempt tracking on service-token minting — now mirrors the human `login()` path

**Left open (replied on each thread):**
- Read-triggered-only SYNC_DELAY escalation design vs. a scheduled job (like `notification-escalation-service` uses via `job-lock.ts`) — agreed this is the better long-term pattern, but reworking the trigger model wasn't done silently as part of a "review comment fix" pass
- `SYNC_DELAY_THRESHOLD_HOURS` as an env var instead of a GoRules `RuleSet` — agreed with the finding, but migrating it is a larger rules-service wiring change than this PR's scope

CI green: `nx run-many -t test -p auth-service sync-service --passWithNoTests` — 26 suites / 306 tests passing.

PR #212, opened 2026-09-02
feat: LMP change contract, referral decisions, audit write path, non-approval edits

### LINKED ISSUE #211: LMP change contract, referral decision persistence, audit write path, non-approval form edits
## Summary

Closes 4 backend gaps identified in a frontend-facing investigation (LMP editing, Referral Follow-up decision visibility, field-level audit trail, non-approval-editable fields), plus a data-correctness bug found along the way is explicitly out of scope here (tracked separately with the risk-referral/rules team).

## What this closes

1. **LMP_CHANGE creation contract** (approval-service) — `POST /lmp-change-requests` + `GET /lmp-change-requests?beneficiaryId=` (required param), idempotent via `localRequestUuid`, matching the existing `requestPayloadJson` shape already read by the decision flow.
2. **Referral decision persistence + REFILL fix** (risk-referral-service) — `Referral` gains `decidedByUserId`/`decidedAt`/`decisionNotes`; REFILL is no longer a silent no-op; `GET /referrals` gains an optional `beneficiaryId` filter.
3. **Mobile-safe, idempotent audit write path** (audit-service) — `POST /audit` widened to permit SUPERVISOR and SAKHI callers within namespaced action-prefix allowlists, idempotent via `localAuditUuid`.
4. **Wire LMP decisions to the audit log** (approval-service) — `decideLmpChangeCard` now writes a best-effort audit entry for both APPROVE and REJECT, closing a live SRS violation ("every edit must be written to the audit log").
5. **Non-approval-editable form fields** (visit-form-service) — new `PATCH /form-submissions/:id/answers`, scoped to the exact SRS Appendix J.4 field list, all-or-nothing validation, real `beforeJson`/`afterJson` audit entries.

## Security note

Three separate Critical IDOR (Insecure Direct Object Reference) bugs were found and fixed during implementation/review, none present in the final merged state:
- A cross-actor idempotency-key collision on the new audit-write path (any caller's `localAuditUuid` could return an unrelated caller's audit entry).
- A missing beneficiary-ownership check on the new non-approval-edit endpoint (any Sakhi could edit any beneficiary's submitted form answers).
- A missing beneficiary-ownership check on the new LMP-change-request create/list endpoints (any Sakhi could create or enumerate LMP change requests for any beneficiary, including exposing PII).

All three were caught by review before merge and are covered by regression tests reproducing the original exploit.

## Explicitly out of scope

- LMP corrections not flowing into rules-service risk/gestational-age calculations — a data-correctness bug, tracked separately with the risk-referral/rules team.
- A SAKHI-facing audit *read* endpoint — open product decision, not resolved here.
- Incentive-release visibility for Accompanied Referrals — open product decision, not resolved here.

Dharanishwaran-ND:
Closes #211
Implements the confirmed clarifications from #191

## Summary

Closes 4 backend gaps identified in a frontend-facing investigation (LMP editing, Referral Follow-up decision visibility, field-level audit trail, non-approval-editable fields), plus GoRules risk-grading fixes from issue #191's clarification round:

- **`POST /lmp-change-requests` + `GET /lmp-change-requests?beneficiaryId=`** (approval-service) — idempotent LMP change request creation, matching the existing `requestPayloadJson` shape already consumed by the decision flow. Beneficiary-scoped, ownership-checked.
- **Referral decision persistence + REFILL fix** (risk-referral-service) — `Referral` gains `decidedByUserId`/`decidedAt`/`decisionNotes`; REFILL is no longer a silent no-op; `GET /referrals` gains an optional `beneficiaryId` filter.
- **Mobile-safe, idempotent audit write path** (audit-service) — `POST /audit` widened to permit SUPERVISOR and SAKHI callers within namespaced action-prefix allowlists (`SUPERVISOR: ['QUICK_RESPONSE_', 'LMP_CHANGE_']`, `SAKHI: ['FORM_ANSWER_EDIT']`), idempotent via `localAuditUuid`.
- **Wire LMP decisions to the audit log** (approval-service) — `decideLmpChangeCard` now writes a best-effort audit entry for both APPROVE and REJECT, closing a live SRS violation ("every edit must be written to the audit log").
- **Non-approval-editable form fields** (visit-form-service) — new `PATCH /form-submissions/:id/answers`, scoped exactly to SRS Appendix J.4's field list (with two ambiguities resolved directly by the product owner — see commit messages), all-or-nothing validation, real `beforeJson`/`afterJson` audit entries.
- **GoRules risk-grading fixes (issue #191)** (rules-service, visit-form-service) — implements 4 of the 7 clarifications ARMMAN confirmed:
  - Hypertension's Mild band now requires `historyOfHypertension` OR `gestationalHypertension` in addition to the BP range (was BP range alone); history merged in from `MOTHER_REGISTRATION`'s existing medical-history question, no new form field needed.
  - Neuro-developmental status: a missed milestone now grades SEVERE (was MILD) and triggers an HR visit (previously did not).
  - Feeding Adequacy now reads the dedicated `feeding_concerns` field (any value other than `no_concerns` grades High Risk + referral), replacing the prior `current_feeding_practice` derivation.
  - New `ACTIVITY_LEVEL` condition: `active_and_moving`→Normal, `reduced_movement`→Mild (assessment finding, no referral), `lethargic`→Severe+referral.
  - Deliberately left unchanged, per what was actually confirmed: BOH's pre-term/LSCS criteria (no new field given), SCD's grade tier (still unconfirmed SEVERE vs. Moderate), infant deformity (explicitly no threshold without further sign-off), Hypotension's shock-features wiring (still need to know which signs qualify), prematurity in later visits (confirmed no effect — already matched).
  - `docs/Appendix_D_High_Risk_Detection_Rules.md` gets a new §1.8 recording all 7 answers and implementation status.

## Security note

Three separate Critical IDOR bugs were found and fixed during implementation/review — none present in the final state, all covered by regression tests reproducing the original exploit:

1. Cross-actor idempotency-key collision on the new audit-write path — a client-supplied `localAuditUuid` colliding with an unrelated caller's row would return that row's contents instead of erroring. Fixed with an `isSameLogicalEntry` check at both call sites.
2. Missing beneficiary-ownership check on `PATCH /form-submissions/:id/answers` — any authenticated Sakhi could edit any beneficiary's submitted form answers system-wide. Fixed by reusing the existing `assertCallerOwnsBeneficiary` guard.
3. Missing beneficiary-ownership check on the new LMP-change-request create/list endpoints — any Sakhi could create or enumerate LMP change requests for any beneficiary, exposing PII on the read side. Fixed by reusing the existing `BeneficiaryClient.getById` ownership pattern (mirroring `reopen-request.service.ts`), including the nonexistent-beneficiary edge case.

## Test plan

- [x] `npx nx run-many -t lint test build --projects=approval-service,risk-referral-service,audit-service,visit-form-service` — all clean
- [x] `npx tsc --noEmit` clean on all 5 touched services (adds rules-service)
- [x] Task 2's migration applied and verified against a real local Postgres instance (column types/nullability confirmed via `psql`)
- [x] Every task independently code-reviewed; both fix rounds (Task 3's IDOR + role-mapping defect, Task 5's IDOR, the final-review fix wave's Task 1 IDOR) each re-reviewed clean after their fixes
- [x] Issue #191's risk-grading changes: 199 rules-service + 700 visit-form-service + 251 risk-referral-service tests pass after merging in and rebuilding against this branch's full dependency set

## Notes

- Two SRS Appendix J.4 mapping ambiguities were escalated during implementation and resolved directly by the product owner (see `feat(visit-form-service): Add PATCH form-submissions answers endpoint` commit message for details) — LMP date/EDD are deliberately excluded from the direct-edit allowlist (they only change via this PR's own approval-gated LMP_CHANGE flow), and Infant Registration's "current length/weight at registration" fields are aliased onto the existing birth-measurement fields since no separate schema field exists for them.
- This branch was merged forward from `develop` to pick up PR #208 (ANC/Infant RISK rule pack seeding, SCD grading) and #209/#210 (visit-form-service cron/m2m auth, auth-service/sync-service service accounts) — it had forked before #208 merged and was missing that work entirely. A missing `node-cron` dependency from #209's cron-job code was also found and fixed as part of that merge (unrelated pre-existing gap, needed for this branch to build/test cleanly).
- Out of scope, tracked separately: LMP corrections not flowing into rules-service risk/GA calculations (data-correctness bug, different team); a SAKHI-facing audit read endpoint and incentive-release visibility (both open product decisions, not resolved here); issue #191's remaining 2 unresolved clarifications (Hypotension shock-features specifics, SCD grade tier confirmation).
- This branch touches auth/audit paths — per repo convention this needs 2 approvals.


apps/risk-referral-service/src/referrals/referral.service.ts
```diff
@@ -25,8 +25,8 @@ export class ReferralService {
     private readonly incentiveClient: IncentiveClient = new IncentiveClient(),
   ) {}
 
-  list() {
-    return this.repository.findMany();
+  list(beneficiaryId?: string) {
```
palani-ND: `list()` takes no `caller`/`authorizationHeader` at all — it just forwards `beneficiaryId` straight to `repository.findMany(beneficiaryId)`. The route (`referral.controller.ts`, `GET /referrals`) only checks `requireRoles('SAKHI', 'SUPERVISOR', 'MANAGER')`, with no roster/ownership scoping layered on top.

Concretely: any authenticated Sakhi can call `GET /referrals?beneficiaryId=<any-uuid>` and get back that beneficiary's full referral history — including the `decidedByUserId`/`decisionNotes` fields this PR adds — for a beneficiary entirely outside her own case list. This PR gets the equivalent case right elsewhere: the new `GET /lmp-change-requests?beneficiaryId=` in the same PR explicitly calls `beneficiaryClient.getById(beneficiaryId, authorizationHeader)` first specifically to enforce SAKHI-own-case/SUPERVISOR-roster scoping before querying — this endpoint is missing that same check.

Worth adding the same ownership guard here (pass `caller`/`authorizationHeader` through from the controller and resolve scoping the way `decide()` already does for SUPERVISOR a few lines below, or the way the LMP-change-request list endpoint does).

apps/audit-service/src/audit/auditLog.service.ts
```diff
@@ -16,20 +77,93 @@ export class AuditLogService {
...
+   * risk-referral-service's visitId idempotent-create handling.
    */
-  create(dto: CreateAuditLogInput, caller: CallerIdentity) {
+  async create(dto: CreateAuditLogInput, caller: CallerIdentity) {
+    let toCreate = dto;
     if (!caller.roles.includes('ADMIN')) {
-      if (!dto.action.startsWith('QUICK_RESPONSE_')) {
-        throw forbidden('SUPERVISOR may only log QUICK_RESPONSE_* actions.');
+      const allowedPrefixes = caller.roles.flatMap((role) => ALLOWED_ACTION_PREFIXES[role] ?? []);
+      if (!allowedPrefixes.some((prefix) => dto.action.startsWith(prefix))) {
```
palani-ND: The authorization check here only validates `dto.action` against the role's allowed prefixes — `entityType`/`entityId` (both free-form strings, up to 80/no-max chars, per `create-auditLog.dto.ts`) are never checked against the action or against anything the caller actually did. The doc comment above (a few lines up) states this widened role "can never... write an arbitrary action/entityType" — that's true for `action` (prefix-checked) and `actorUserId` (forced to caller), but not for `entityType`/`entityId`.

Concretely: a SAKHI can call `POST /audit` directly (route allows `requireRoles('ADMIN', 'SUPERVISOR', 'SAKHI')`) with `action: 'FORM_ANSWER_EDIT'` (passes the prefix check) but `entityType`/`entityId`/`beforeJson`/`afterJson` set to anything she likes — fabricating an audit-log entry that appears to document an edit to a beneficiary she has no relationship to, self-attributed (since `actorUserId` is forced to her own id, this isn't impersonation, but it is a false record). Same gap for SUPERVISOR's `LMP_CHANGE_`/`QUICK_RESPONSE_` prefixes.

Since `POST /audit` is meant to be called only by other services forwarding a caller's already-validated decision (not by the mobile app directly with arbitrary entity data), worth either validating `entityType` against an allowlist per action prefix, or requiring the caller to also supply proof the entity is theirs (mirroring the ownership checks the calling services already did before invoking `AuditClient.log`).

apps/risk-referral-service/src/referrals/referral.repository.ts
```diff
@@ -169,19 +178,39 @@ export class ReferralRepository {
...
+   * updateStatus there's no status transition to protect with a conditional
+   * updateMany. The 409 guard (existing.status === 'PENDING_FOLLOWUP') is
+   * re-checked by the caller (referral.service.ts) immediately before this
+   * call, so a plain `update` is safe here.
+   */
+  updateDecisionOnly(
+    id: string,
+    decision: { decidedByUserId: string; decidedAt: Date; decisionNotes: string | null },
+  ) {
+    return this.prisma.referral.update({ where: { id }, data: decision });
```
palani-ND: Unlike `updateStatus` a few lines above (which uses `updateMany` conditioned on `status: fromStatus`, so a stale read can't silently clobber a since-changed row), this is a plain, unconditional `update` by id. The doc comment says the `PENDING_FOLLOWUP` guard is "re-checked by the caller ... immediately before this call, so a plain `update` is safe here" — but that re-check (`referral.service.ts`) reads from the same `existing` object fetched at the very top of `decide()`, before the SUPERVISOR roster-scoping HTTP call — it's not a fresh read immediately before this call.

Concretely: two decisions race on the same `PENDING_FOLLOWUP` referral — Supervisor A submits LAPSE, Supervisor B (or a Sakhi-app retry) submits REFILL around the same time. Both pass the stale `status !== 'PENDING_FOLLOWUP'` check. If A's `updateStatus` (conditional) commits first, the referral is now `LAPSED`. B's `updateDecisionOnly` still runs unconditionally afterward and succeeds anyway — silently overwriting `decidedByUserId`/`decidedAt`/`decisionNotes` to B's values while `status` stays `LAPSED`. The referral ends up correctly LAPSED, but its decision-audit trail now wrongly attributes the decision to B, with no 409 to signal anything went wrong — undermining the very audit trail this PR adds REFILL persistence for.

Worth making this a conditional `updateMany` too (gated on `status: 'PENDING_FOLLOWUP'`), same as its siblings.

apps/risk-referral-service/src/risk-assessments/riskAssessment.service.ts
```diff
@@ -182,6 +191,32 @@ export class RiskAssessmentService {
...
+    // Best-effort HR visit generation (SRS FR-S-5.2(b)) — only when this
+    // evaluation actually detected an HR condition, the phase can carry an
+    // HR visit at all (HR_VISIT_ELIGIBLE_PHASES), and the caller supplied
+    // the visit's actual completion date (older/other callers that omit it
+    // simply don't get this trigger — no HR-visit generation was ever
+    // wired up for them before this, so this is strictly additive).
+    // Failure here must never roll back or fail the already-committed
+    // RiskAssessment/RiskFlag write above.
+    if (hrDetectedFlag && HR_VISIT_ELIGIBLE_PHASES.has(dto.riskPhase) && dto.actualCompletionDate) {
+      const result = await generateHrVisitSchedule(
```
palani-ND, line 203: Two ways this silently no-ops rather than actually generating the HR visit:

1. **Role mismatch**: `POST /risk-assessments` (this call's own route) allows `requireRoles('SAKHI', 'SUPERVISOR', 'MANAGER', 'ADMIN')`, but `generateHrVisitSchedule` forwards the caller's own `authorizationHeader` to `POST /visit-schedules/generate`, which only allows `requireRoles('SAKHI', 'SUPERVISOR')`. A MANAGER/ADMIN-originated risk assessment that detects an HR condition gets a 403 back from visit-form-service, which just becomes `result.ok === false` → `console.error` here — the RiskAssessment/RiskFlag write still succeeds, but the HR-visit schedule silently never gets created for that caller.

2. **Pre-existing bug, newly reachable**: `scheduleMapper.ts`'s HR branch (`toBulkScheduleRows`, unchanged by this PR) hardcodes `sequenceNo = 1` on every call: `toRow(beneficiaryId, hrType, hrType, 1, hrVisit, 'ACTUAL_VISIT')`. Since `localScheduleUuid` is derived deterministically from `beneficiaryId` + `visitCode` (which embeds that fixed `1`), every HR detection for the same beneficiary+phase produces the *same* `localScheduleUuid` — but `hr.rulesJson.ts`'s own doc comment says ANC/INC HR visits are "cumulative — a new HR visit fires on every detection." The second detection computes a different `scheduledDate` but reuses the first detection's key, so the bulk-persistence idempotency check (`visitSchedule.service.ts`) sees a row with the same key but different content and throws a 409 — which again just becomes `result.ok === false` here. This PR is the first thing that actually calls this path with a real caller, so it's the first time this latent defect becomes reachable in production: every HR detection after the first one for a given beneficiary+phase silently fails to schedule a follow-up, contradicting the "cumulative" design this PR is built to support.

Neither failure is visible to anyone beyond a server log — worth at minimum surfacing these two as a monitored metric/alert rather than `console.error`, since a bare log line is easy to miss for a beneficiary-safety-relevant miss.

apps/visit-form-service/src/forms/form.service.ts
```diff
@@ -857,6 +876,116 @@ export class FormService {
...
+    const answerRows = buildFormAnswers(fields, editedFormData);
+
+    const updated = await this.repository.updateSubmissionAnswers(
+      submissionId,
+      mergedFormDataJson,
+      answerRows,
+    );
+
+    try {
+      await this.auditClient.log(
```
palani-ND, line 972: This is the third mobile-facing sync write added across this PR (alongside `localRequestUuid` on LMP change requests and `localAuditUuid` on audit-log entries), but `patchFormSubmissionAnswersSchema`/`updateSubmissionAnswers` has no equivalent client-generated idempotency key, and this `AuditClient.log()` call never sends a `localAuditUuid`.

A dropped-connection retry of this Sakhi-facing `PATCH` (the underlying `formDataJson`/`FormAnswer` write is idempotent for identical values, so that part is fine) has nothing to dedupe the resulting audit-log call on — so the retry writes a second `FORM_ANSWER_EDIT` `audit_log` row for the same logical edit. Worth adding a `localAuditUuid` here too, same as the LMP-change-request and audit-service call sites, so the audit trail for this PII-adjacent flow doesn't double up on retry.

apps/approval-service/src/lmp-change-requests/lmp-change-request.service.ts
```diff
@@ -0,0 +1,181 @@
...
+    detail: Awaited<ReturnType<QuickResponseService['getLmpChangeRequestDetail']>>;
+    wasCreated: boolean;
+  }> {
+    const beneficiary = await this.beneficiaryClient.getById(
+      dto.beneficiaryId,
+      authorizationHeader,
+    );
+    if (!beneficiary) throw notFound('The beneficiary linked to this request was not found.');
+
+    const existing = await this.repository.findByLocalRequestUuid(dto.localRequestUuid);
```
palani-ND, line 97: This idempotency lookup returns whatever row matches `dto.localRequestUuid`, without checking it actually belongs to `dto.beneficiaryId`/this caller — unlike `audit-service`'s `isSameLogicalEntry` (added elsewhere in this same PR specifically to close this class of bug) which 409s on a UUID that matches a different logical entry instead of silently returning it.

To be clear, this isn't the cross-actor IDOR that shape might suggest at first glance: `getLmpChangeRequestDetail(existing.id, ...)` re-resolves `existing.beneficiaryId` (not `dto.beneficiaryId`) through `beneficiaryClient.getById`, which does enforce SAKHI-own-case/SUPERVISOR-roster scoping server-side in beneficiary-service — so a genuinely out-of-scope collision correctly 403s before any data crosses the boundary. The narrower gap: if the *same* caller (in scope for more than one beneficiary) submits a `localRequestUuid` that happens to collide with their own earlier request for a *different* beneficiary — e.g. a client-side UUID-generation bug — `create()` silently returns that other, pre-existing request's detail with `wasCreated: false`/200, instead of creating a new request for the beneficiary actually named in this call or surfacing the collision as a 409.

Worth adding the same `existing.beneficiaryId === dto.beneficiaryId && existing.requestedByUserId === requestedByUserId` check audit-service now has, for consistency and to catch that client-bug case as a clear error instead of a silently-wrong response.

apps/approval-service/src/lmp-change-requests/lmp-change-request.service.ts
```diff
@@ -0,0 +1,181 @@
+import { HttpError, notFound } from '@armman/service-commons';
+import type { LmpChangeRequestRepository } from './lmp-change-request.repository';
+import type { LookupClient } from '../quick-response/lookup.client';
+import type { QuickResponseService } from '../quick-response/quick-response.service';
+import type { BeneficiaryClient } from '../quick-response/beneficiary.client';
+import type { CreateLmpChangeRequestInput } from './dto/create-lmpChangeRequest.dto';
+
+/** Narrows a caught Prisma error to a unique-constraint violation (P2002). */
+function isUniqueConstraintViolation(err: unknown): boolean {
```
palani-ND: This (and the identical copy added in `audit-service/src/audit/auditLog.service.ts`) treats any P2002 as "the idempotency-key collision," without checking which column collided — unlike the repo's existing `referral.prisma-errors.ts`, which takes a `columnName` parameter specifically to avoid misattributing an unrelated unique-constraint hit.

Safe today only because `approval_requests`/`audit_log` each currently have exactly one non-PK unique column (the one this PR just added). The moment either table gains a second unique constraint, a P2002 from that unrelated constraint would be silently caught here and misreported as an idempotent replay of `localRequestUuid`/`localAuditUuid`. Worth adopting the column-scoped version now while there's only one call site each to update.

apps/visit-form-service/src/forms/form.repository.ts
```diff
@@ -124,6 +124,78 @@ export class FormRepository {
...
+    submissionId: string,
+    mergedFormDataJson: Record<string, unknown>,
+    answerRows: FormAnswerRow[],
+  ) {
+    return this.prisma.$transaction(async (tx) => {
+      const submission = await tx.formSubmission.update({
+        where: { id: submissionId },
+        data: { formDataJson: mergedFormDataJson as never },
+      });
+
```
palani-ND, line 167: Inside the single transaction, this loops over `answerRows` (up to 20 per the DTO's `.max(20)`) issuing one `updateMany` plus a conditional `create` per row, sequentially awaited — up to 40 round trips to Postgres per `PATCH /form-submissions/:id/answers`, all held inside one open transaction/connection. A 20-field edit (the max allowed) does up to 40 sequential queries instead of a batched write. Since this is a Sakhi-facing sync endpoint (likely to see full-allowlist edits during offline-first catch-up), worth collapsing this to a single batched `updateMany` covering all fieldCodes plus one `createMany` for whichever didn't already exist, rather than one round trip per field.

apps/visit-form-service/src/forms/audit.client.ts
```diff
@@ -0,0 +1,49 @@
...
+ * quick-response/audit.client.ts (Task 4) uses, kept as its own copy per
+ * this codebase's per-service client convention (no cross-service imports).
+ *
+ * Unlike Task 4's AuditClient, this one also accepts `beforeJson` — the
+ * prior values of the edited fields — since FORM_ANSWER_EDIT is the first
+ * caller in this codebase to populate it (SRS Appendix J.4: "Every edit
+ * must be written to the audit log (field, old value, new value...)").
+ */
+export class AuditClient {
+  async log(
```
palani-ND, line 24: This is the third near-identical hand-copied `AuditClient` this PR/repo now has (the pre-existing `closure-reopen-service/src/reopen-requests/audit.client.ts`, and this PR's own new `approval-service/src/quick-response/audit.client.ts`) — same fetch/error-mapping shape, no shared implementation in `libs/service-commons`. They've already drifted: this copy and approval-service's new one wrap the fetch in `AbortSignal.timeout(DOWNSTREAM_FETCH_TIMEOUT_MS)`, but the older closure-reopen-service copy has no timeout at all, so a hung `audit-service` call there can hang the caller's request indefinitely while these two are protected. Worth extracting one shared client (e.g. in `libs/service-commons`) now that a third copy has been added, before a fourth service needs audit-write access and copies it again.

apps/approval-service/src/lmp-change-requests/lmp-change-request.service.ts
```diff
@@ -0,0 +1,181 @@
...
+   */
+  async create(
+    dto: CreateLmpChangeRequestInput,
+    requestedByUserId: string,
+    authorizationHeader: string,
+  ): Promise<{
+    detail: Awaited<ReturnType<QuickResponseService['getLmpChangeRequestDetail']>>;
+    wasCreated: boolean;
+  }> {
+    const beneficiary = await this.beneficiaryClient.getById(
```
palani-ND, line 91: The beneficiary-ownership check (`beneficiaryClient.getById(...)` + `if (!beneficiary) throw notFound(...)`) is inlined verbatim here and again in `listByBeneficiaryId` below — worth factoring into one private helper both methods call, especially since a fix to the not-found message or ownership semantics would otherwise need to be applied in both places in this same file to stay in sync.

palani-ND:
## Code review

Left 10 inline comments. This PR does a genuinely careful job on the IDOR class it explicitly went hunting for (the LMP-change-request/audit idempotency-replay checks are correct, and I traced the one path that looked suspicious — `lmp-change-request.service.ts`'s `findByLocalRequestUuid` — closely enough to rule out cross-actor leakage: the downstream `getLmpChangeRequestDetail` call re-resolves the *actual* linked beneficiary through beneficiary-service, which does its own hard ownership check, so an out-of-scope collision correctly 403s). But two endpoints elsewhere in the same PR ended up with the exact gap this PR's own security review was built to catch:

- **`GET /referrals?beneficiaryId=`** (new in this PR) has no ownership/roster scoping at all — `ReferralService.list()` doesn't even take a `caller` parameter. Any authenticated Sakhi can pull any beneficiary's full referral history, including the new decision fields. The sibling `GET /lmp-change-requests?beneficiaryId=` added in this same PR gets this right.
- **`POST /audit`'s widened SAKHI/SUPERVISOR access** only validates `action` against the role's allowlist prefix — `entityType`/`entityId` are free-form and unchecked, despite the code's own comment claiming the opposite. A Sakhi can self-attribute a `FORM_ANSWER_EDIT` audit entry against any `entityId` she names.

Also flagged: a race in the new REFILL decision path where an unconditional update (unlike its siblings) can silently misattribute a decision's audit trail during a concurrent decide; a role mismatch (`MANAGER`/`ADMIN` can trigger risk assessments but not the downstream HR-visit generation their SYSTEM-forwarded call needs) that combines with a pre-existing-but-now-reachable bug in `scheduleMapper.ts`'s hardcoded `sequenceNo`, so HR-visit auto-generation after the first detection per beneficiary+phase silently fails; a missing idempotency key on the third new mobile-sync write (`PATCH /form-submissions/:id/answers`) that can double-write an audit row on retry; and a handful of lower-priority items (a narrower same-caller UUID-collision edge case, an unscoped unique-constraint check, an N+1 in the new batched answer-update, and the `AuditClient` now being hand-copied a third time across services).

I also want to flag that a background review pass initially reported the LMP-change-request idempotency lookup as a full cross-actor IDOR (same class as the two you already found and fixed). I traced it end-to-end before writing anything up and that read turned out to be wrong — the downstream ownership re-check holds. Mentioning this so the inline comment on that line (about a narrower same-caller collision case) isn't mistaken for the more severe claim.

Everything else — the GoRules clarification wiring (hypertension history-gate, neuro MILD→SEVERE, feeding_concerns, activity_level), the IDOR fixes described in the PR body, and the bulk of the RBAC/idempotency work — checked out correctly against the code and the linked issue's confirmed answers.

apps/risk-referral-service/src/referrals/referral.service.ts
Dharanishwaran-ND: Confirmed and fixed in 376f496. `list()` now takes `caller`/`authorizationHeader` and enforces: SAKHI must own the beneficiary (via `assertSakhiOwnsReferral`), SUPERVISOR must have it on their roster (via `listSakhiIdsForSupervisor`), and only MANAGER may omit `beneficiaryId` — SAKHI/SUPERVISOR now get 400 if they omit it. Controller updated to pass `req.user`/authorization header through, 401 if either is missing. Full unit + route test coverage added for all four caller/scope combinations.

apps/audit-service/src/audit/auditLog.service.ts
Dharanishwaran-ND: Confirmed and fixed in 5b175a7. Added a `REQUIRED_ENTITY_TYPE_BY_ACTION_PREFIX` map: FORM_ANSWER_EDIT now requires entityType "FormSubmission", LMP_CHANGE_* requires "MotherCaseDetails" — checked before the actorUserId check, 403 on mismatch. Removed the stale doc-comment claim that widened roles can't forge entityType. QUICK_RESPONSE_ deliberately left unconstrained since nothing in this codebase writes that action yet — will need a similar constraint once a real caller exists. Test coverage added for both the pass and mismatch cases.

apps/risk-referral-service/src/referrals/referral.repository.ts
Dharanishwaran-ND: Confirmed and fixed in 473c306. `updateDecisionOnly` now mirrors `updateStatus`'s conditional `updateMany` gated on `status: 'PENDING_FOLLOWUP'`, returning a boolean instead of the updated row. `decide()`'s REFILL branch checks the result and throws 409 (re-fetching via `findById` on success), same shape as the LAPSE/COMPLETE branches. Added a test asserting the 409 on a concurrent status change.

apps/risk-referral-service/src/risk-assessments/riskAssessment.service.ts
Dharanishwaran-ND: Confirmed as a real gap and fixed. Widened `POST /visit-schedules/generate`'s `requireRoles` to include MANAGER/ADMIN (ab90930) so `generateHrVisitSchedule`'s forwarded caller header works regardless of which role submitted the risk assessment — safe since `assertCanTouchBeneficiary` already treats MANAGER/ADMIN as unrestricted. Also fixed the second bug you flagged: the HR branch of `toBulkScheduleRows` was hardcoding `sequenceNo: 1`, which would collide on a second HR visit for the same beneficiary/phase — now computed from `repository.countByBeneficiaryAndVisitType`.

apps/visit-form-service/src/forms/form.service.ts
Dharanishwaran-ND: Confirmed and fixed in 7b072c8. Added optional `localAuditUuid` to `patchFormSubmissionAnswersSchema`, threaded through `AuditClient.log`/`FormService.updateSubmissionAnswers` to the audit-service call, matching the `localScheduleUuid`/`localSubmissionUuid`/`localRequestUuid` idempotency-key convention used elsewhere in this PR.

apps/approval-service/src/lmp-change-requests/lmp-change-request.service.ts
Dharanishwaran-ND: Confirmed and fixed in f5b49bb. Added the same-caller/same-beneficiary check (`existing.beneficiaryId !== dto.beneficiaryId || existing.requestedByUserId !== requestedByUserId`) on both the sequential pre-check path and the concurrent P2002-race `winner` path, 409ing on a genuine mismatch — same pattern as audit-service's `isSameLogicalEntry`, which you correctly noted this was missing relative to.

apps/approval-service/src/lmp-change-requests/lmp-change-request.service.ts
Dharanishwaran-ND: Confirmed and fixed in both approval-service (f5b49bb) and audit-service (5b175a7). Both `isUniqueConstraintViolation` copies now take a `columnName` parameter and check `meta.target`, mirroring `referral.prisma-errors.ts`'s pattern — call sites updated to `local_request_uuid`/`local_audit_uuid` respectively (confirmed against each schema's `@map(...)` column name).

apps/visit-form-service/src/forms/form.repository.ts
Dharanishwaran-ND: Agreed this is real, but scoping it out of this PR — it's a performance/scale concern (up to 40 round trips per request within a single transaction) rather than a correctness bug, and batching it properly (single `updateMany`-then-`createMany` or similar) touches the transaction shape enough that I'd rather do it as its own focused change with its own test coverage. Will track separately.

apps/visit-form-service/src/forms/audit.client.ts
Dharanishwaran-ND: Agreed — deferring the shared `AuditClient` extraction to `libs/service-commons` out of this PR's scope. Three near-identical copies is a real smell, but extracting it safely means checking all three call sites' error-handling expectations don't quietly diverge, which deserves its own PR rather than riding along here.

apps/approval-service/src/lmp-change-requests/lmp-change-request.service.ts
Dharanishwaran-ND: Agreed, deferring the ownership-check dedup out of this PR's scope. Will factor `beneficiaryClient.getById(...)` + `notFound` into a shared private helper used by both `create` and `listByBeneficiaryId` in a follow-up — didn't want to bundle a refactor into this fix-focused pass.

PR #215, opened 2026-09-02
feat(closure-reopen,risk-referral,beneficiary): overdue-referral escalation

logeshn-1:
## Summary
- Auto-escalates overdue referral follow-ups via a cron job in risk-referral-service, guarded by the shared job-lock (as notification-escalation-service and visit-form-service already use) so replicas never double-process the same tick
- Enriches quick response cards in closure-reopen-service with beneficiary/sakhi names via a new approval-service client
- Adds a system-only post-edd-pending beneficiaries endpoint to beneficiary-service
- Fixes the shared bootstrap error-swallowing bug in audit-service and api-gateway
- Combines the last two module groups of the feature/transfer-escalation split (closure-reopen-service + risk-referral-service + audit-service + api-gateway, and beneficiary-service) into one PR
- Fixed one CI-only failure while preparing this PR: approval.client.spec.ts was importing the real client class, which pulls in Zod-validated app-config and process.exit(1)s without a full env — mocked app-config instead, matching the pattern already used elsewhere in this repo

## Test plan
- [x] `nx run-many -t test` across closure-reopen-service, risk-referral-service, audit-service, api-gateway, beneficiary-service — all passing
- [x] `nx run-many -t lint` and `-t build` across all 5 modules — clean

### DIFF
--- a/apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
+++ b/apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
@@ -26,47 +26,29 @@ export type {
   PiiCreateData,
 } from './beneficiary.repository.types';
 
-interface ListCursor {
-  createdAt: string;
-  id: string;
-}
-
-/** Encodes a row's sort key as an opaque pagination cursor. */
-function encodeCursor(row: { createdAt: Date; id: string }): string {
-  const cursor: ListCursor = { createdAt: row.createdAt.toISOString(), id: row.id };
+/**
+ * Encodes a row's (sortKey, id) pair as an opaque pagination cursor, keyed
+ * by `field` — generic so every cursor-paginated method in this file (and
+ * any future one) shares the same codec instead of each cloning its own
+ * copy differing only in the sort-key field name.
+ */
+function encodeCursor<F extends string>(
+  field: F,
+  row: { [K in F]: Date } & { id: string },
+): string {
+  const cursor = { [field]: row[field].toISOString(), id: row.id };
   return Buffer.from(JSON.stringify(cursor)).toString('base64url');
 }
 
-/** Decodes a cursor produced by encodeCursor; returns null on any malformed input. */
-function decodeCursor(cursor: string): ListCursor | null {
+/** Decodes a cursor produced by encodeCursor for the same `field`; returns null on any malformed input. */
+function decodeCursor<F extends string>(
+  field: F,
+  cursor: string,
+): ({ [K in F]: string } & { id: string }) | null {
   try {
     const parsed = JSON.parse(Buffer.from(cursor, 'base64url').toString('utf8'));
-    if (typeof parsed?.createdAt === 'string' && typeof parsed?.id === 'string') {
-      return parsed as ListCursor;
-    }
-    return null;
-  } catch {
-    return null;
-  }
-}
-
-interface PostEddCursor {
-  eddDate: string;
-  id: string;
-}
-
-/** Encodes findMotherIdsWithEddOnOrBefore's sort key (eddDate, id) as an opaque cursor. */
-function encodePostEddCursor(row: { eddDate: Date; id: string }): string {
-  const cursor: PostEddCursor = { eddDate: row.eddDate.toISOString(), id: row.id };
-  return Buffer.from(JSON.stringify(cursor)).toString('base64url');
-}
-
-/** Decodes a cursor produced by encodePostEddCursor; returns null on any malformed input. */
-function decodePostEddCursor(cursor: string): PostEddCursor | null {
-  try {
-    const parsed = JSON.parse(Buffer.from(cursor, 'base64url').toString('utf8'));
-    if (typeof parsed?.eddDate === 'string' && typeof parsed?.id === 'string') {
-      return parsed as PostEddCursor;
+    if (typeof parsed?.[field] === 'string' && typeof parsed?.id === 'string') {
+      return parsed as { [K in F]: string } & { id: string };
     }
     return null;
   } catch {
@@ -122,7 +104,7 @@ export class BeneficiaryRepository {
       };
     }
 
-    const decodedCursor = filters.cursor ? decodeCursor(filters.cursor) : null;
+    const decodedCursor = filters.cursor ? decodeCursor('createdAt', filters.cursor) : null;
 
     const rows = await this.prisma.beneficiaryCase.findMany({
       where: decodedCursor
@@ -142,7 +124,7 @@ export class BeneficiaryRepository {
     const hasMore = rows.length > filters.limit;
     const items = hasMore ? rows.slice(0, filters.limit) : rows;
     const lastItem = items[items.length - 1];
-    return { items, nextCursor: hasMore && lastItem ? encodeCursor(lastItem) : null };
+    return { items, nextCursor: hasMore && lastItem ? encodeCursor('createdAt', lastItem) : null };
   }
 
   /**
@@ -848,7 +830,7 @@ export class BeneficiaryRepository {
     items: { beneficiaryId: string; registrationDate: Date; eddDate: Date }[];
     nextCursor: string | null;
   }> {
-    const decodedCursor = cursor ? decodePostEddCursor(cursor) : null;
+    const decodedCursor = cursor ? decodeCursor('eddDate', cursor) : null;
 
     const rows = await this.prisma.beneficiaryCase.findMany({
       where: {
@@ -896,7 +878,7 @@ export class BeneficiaryRepository {
     const lastRow = rowsWithEdd[rowsWithEdd.length - 1];
     const nextCursor =
       hasMore && lastRow
-        ? encodePostEddCursor({ eddDate: lastRow.motherCaseDetails.eddDate, id: lastRow.id })
+        ? encodeCursor('eddDate', { eddDate: lastRow.motherCaseDetails.eddDate, id: lastRow.id })
         : null;
 
     return { items, nextCursor };
--- a/apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
+++ b/apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
@@ -50,6 +50,30 @@ function decodeCursor(cursor: string): ListCursor | null {
   }
 }
 
+interface PostEddCursor {
+  eddDate: string;
+  id: string;
+}
+
+/** Encodes findMotherIdsWithEddOnOrBefore's sort key (eddDate, id) as an opaque cursor. */
+function encodePostEddCursor(row: { eddDate: Date; id: string }): string {
+  const cursor: PostEddCursor = { eddDate: row.eddDate.toISOString(), id: row.id };
+  return Buffer.from(JSON.stringify(cursor)).toString('base64url');
+}
+
+/** Decodes a cursor produced by encodePostEddCursor; returns null on any malformed input. */
+function decodePostEddCursor(cursor: string): PostEddCursor | null {
+  try {
+    const parsed = JSON.parse(Buffer.from(cursor, 'base64url').toString('utf8'));
+    if (typeof parsed?.eddDate === 'string' && typeof parsed?.id === 'string') {
+      return parsed as PostEddCursor;
+    }
+    return null;
+  } catch {
+    return null;
+  }
+}
+
 /** Data-access layer for beneficiary cases. Only this domain touches these tables. */
 export class BeneficiaryRepository {
   constructor(private readonly prisma: PrismaService) {}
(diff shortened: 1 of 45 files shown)


apps/risk-referral-service/src/referrals/sakhi.client.ts
```diff
@@ -8,6 +8,34 @@ interface ApiSakhi {
...
+ * overdueFollowup.job.ts to resolve which Supervisor to notify for a
+ * beneficiary's overdue referral follow-up.
+ */
+export async function findSakhiById(
+  sakhiId: string,
+  authorizationHeader: string,
+): Promise<ApiSakhi | null> {
+  let res: Response;
+  try {
+    res = await fetch(`${API_GATEWAY_BASE_URL}/api/v1/sakhis/${sakhiId}`, {
```
palani-ND, line 23: This calls `GET /api/v1/sakhis/:sakhiId` through the gateway using the job's SYSTEM-role service token (`ServiceTokenClient` issues `roles: ['SYSTEM']`, confirmed in `auth.routes.ts`/`auth.service.spec.ts`). But `auth-service`'s route guard on that endpoint (`apps/auth-service/src/sakhis/sakhi.routes.ts`) is `requireRoles('SAKHI', 'SUPERVISOR', 'MANAGER', 'ADMIN')` — **`SYSTEM` isn't in the list**, and that file isn't touched by this PR.

Concretely: every tick of `runOverdueFollowupJob` that resolves a real beneficiary calls this function, gets a 403 back, and `findSakhiById` turns that into a thrown `badGateway` (line 31 below) — before `createEscalationEvent` is ever reached. That's caught by the per-followup `catch` in `overdueFollowup.job.ts` and logged as an error, silently, forever. As it stands, this PR's headline feature — auto-escalating overdue referral follow-ups — never actually raises an escalation or a notification in production; it just logs a failure once per overdue follow-up per day.

Needs either adding `SYSTEM` to that route's `requireRoles(...)` in auth-service, or (if that's out of scope for this PR) at minimum a heads-up in the description that a follow-up auth-service change is required before this can work.

apps/risk-referral-service/src/jobs/overdueFollowup.job.ts
```diff
@@ -0,0 +1,116 @@
...
+        {
+          beneficiaryId,
+          escalationType: 'REFERRAL_FOLLOWUP_MISSED',
+          referralId: followup.referralId,
+          assignedSupervisorId: supervisorId ?? undefined,
+        },
+        systemAccessToken,
+      );
+
+      if (supervisorId && event.status === 'OPEN') {
```
palani-ND: `event.status === 'OPEN'` is true both when `createEscalationEvent` just created a brand-new row and when it reused an already-open one for the same key (`EscalationRepository.createOrReuseOpen` returns the existing row, still `status: 'OPEN'`, with no "was this newly created" flag). Meanwhile `NotificationRepository.create` (notification-escalation-service) has no dedup at all.

Since this job never updates `followupStatus` on the rows it processes (only `ReferralRepository.updateStatus`, called from the decision route, does that), the same PENDING-and-overdue follow-up gets re-selected by `findOverduePendingFollowups` on every subsequent daily tick until a Supervisor actually resolves it. Each of those ticks re-enters this branch with `event.status === 'OPEN'` (the reused row) and fires a brand-new `createNotification` — so the assigned Supervisor gets a fresh "A referral follow-up is overdue and needs review" notification every single day the follow-up remains unresolved, not once.

Worth gating this on whether the escalation was actually newly created (e.g. have `createOrReuseOpen`/`EscalationService.create` return a `wasCreated` flag, or compare `createdAt`/`id` against a value threaded through) rather than on `status === 'OPEN'` alone.

apps/risk-referral-service/src/referrals/referral.repository.ts
```diff
@@ -162,6 +162,33 @@ export class ReferralRepository {
...
 
+  /**
+   * Job-only read (overdueFollowup.job.ts) — every PENDING follow-up whose
+   * `followupDate` has passed, system-wide (not scoped to a caller's roster
+   * — this job runs as a machine identity, not on behalf of any one
+   * Supervisor). Bounded by `take` so one tick can't try to process an
+   * unbounded backlog in a single run; a backlog larger than that just gets
+   * picked up on the next tick.
+   */
+  findOverduePendingFollowups(today: Date, take: number) {
```
palani-ND, line 181: The doc comment above this (line 170-171) says "a backlog larger than that just gets picked up on the next tick," but nothing in this job or repository ever changes `followupStatus` away from `PENDING` for the rows it processes — only the decision route's `updateStatus` does that, and that's a separate, Supervisor-driven flow.

So once the PENDING-and-overdue backlog exceeds `MAX_FOLLOWUPS_PER_TICK` (500, `overdueFollowup.job.ts`), `orderBy: followupDate asc` + `take: 500` returns the same oldest 500 rows on every tick, forever — not a rotating window. Rows ranked 501+ (the newer overdue ones) never surface at all until the backlog shrinks below 500, which it structurally can't since this job doesn't mark anything as handled. Combined with the notification-dedup issue flagged elsewhere in this PR, the practical effect is: the oldest 500 get a duplicate escalation notification every day, and everything past #500 is silently starved indefinitely.

apps/risk-referral-service/src/jobs/overdueFollowup.job.ts
```diff
@@ -0,0 +1,116 @@
...
+        ? await findSakhiById(beneficiary.sakhiId, authorizationHeader)
+        : null;
+      const supervisorId = sakhi?.supervisorId ?? null;
+
+      const event = await createEscalationEvent(
+        {
+          beneficiaryId,
+          escalationType: 'REFERRAL_FOLLOWUP_MISSED',
+          referralId: followup.referralId,
+          assignedSupervisorId: supervisorId ?? undefined,
```
palani-ND: `assignedSupervisorId: z.string().uuid()` on the escalation-event DTO is required, not `.optional()` — its own comment explains why ("an omitted value silently orphans the row from every Supervisor's feed"). But this line sends `supervisorId ?? undefined` whenever a beneficiary's Sakhi has no `supervisorId` set (or, independently, whenever `beneficiaryClient.getById` returns `null` for a stale/deleted beneficiary — `sakhi` short-circuits to `null` too). Since the request body is JSON, an `undefined` value drops the key entirely, and `.strict()` validation then rejects it as missing → this per-followup iteration throws, caught by the outer `catch`, and no escalation is raised for that follow-up at all.

This is currently masked by the SYSTEM-role gap flagged elsewhere in this PR (nothing gets this far today), but it'll become a live gap for unassigned/orphaned Sakhis as soon as that's fixed. Worth deciding: should an unassigned Sakhi's overdue follow-up escalate to nobody (skip, log, and `continue`) rather than attempting a call that's guaranteed to fail?

apps/risk-referral-service/src/jobs/overdueFollowup.job.ts
```diff
@@ -0,0 +1,116 @@
+import cron from 'node-cron';
+import { acquireJobLock, ServiceTokenClient } from '@armman/service-commons';
+import type { PrismaService } from '../prisma/prisma.service';
+import { ReferralRepository } from '../referrals/referral.repository';
+import { BeneficiaryClient } from '../referrals/beneficiary.client';
+import { findSakhiById } from '../referrals/sakhi.client';
+import { createEscalationEvent, createNotification } from '../referrals/systemEscalation.client';
+
+const JOB_NAME = 'referral-followup-overdue-escalation';
+const LOCK_DURATION_MS = 55 * 60 * 1000; // slightly under the default daily tick's own margin
```
palani-ND, line 10: `job-lock.ts`'s own doc comment is explicit: *"a consumer processing an unbounded or highly variable-sized batch should build its own renewal/heartbeat on top of this rather than assuming a fixed `lockDurationMs` is safe."* This job is exactly that case — up to `MAX_FOLLOWUPS_PER_TICK` (500) items, each doing 2-4 **sequential** awaited HTTP calls through the gateway (beneficiary fetch → sakhi fetch → escalation create → notification create), with no batching or concurrency and no lock renewal.

At even ~100-150ms/call that's a plausible several-minutes runtime under normal conditions, and a real risk of exceeding 55 minutes under a large first-deploy backlog or degraded downstream latency — at which point a second replica's next tick re-acquires the same lock and double-processes, which is the exact failure this lock exists to prevent (worse now given the duplicate-notification issue flagged elsewhere). Worth either running the per-item calls concurrently with a bounded pool, or adding the renewal/heartbeat the doc comment calls out.

apps/closure-reopen-service/src/closures/closure.service.ts
```diff
@@ -222,14 +268,24 @@ export class ClosureService {
     }
 
     try {
+      const [sakhiName, cardId] = await Promise.all([
+        this.resolveSakhiName(existing.submittedByUserId, authorizationHeader),
+        this.resolveQuickResponseCardId(id, authorizationHeader),
+      ]);
+      const beneficiaryName = beneficiary?.pii.fullName ?? null;
```
palani-ND: `getById`'s return type widened this PR from `BeneficiaryCaseRecord` to `BeneficiaryCaseDetail`, adding a required (non-optional) `pii: { fullName: string }`. Here and at the mirrored spot in `reopen-request.service.ts`, only `beneficiary` is optional-chained (`beneficiary?.pii.fullName`) — `.pii` itself isn't. If beneficiary-service's response ever omits `pii` (contract drift, a partial/degraded response), this throws, which is caught by the surrounding try/catch around the notify step, so the decision itself still succeeds — but the Sakhi silently gets no notification at all instead of the old, always-safe "Your ... request was approved" copy.

I couldn't find a live path in beneficiary-service that actually omits `pii` today, so this is latent rather than a confirmed active bug — `beneficiary?.pii?.fullName` would remove the risk for free.

apps/closure-reopen-service/src/closures/closure.service.ts
```diff
@@ -33,8 +34,48 @@ export class ClosureService {
     private readonly lookupClient: LookupClient,
     private readonly notificationClient: NotificationClient,
     private readonly beneficiaryClient: BeneficiaryClient,
```
palani-ND, line 40: `resolveSakhiName` and `resolveQuickResponseCardId` here are copy-pasted verbatim (including the doc comments) into the sibling `ClosureService` — both classes already take the same `SakhiClient`/`ApprovalClient`/`BeneficiaryClient` constructor deps, so this is in-PR duplication, not an inherited convention. A future change to the fallback behavior (e.g. how a missing card id is handled, or the log format) has to be made in both places and will drift the moment one is updated and the other isn't. Worth factoring both into a small shared helper (e.g. `decision-notification.helper.ts`) both services call.

apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
```diff
@@ -50,6 +50,30 @@ function decodeCursor(cursor: string): ListCursor | null {
   }
 }
 
+interface PostEddCursor {
+  eddDate: string;
+  id: string;
+}
+
+/** Encodes findMotherIdsWithEddOnOrBefore's sort key (eddDate, id) as an opaque cursor. */
+function encodePostEddCursor(row: { eddDate: Date; id: string }): string {
```
palani-ND: `PostEddCursor`/`encodePostEddCursor`/`decodePostEddCursor` here duplicate `ListCursor`/`encodeCursor`/`decodeCursor` 24 lines above in this same file — identical JSON-stringify/base64url-encode and parse/validate/null-on-malformed logic, differing only in the sort-key field name (`createdAt` vs `eddDate`). Since this file is already the largest in the module (951 lines, well past the repo's own ~250-line split guideline — the PR adds ~99 more lines to it), this was a natural spot to generalize instead: a `encodeCursor<T extends Record<string, string>>`/`decodeCursor<T>` keyed by field name would serve both `findMany` and `findMotherIdsWithEddOnOrBefore`, and the next cursor-paginated method wouldn't be tempted to add a third clone.

apps/risk-referral-service/src/main.ts
```diff
@@ -22,4 +30,7 @@ async function bootstrap(): Promise<void> {
   process.on('SIGINT', () => shutdown('SIGINT'));
 }
 
-void bootstrap();
+bootstrap().catch((err) => {
```
palani-ND: Good fix for the error-swallowing `void bootstrap()` bug — but the same 4-line `bootstrap().catch((err) => { console.error(...); process.exit(1); })` block is now pasted verbatim into 5 `main.ts` files in this PR (`api-gateway`, `audit-service`, `beneficiary-service`, `closure-reopen-service`, and here), and it already existed identically in `notification-escalation-service`/`visit-form-service` before this PR — 7+ copies of the same idiom repo-wide. `libs/service-commons` (whose whole purpose is centralizing this kind of cross-cutting concern) has no `bootstrap` helper. Worth adding `export function bootstrapService(fn: () => Promise<void>): void` there once and calling it from every `main.ts`, so a 6th service can't silently reintroduce the `void bootstrap()` bug this PR just fixed.

palani-ND:
## Code review

Left 9 inline comments across the 5-module split this PR combines. Two are worth calling out here since they compound:

**The headline feature (overdue-referral auto-escalation) can't fire as written.** `auth-service`'s `GET /sakhis/:sakhiId` route guard (`requireRoles('SAKHI', 'SUPERVISOR', 'MANAGER', 'ADMIN')`) doesn't include `SYSTEM`, but that's the only role the cron job's service token carries. Every call from `findSakhiById` 403s, which aborts each follow-up's processing before an escalation is ever created — verified by tracing the actual route guard, the token-issuance code, and the call site. Two independent review passes converged on this same finding independently, which is why I'm confident in it. That route lives in `auth-service`, outside this PR's changed files, so it couldn't be inline-commented directly — flagging it here.

**Once that's fixed, two more issues in the same job will surface:** `assignedSupervisorId: supervisorId ?? undefined` fails the escalation DTO's required-field validation for any Sakhi with no assigned Supervisor (inline comment on `overdueFollowup.job.ts:65`), and the "was this newly created?" gate for sending a notification is actually true on every re-processing of an unresolved follow-up — combined with `findOverduePendingFollowups` never marking anything as handled, this means the same oldest-500 backlog gets a duplicate notification every day, while anything past #500 is starved indefinitely (inline comments on `overdueFollowup.job.ts:70` and `referral.repository.ts:173`).

The rest are lower-severity: a fixed 55-minute job lock with no renewal, against `job-lock.ts`'s own documented warning for unbounded-batch jobs (`overdueFollowup.job.ts:10`); a latent null-safety gap from the widened `BeneficiaryCaseDetail` type (`closure.service.ts:275`); and a few duplication/simplification notes (shared `resolveSakhiName`/`resolveQuickResponseCardId`, a second cursor codec that could generalize the first, and the `bootstrap().catch()` block now pasted into 7+ `main.ts` files repo-wide with no shared helper in `service-commons`).

Everything else — auth guards, IDOR checks, role scoping, the job-lock wiring pattern, and all touched tests — looked correct and unweakened; the closure/reopen-request enrichment and beneficiary-service cursor pagination both check out against their own doc comments.

apps/risk-referral-service/src/referrals/sakhi.client.ts
logeshn-1: Fixed in 7efcf32 — added SYSTEM to GET /sakhis/:sakhiId's requireRoles(...) in auth-service. Confirmed this was the root cause: the job's service token carries only SYSTEM, so every call was 403ing and silently swallowed by the per-followup catch — the headline feature never actually fired.

apps/risk-referral-service/src/jobs/overdueFollowup.job.ts
logeshn-1: Fixed in 7efcf32 — createOrReuseOpen (notification-escalation-service) now returns a wasCreated flag, threaded through the POST response. The job now gates createNotification on event.wasCreated instead of status === 'OPEN', so a reused escalation no longer re-notifies on every tick.

apps/risk-referral-service/src/referrals/referral.repository.ts
logeshn-1: Fixed in 7efcf32 — added lastEscalatedAt to ReferralFollowup (new migration), stamped after each successful escalation, and findOverduePendingFollowups now orders by it (nulls first) instead of followupDate alone, so the backlog rotates through the whole set over time instead of repeating the same oldest 500 forever.

apps/risk-referral-service/src/jobs/overdueFollowup.job.ts
logeshn-1: Fixed in 7efcf32 — the job now skips (logs and continues) when no Supervisor can be resolved, rather than sending assignedSupervisorId: undefined and letting the DTO's .strict() validation reject the call.

apps/risk-referral-service/src/jobs/overdueFollowup.job.ts
logeshn-1: Addressed in 7efcf32 via bounded concurrency (processing up to 15 follow-ups in parallel per batch) rather than lock renewal — cuts the worst-case sequential runtime by roughly the concurrency factor, which should keep this well clear of LOCK_DURATION_MS without needing to extend job-lock.ts's shared API for every consumer.

apps/closure-reopen-service/src/closures/closure.service.ts
logeshn-1: Fixed in 7efcf32 — changed to beneficiary?.pii?.fullName in both closure.service.ts and reopen-request.service.ts.

apps/closure-reopen-service/src/closures/closure.service.ts
logeshn-1: Fixed in 7efcf32 — extracted both into a shared decision-notification.helper.ts, used by ClosureService and ReopenRequestService.

apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
logeshn-1: Fixed in 7efcf32 — generalized encodeCursor/decodeCursor to be generic over the sort-key field name, used by both findMany and findMotherIdsWithEddOnOrBefore.

apps/risk-referral-service/src/main.ts
logeshn-1: Fixed in 7efcf32 — added bootstrapService() to service-commons and switched all 5 main.ts files this PR touches (api-gateway, audit-service, beneficiary-service, closure-reopen-service, risk-referral-service) to use it.

PR #216, opened 2026-09-02
fix(notification-escalation-service): allow a sakhi to notify her own assigned supervisor

logeshn-1:
## Summary
- `POST /notifications` only allowed `ADMIN`/`SUPERVISOR`/`SYSTEM`, so `ApprovalRequestService.notifySupervisor()` forwarding the submitting Sakhi's own token got a silent 403 on every reopen/closure/LMP/referral/accompanied-referral submission — the Supervisor never received a notification.
- Widens the role guard to include `SAKHI`, and restricts a SAKHI caller to notifying only her own assigned Supervisor (resolved via her own Sakhi record, not the recipient's) — same ownership-check shape as the existing SUPERVISOR-notifying-her-own-Sakhi branch.
- Adds the missing downstream fetch timeout to approval-service's `NotificationClient`, matching every sibling client's convention.

Note: this branch is based on `feature/transfer-escalation`, so this diff includes that branch's full history relative to `develop` (48 commits) — the notification fix itself is the top commit, `85fe8fc`.

## Test plan
- [x] 26/26 unit tests pass in `notification.service.spec.ts` (4 new: SAKHI→own supervisor succeeds; SAKHI→wrong recipient/self/unresolvable-sakhi all 403)
- [x] 179/179 approval-service tests pass (unaffected, regression check)
- [x] Lint clean on both services
- [x] Build clean on both services
- [x] Live end-to-end repro: submitted reopen requests as a Sakhi → confirmed real `SUPERVISOR_APPROVAL_REQUESTED` notifications landed on the assigned Supervisor's `GET /notifications`, with correct title/body/linked Quick Response card

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

PR #217, opened 2026-09-02
fix(notification-escalation-service): allow a sakhi to notify her own assigned supervisor

logeshn-1:
## Summary
- `POST /notifications` only allowed `ADMIN`/`SUPERVISOR`/`SYSTEM`, so `ApprovalRequestService.notifySupervisor()` forwarding the submitting Sakhi's own token got a silent 403 on every reopen/closure/LMP/referral/accompanied-referral submission — the Supervisor never received a notification.
- Widens the role guard to include `SAKHI`, and restricts a SAKHI caller to notifying only her own assigned Supervisor (resolved via her own Sakhi record, not the recipient's) — same ownership-check shape as the existing SUPERVISOR-notifying-her-own-Sakhi branch.
- (Supersedes #216, which was built off `feature/transfer-escalation` and carried 96 unrelated files/48 commits, causing merge conflicts. This branch is built directly off current `develop` with only the 3 files this fix actually touches.)

## Test plan
- [x] 26/26 unit tests pass in `notification.service.spec.ts` (4 new: SAKHI→own supervisor succeeds; SAKHI→wrong recipient/self/unresolvable-sakhi all 403)
- [x] Lint clean
- [x] Build clean
- [x] Previously verified live end-to-end: submitted reopen requests as a Sakhi → confirmed real `SUPERVISOR_APPROVAL_REQUESTED` notifications landed on the assigned Supervisor's `GET /notifications`, with correct title/body/linked Quick Response card

### DIFF
--- a/apps/notification-escalation-service/src/notifications/notification.service.ts
+++ b/apps/notification-escalation-service/src/notifications/notification.service.ts
@@ -23,22 +23,32 @@ export class NotificationService {
   /**
    * ADMIN and SYSTEM may notify anyone — SYSTEM is a machine identity (the
    * missed-visit/referral-followup/sync-delay cron jobs' service-account
-   * token, per requireRoles('ADMIN', 'SUPERVISOR', 'SYSTEM') on this route)
-   * sending recipientUserId values that are Supervisors, not Sakhis, so the
-   * Sakhi-roster ownership check below doesn't apply to it (and would always
-   * 403 it, since sakhiClient.findById on a Supervisor id resolves to
-   * nothing). A SUPERVISOR (the role widened for approval-service's Quick
-   * Response decisions to forward through) may only notify a Sakhi actually
-   * assigned to them — verified via auth-service, same ownership check
-   * supervisor-operations-service already applies to its own Sakhi-scoped
-   * endpoints. Without this, the widened role would let any Supervisor
+   * token, per requireRoles('ADMIN', 'SUPERVISOR', 'SYSTEM', 'SAKHI') on
+   * this route) sending recipientUserId values that are Supervisors, not
+   * Sakhis, so the Sakhi-roster ownership check below doesn't apply to it
+   * (and would always 403 it, since sakhiClient.findById on a Supervisor id
+   * resolves to nothing). A SUPERVISOR (the role widened for
+   * approval-service's Quick Response decisions to forward through) may
+   * only notify a Sakhi actually assigned to them — verified via
+   * auth-service, same ownership check supervisor-operations-service
+   * already applies to its own Sakhi-scoped endpoints. A SAKHI (widened for
+   * ApprovalRequestService.notifySupervisor(), which forwards the
+   * submitting Sakhi's own token) may only notify her own assigned
+   * Supervisor — the reverse direction of the SUPERVISOR check, resolved by
+   * looking up the caller's own Sakhi record rather than the recipient's.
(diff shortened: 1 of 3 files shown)


apps/notification-escalation-service/src/notifications/notification.service.ts
```diff
@@ -23,22 +23,32 @@ export class NotificationService {
...
+   * looking up the caller's own Sakhi record rather than the recipient's.
+   * Without either check, the widened roles would let any Supervisor/Sakhi
    * notify any recipientUserId.
    *
    * For escalation/operational notification types, also best-effort fans
    * out a copy to the Sakhi's assigned Supervisor (see supervisor-fanout.ts).
    */
   async create(dto: CreateNotificationInput, caller: CallerIdentity, authorizationHeader: string) {
-    if (!caller.roles.includes('ADMIN') && !caller.roles.includes('SYSTEM')) {
+    if (caller.roles.includes('SAKHI')) {
```
palani-ND: **Correctness:** The new SAKHI check is evaluated *before* the ADMIN/SYSTEM bypass, so a caller whose JWT `roles` claim includes `SAKHI` alongside a higher-privilege role (e.g. `ADMIN` or `SUPERVISOR`) is wrongly routed into the restrictive "notify only her own assigned Supervisor" branch instead of being allowed through as ADMIN/SYSTEM, or evaluated under the SUPERVISOR ownership rule.

This is reachable: `UserRole` is one-to-many per `User` in `auth-service/prisma/schema.prisma`, and `AuthService.issueTokens` (`apps/auth-service/src/auth/auth.service.ts:435`) puts every active role code into the JWT's `roles` array without dedup/precedence — so a user with concurrent ADMIN+SAKHI or SUPERVISOR+SAKHI role assignments gets a multi-value `roles` claim. The codebase already anticipates and guards against exactly this in `AuthService.reactivateUser` (`apps/auth-service/src/auth/auth.service.ts:373-380`): `caller.roles.includes('SUPERVISOR') && !caller.roles.includes('MANAGER') && !caller.roles.includes('ADMIN')` — i.e. only apply the restrictive check when the caller doesn't *also* hold a higher-privilege role.

Here, that precedence isn't preserved: `if (caller.roles.includes('SAKHI'))` fires unconditionally, so an ADMIN/SUPERVISOR who also carries a SAKHI role assignment gets incorrectly 403'd (or silently mis-scoped) on `POST /notifications`, contradicting this function's own JSDoc ("ADMIN and SYSTEM may notify anyone").

Suggested fix: check `ADMIN`/`SYSTEM` (and arguably `SUPERVISOR`, mirroring `reactivateUser`'s pattern) first, and only fall into the SAKHI-ownership branch when none of those apply — e.g. reorder to `if (caller.roles.includes('ADMIN') || caller.roles.includes('SYSTEM')) { … } else if (caller.roles.includes('SAKHI')) { … } else { … }`.

logeshn-1: Good catch, confirmed — traced it through AuthService.issueTokens (no role precedence/dedup in the JWT claim) and reactivateUser's existing pattern, exactly as you described. Fixed in 6cb5fdd: ADMIN/SYSTEM now checked first, then SUPERVISOR, then SAKHI — matching reactivateUser's higher-privilege-first precedence. Added two regression tests covering a caller holding both ADMIN+SAKHI and SUPERVISOR+SAKHI roles simultaneously, confirming each gets routed to the correct branch.

PR #218, opened 2026-09-03
fix(approval-service): restore missing quick-response batch details endpoint

logeshn-1:
## Summary
- Restores `GET /quick-response/details?cardIds=uuid1,uuid2,...` — a batch endpoint the Supervisor app's Quick Response screen depends on to render card content after fetching the list via `GET /quick-response?status=PENDING`.
- This endpoint existed only on an abandoned, never-merged branch (`origin/quick-response-supervisor-notification`, commit `85fe8fc`) and was never part of `develop`'s history. It is **not a regression** — it's a gap that was never closed after that branch diverged.
- Registers `/quick-response/details` **before** `/quick-response/:cardId` in the router, since Express matches routes in registration order — this is the actual root cause of the reported symptom (`"cardId: Invalid uuid"` on any comma-joined batch request, because `:cardId`'s UUID validator was swallowing the `details` path segment).
- Batch resolution reuses the existing single-card `getCardDetail()` logic concurrently per id (`Promise.allSettled`), silently omitting any id that's not found, inaccessible to the caller, or fails enrichment — matching this service's existing "don't fail the page for one bad row" convention (`resolveSakhiNamesById`/`safeResolve`).
- Second commit fixes a duplicate `localRequestUuid` mock property left behind by cherry-picking onto a `develop` that had already added the same field independently via a separate, more recent migration (PR #212).

## Known issues / carried-over from manual QA (not fixed by this PR)
- **`sakhiId: Invalid uuid` in `SakhiClient.getManyByIds`** — `approval-service`'s `resolveSakhiNamesById` calls `GET /sakhis/by-ids?ids=...` on `auth-service`, but that route does not exist on `develop` — only `GET /sakhis/:sakhiId` does. The comma-joined `by-ids` literal gets swallowed by the `:sakhiId` UUID validator (same class of bug as the one this PR fixes, in a different service). This causes every Quick Response list response to have `sakhiName: null`. **Follow-up needed**: add `GET /sakhis/by-ids` to `auth-service`, mirroring `beneficiary-service`'s existing `GET /beneficiaries/by-ids-with-risk` convention, registered before `/sakhis/:sakhiId`.
- **Reactivate-on-approve 409s tolerated, not fixed** — approving a REOPEN card whose beneficiary is already `ACTIVE` (e.g. test data that was never actually closed) correctly 409s `beneficiaryClient.reactivateCase()`, which `ReopenRequestService.decide()` already catches and logs without failing the decision. Not a bug, just noting it surfaced repeatedly during QA against non-closed test beneficiaries.
- **Fire-and-forget `approval_requests` card creation can silently drop a card** — `closure-reopen-service`'s `create()` wraps its call to raise the linked Quick Response card in a try/catch that swallows any failure (lookup miss, approval-service transient error) — if that call fails, the `reopen_requests` row exists but never surfaces in the Supervisor's feed, with no retry/reconciliation. Observed intermittently during manual testing. Separate, larger fix (synchronous linkage or a reconciliation job) — out of scope here.
- **`decide()` never syncs back to `approval_requests.decisionStatusLookupId`** — deciding a reopen request via `closure-reopen-service`'s own `PATCH /reopen-requests/:id/decision` (bypassing approval-service's `/quick-response/:cardId/decision`) updates `reopen_requests.supervisorStatus` but never updates the linked `approval_requests` row, so the two tables can drift. Also out of scope here; needs its own design discussion.

## Test plan
- [x] `nx build approval-service` — succeeds
- [x] `nx lint approval-service` — clean (pre-existing unrelated warnings only)
- [x] `nx test approval-service` — 224/224 passing, including 5 new tests for the batch method (positive, dedupe, partial-not-found, partial-forbidden, empty-result) and 10 new DTO validation tests
- [x] Manually verified against a local dev backend: `GET /quick-response/details?cardIds=<2 uuids>` returns `200` with resolved card details; route-order regression guard confirmed (`/details` no longer 400s as an invalid `:cardId`)

### DIFF
--- a/apps/approval-service/src/quick-response/dto/get-quick-response-details.dto.ts
+++ b/apps/approval-service/src/quick-response/dto/get-quick-response-details.dto.ts
@@ -1,0 +1,52 @@
+import { z } from 'zod';
+
+/**
+ * Batch is a single query, not N concurrent per-card lookups, but still
+ * capped so a caller can't ask for an unbounded IN(...) list in one request
+ * — this is the entire fix for the Quick Response card-detail screen's
+ * gateway-overloading N×4 concurrent fan-out (see
+ * QuickResponseService.getCardDetails's doc comment). Kept below
+ * list-quick-response.dto.ts's page `limit` max of 100: the Supervisor app's
+ * card screen never opens more than a page's worth of cards at once.
+ */
+export const MAX_BATCH_CARD_IDS = 50;
+
+/**
+ * Query schema for `GET /quick-response/details?cardIds=...`. Kept as a
+ * plain string + `.refine()` (not `z.coerce.*`/`.transform()`), matching
+ * closure-reopen-service's decision-status-query.dto.ts and every other
+ * "by-ids" DTO in this codebase's own note on why — zod-to-openapi cannot
+ * introspect a `ZodEffects`-wrapped transform and crashes OpenAPI doc
+ * generation at startup. The split into an array happens in the controller
+ * via `parseCardIdsParam`.
+ */
+export const getQuickResponseDetailsSchema = z
+  .object({
+    cardIds: z
+      .string()
+      .trim()
+      .min(1)
+      .refine(
+        (v) => {
+          const ids = v.split(',');
+          return (
+            ids.length <= MAX_BATCH_CARD_IDS &&
+            ids.every((id) => z.string().uuid().safeParse(id.trim()).success)
+          );
+        },
+        {
+          message: `cardIds: must be a comma-separated list of at most ${MAX_BATCH_CARD_IDS} uuids`,
+        },
+      ),
+  })
+  .strict();
+
+export type GetQuickResponseDetailsInput = z.infer<typeof getQuickResponseDetailsSchema>;
+
+/** Splits the comma-separated `cardIds` query param into an array, trimming each entry. */
+export function parseCardIdsParam(cardIds: string): string[] {
+  return cardIds
+    .split(',')
+    .map((id) => id.trim())
+    .filter(Boolean);
+}
--- a/apps/approval-service/src/quick-response/dto/get-quick-response-details.dto.ts
+++ b/apps/approval-service/src/quick-response/dto/get-quick-response-details.dto.ts
@@ -1,13 +1,14 @@
 import { z } from 'zod';
 
 /**
- * Batch is a single query, not N concurrent per-card lookups, but still
- * capped so a caller can't ask for an unbounded IN(...) list in one request
- * — this is the entire fix for the Quick Response card-detail screen's
- * gateway-overloading N×4 concurrent fan-out (see
- * QuickResponseService.getCardDetails's doc comment). Kept below
- * list-quick-response.dto.ts's page `limit` max of 100: the Supervisor app's
- * card screen never opens more than a page's worth of cards at once.
+ * Caps how many card ids one batch request can ask for, so a caller can't
+ * pass an unbounded id list in one request. This only collapses the
+ * client→approval-service hop count from N to 1 — each card in the batch
+ * still fans out to ~4 downstream services during enrichment (see
+ * QuickResponseService.getCardDetails's doc comment), so downstream fan-out
+ * is not eliminated by this cap alone. Kept below list-quick-response.dto.ts's
+ * page `limit` max of 100: the Supervisor app's card screen never opens more
+ * than a page's worth of cards at once.
  */
 export const MAX_BATCH_CARD_IDS = 50;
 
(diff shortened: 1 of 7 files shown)


apps/approval-service/src/quick-response/quick-response.service.ts
```diff
@@ -339,6 +339,35 @@ export class QuickResponseService {
...
+   * Deliberately best-effort per id, mirroring safeResolve()'s "don't fail
+   * the page" contract: a card that's not found, inaccessible to this
+   * caller, or fails enrichment is silently omitted from the result rather
+   * than failing the whole batch — one bad id in a 6-card page shouldn't
+   * blank the other 5.
+   */
+  async getCardDetails(cardIds: string[], caller: CallerScope, authorizationHeader: string) {
+    const uniqueIds = Array.from(new Set(cardIds));
+    const settled = await Promise.allSettled(
+      uniqueIds.map((cardId) => this.getCardDetail(cardId, caller, authorizationHeader)),
```
palani-ND: `resolveOwnSakhiIds(caller, authorizationHeader)` inside `getCardDetail` depends only on `caller`/`projectId`, never on `cardId` — but fanning out `getCardDetail` per id here re-invokes it once per card. For a SUPERVISOR caller requesting a full 50-id batch (`MAX_BATCH_CARD_IDS`), that's up to 50 concurrent, byte-identical `GET /projects/:projectId/sakhis` calls to auth-service — 49 of them redundant. Worth resolving the roster once above this `.map()` and threading it through, rather than re-deriving it per id.

apps/approval-service/src/quick-response/quick-response.service.ts
```diff
@@ -339,6 +339,35 @@ export class QuickResponseService {
...
+    );
+    return settled
+      .filter(
+        (
+          result,
+        ): result is PromiseFulfilledResult<Awaited<ReturnType<typeof this.getCardDetail>>> =>
+          result.status === 'fulfilled',
```
palani-ND: This silently drops every rejected promise, not just the "not found / forbidden / enrichment failure" cases the doc comment above describes — an unexpected exception (e.g. a DB or downstream service outage inside `resolveCommonFields`) is swallowed the same way, with zero logging. Every other best-effort helper in this file (`safeResolve`, `resolveSakhiName`, `resolveOwnSakhiIds`, etc.) calls `console.error` before falling back to a safe default; this method doesn't, so a real infra failure affecting one id in a batch becomes invisible to monitoring while looking identical to a normal not-found case.

apps/approval-service/src/quick-response/dto/get-quick-response-details.dto.ts
```diff
@@ -0,0 +1,52 @@
+import { z } from 'zod';
+
+/**
+ * Batch is a single query, not N concurrent per-card lookups, but still
+ * capped so a caller can't ask for an unbounded IN(...) list in one request
+ * — this is the entire fix for the Quick Response card-detail screen's
```
palani-ND: This claim doesn't hold. `getCardDetails` still runs `Promise.allSettled` over N fully independent `getCardDetail()` calls, and each one still fans out to ~4 downstream services via the single-id lookups (`beneficiaryClient.getById`, `sakhiClient.getById`, `geographyClient.getById`, etc.) rather than the true batch endpoints (`beneficiaryClient.getManyWithRisk`, `sakhiClient.getManyByIds`) that `list()` in the same service already uses. Only the client→approval-service hop count actually dropped from N to 1 — downstream fan-out for a 50-id batch is still up to ~200 concurrent requests, same order of magnitude as before, just relocated server-side. Worth correcting this comment's scope, or wiring `getCardDetails` to the existing batch clients the way `list()` does.

apps/approval-service/src/quick-response/quick-response.service.ts
logeshn-1: Fixed — resolveOwnSakhiIds is now resolved once in getCardDetails and threaded through to each getCardDetail call via a new optional resolvedSakhiIds param, instead of being re-derived per id. See 371dd76.

apps/approval-service/src/quick-response/quick-response.service.ts
logeshn-1: Fixed — getCardDetails now distinguishes expected HttpError failures (not-found/forbidden/enrichment) from unexpected rejections. An unexpected error is now logged via console.error before the id is omitted from the batch, matching the rest of this file's best-effort helpers (safeResolve, resolveOwnSakhiIds, etc.). See 371dd76.

apps/approval-service/src/quick-response/dto/get-quick-response-details.dto.ts
logeshn-1: Good catch — the comment overclaimed. Reworded to say the batch cap only collapses the client→approval-service hop count from N to 1; it does not eliminate the downstream fan-out, which still happens per-card during enrichment. See 371dd76. Left getCardDetails on the single-id clients for now rather than wiring in the true batch clients (getManyWithRisk/getManyByIds) used by list() — that's a larger change to enrichApprovalRequestCard/resolveCommonFields's per-card-type branching and felt out of scope for this PR; happy to open a follow-up ticket for it if you agree.

PR #219, opened 2026-09-03
fix(approval-service): add quick response batch card-details endpoint

logeshn-1:
## Summary
- Cherry-picks the quick-response batch card-details endpoint (`GET /quick-response/details`) from PR #218 (`feature/cr-quick-response-details-batch-v2`) onto this branch, so it's available for local reopen-request/notification testing.
- Duplicate of #218's content — same 2 commits (2bb704b, c2a4428). If #218 merges first, this PR should be closed instead of merged to avoid a duplicate history.

## Test plan
- [x] `GET /quick-response/details?cardIds=...` verified locally against the dev Supabase DB, returns resolved card details for accessible ids, silently omits inaccessible/not-found ones per spec.
- [x] Confirmed via `apps/approval-service/src/quick-response/dto/get-quick-response-details.dto.spec.ts` and `quick-response.service.spec.ts` (carried over from the source branch).

### DIFF
--- a/apps/approval-service/src/quick-response/quick-response.service.ts
+++ b/apps/approval-service/src/quick-response/quick-response.service.ts
@@ -321,10 +321,18 @@ export class QuickResponseService {
    * built"). Closure Review's "completion tracker" is likewise omitted —
    * no backing data source was identified for it.
    */
-  async getCardDetail(cardId: string, caller: CallerScope, authorizationHeader: string) {
+  async getCardDetail(
+    cardId: string,
+    caller: CallerScope,
+    authorizationHeader: string,
+    resolvedSakhiIds?: string[] | null,
+  ) {
     const approvalRow = await this.repository.findById(cardId);
     if (approvalRow) {
-      const sakhiIds = await this.resolveOwnSakhiIds(caller, authorizationHeader);
+      const sakhiIds =
+        resolvedSakhiIds !== undefined
+          ? resolvedSakhiIds
+          : await this.resolveOwnSakhiIds(caller, authorizationHeader);
       if (sakhiIds && !sakhiIds.includes(approvalRow.requestedByUserId)) {
         throw forbidden('You do not have access to this Quick Response card.');
       }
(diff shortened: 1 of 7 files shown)


apps/approval-service/src/quick-response/quick-response.service.ts
```diff
@@ -339,6 +339,35 @@ export class QuickResponseService {
...
+   * Deliberately best-effort per id, mirroring safeResolve()'s "don't fail
+   * the page" contract: a card that's not found, inaccessible to this
+   * caller, or fails enrichment is silently omitted from the result rather
+   * than failing the whole batch — one bad id in a 6-card page shouldn't
+   * blank the other 5.
+   */
+  async getCardDetails(cardIds: string[], caller: CallerScope, authorizationHeader: string) {
+    const uniqueIds = Array.from(new Set(cardIds));
+    const settled = await Promise.allSettled(
+      uniqueIds.map((cardId) => this.getCardDetail(cardId, caller, authorizationHeader)),
```
palani-ND: Correctness/consistency: `getCardDetails` runs `getCardDetail` concurrently per id via `Promise.allSettled`, and each `getCardDetail` call independently invokes `resolveOwnSakhiIds` (which calls `sakhiClient.getOwnSakhiIds`). For a batch of N cards from the same non-privileged SUPERVISOR caller, that's N concurrent, identical HTTP calls for the exact same roster.

`resolveOwnSakhiIds` fails closed to `[]` on error (line ~228-231), so if any one of those N duplicate calls transiently fails (network blip, downstream rate limiting) while its siblings succeed, only that specific card's authorization check sees `sakhiIds = []` and gets rejected with `forbidden()` — even though the caller is genuinely authorized and every other card in the very same request resolved the roster successfully. The result is a non-deterministic, per-card omission from the batch response for a caller who should see all their own cards.

This also undercuts the endpoint's stated purpose (avoiding N× duplicate downstream fan-out) — the batch still performs N redundant roster fetches instead of resolving the roster once per batch call and reusing it across all requested ids.

Suggested fix: resolve `resolveOwnSakhiIds(caller, authorizationHeader)` once in `getCardDetails` and thread the result into the per-card checks, rather than letting each `getCardDetail` call re-derive it.

apps/approval-service/src/quick-response/quick-response.service.ts
```diff
@@ -339,6 +339,35 @@ export class QuickResponseService {
...
+    );
+    return settled
+      .filter(
+        (
+          result,
+        ): result is PromiseFulfilledResult<Awaited<ReturnType<typeof this.getCardDetail>>> =>
+          result.status === 'fulfilled',
```
palani-ND: Error-swallowing: this filter keeps only fulfilled results and silently drops every rejected one, with no distinction between `notFound()`/`forbidden()` (legitimate, documented "don't fail the page" cases) and an unexpected failure — e.g. `resolveCommonFields`'s `beneficiaryClient.getById` call is explicitly *not* wrapped in `safeResolve` (per that method's own doc comment, "that one is core... allowed to fail the request"), so a raw network/500 error from it propagates out of `getCardDetail` and is caught here the same as a 404.

For the single-card `GET /quick-response/:cardId` endpoint, that same failure surfaces as a 500 to the caller. For this batch endpoint, it's silently absorbed into a shorter 200 response with no indication a downstream system is failing — a caller can't tell "these ids don't exist / aren't yours" apart from "the beneficiary service is down right now", and a genuine outage look identical to normal partial results.

Worth at least logging (or otherwise surfacing) rejections whose error isn't a recognized not-found/forbidden case, so a systemic downstream failure doesn't disappear as quietly-fewer-cards.

apps/approval-service/src/quick-response/quick-response.service.ts
Dharanishwaran-ND: Fixed by merging in `develop`, which already carries this exact fix from #218 (commit 371dd76). `getCardDetails` now resolves `resolveOwnSakhiIds` once per batch call and threads it into each `getCardDetail` call via a new `resolvedSakhiIds` param, eliminating the N redundant roster fetches and the per-card transient-failure inconsistency you flagged.

apps/approval-service/src/quick-response/quick-response.service.ts
Dharanishwaran-ND: Fixed by the same merge (develop's 371dd76 from #218). The batch's fulfilled-only filter now logs any rejection that isn't an `HttpError` (i.e. not a recognized notFound/forbidden case) before dropping it, so an unexpected downstream failure like the unwrapped `beneficiaryClient.getById` call is now surfaced via console.error instead of silently looking like a normal not-found.

Dharanishwaran-ND:
Closing as redundant — this branch's content (the quick-response batch `getCardDetails` endpoint, including the fixes for both review comments above) already shipped to `develop` separately via #218 (commit 371dd76). After merging `develop` into this branch to resolve conflicts, the diff against `develop` is now empty (0 files changed), so there's nothing left for this PR to merge.

PR #220, opened 2026-09-03
feat(data-restore): restore beneficiary and visit data on sakhi reactivation

logeshn-1:
## Summary
- Adds a DATA_RESTORE approval flow: when a Supervisor approves reactivating a Sakhi, beneficiary-service and visit-form-service each expose a `PATCH /restore` endpoint that undoes soft-deletes (`isDeleted`/`deletedAt`) across the case, PII, mother/child details, consent, visit, schedule, submission, and answer tables for that Sakhi.
- approval-service's quick-response decide path calls both restore endpoints after reactivating the user, and writes the outcome (including partial failures) into the audit log rather than failing the whole approval.
- `DATA_RESTORE_*` added to audit-service's allowed action prefixes for the `SUPERVISOR` role, mapped to entity type `User`.

## Known issue — please review before merge
The new `PATCH /beneficiaries/restore` and `PATCH /visit-instances/restore` routes are gated only by `requireRoles('SUPERVISOR', 'SYSTEM', 'ADMIN')`, with **no roster/geography scope check**. Every other Supervisor-facing mutation in this same approval flow (e.g. `decideApprovalRequestCard`) explicitly verifies the deciding Supervisor owns the target Sakhi's roster before acting. As written, any authenticated Supervisor can restore another team's soft-deleted beneficiary/visit data by calling these endpoints directly with an arbitrary `sakhiUserId`, bypassing the roster-scoping invariant enforced everywhere else in this flow.

Suggested fix: add the same roster/ownership check used in `decideApprovalRequestCard` (or pass a pre-validated caller context) before either repository performs the restore.

Also worth a follow-up (non-blocking): if both downstream restore calls fail *and* the audit-log write itself fails, there is no durable record of the partial failure — only console logs. Consider whether that gap is acceptable for this rollout.

## Test plan
- [ ] Confirm Supervisor decision on a DATA_RESTORE card restores soft-deleted beneficiary + visit data end-to-end
- [ ] Confirm a Supervisor outside the Sakhi's roster is currently **not** blocked from calling the restore endpoints directly (reproduces the issue above)
- [ ] Confirm partial failure (one downstream restore fails) still returns 200 to the Supervisor with the failure recorded in the audit log's `afterJson`
- [ ] `npm run lint && npm run affected:test`

### DIFF
--- a/apps/approval-service/src/quick-response/quick-response.service.ts
+++ b/apps/approval-service/src/quick-response/quick-response.service.ts
@@ -916,7 +916,13 @@ export class QuickResponseService {
         authorizationHeader,
       );
     } else if (existing.requestType === 'DATA_RESTORE') {
-      result = await this.decideDataRestoreCard(cardId, existing, dto, authorizationHeader);
+      result = await this.decideDataRestoreCard(
+        cardId,
+        existing,
+        dto,
+        decidedByUserId,
+        authorizationHeader,
+      );
     } else if (existing.requestType === 'CLOSURE_REVIEW') {
       result = await this.decideClosureReviewCard(cardId, existing, dto, authorizationHeader);
     } else if (existing.requestType === 'REFERRAL_INCOMPLETE') {
@@ -1343,31 +1349,109 @@ export class QuickResponseService {
   }
 
   /**
-   * Decides a DATA_RESTORE card. Note this does NOT implement the SRS's
-   * literal FR-SV-4.6 wording ("data restore is initiated for that Sakhi's
-   * device") — that flow remains unconfirmed with ARMMAN and unbuilt. This
-   * implements a specifically-approved narrower behavior instead: on
-   * APPROVE, reactivates the requesting Sakhi's own user account
-   * (requestedByUserId) via auth-service's PATCH /users/:id/reactivate.
-   * Not tolerated — same rule as every other reactivation this service
-   * performs (LMP change, reopen, referral/closure decisions): a Supervisor
-   * who approved this needs to know if the reactivation didn't actually
-   * happen, not receive a false "success". REJECT makes no auth-service
-   * call and grants no account changes. Either way the Sakhi is notified,
-   * best-effort.
+   * Decides a DATA_RESTORE card. On APPROVE: reactivates the requesting
+   * Sakhi's own user account (requestedByUserId) via auth-service's
+   * PATCH /users/:id/reactivate, then restores her beneficiary and
+   * visit/form data via beneficiary-service's PATCH /beneficiaries/restore
+   * and visit-form-service's PATCH /visits/restore. REJECT makes none of
+   * these calls and grants no account/data changes.
+   *
+   * Note: nothing in this codebase currently creates a valid DATA_RESTORE
+   * card — there is no deactivation trigger anywhere that soft-deletes a
+   * Sakhi's data in the first place (auth-service only has the reverse,
+   * reactivateUser). This method is built ready for that trigger once
+   * product defines it; until then, approving a DATA_RESTORE card is a
+   * no-op restore (nothing was soft-deleted, so both restore calls report
+   * zero rows restored) beyond the account reactivation.
+   *
+   * User reactivation is NOT tolerated — same rule as every other
+   * reactivation this service performs (LMP change, reopen, referral/
+   * closure decisions): a Supervisor who approved this needs to know if the
+   * reactivation didn't actually happen, not receive a false "success".
+   *
+   * The two data-restore calls ARE tolerated (best-effort, independently),
+   * because unlike reactivateUser they run after the decision is already
+   * committed and the account is already reactivated — there is no
+   * cross-service transaction to roll back to (forklift architecture), so
+   * failing the whole request here would leave the account reactivated
+   * anyway while telling the caller the decision failed, which is worse
+   * than the alternative. Instead, each restore call's own success/failure
+   * is written verbatim into the audit-log entry's afterJson (queryable,
+   * durable — not just a console.error) so a partial failure is visible for
+   * manual follow-up rather than silently lost. Sakhi notification remains
+   * best-effort, matching every other card type's decide path.
    */
   private async decideDataRestoreCard(
     cardId: string,
     existing: { requestedByUserId: string },
     dto: DecideQuickResponseInput,
+    decidedByUserId: string,
     authorizationHeader: string,
   ) {
     if (dto.decision !== 'APPROVE' && dto.decision !== 'REJECT') {
       throw badRequest('decision: Must be APPROVE or REJECT for a DATA_RESTORE card.');
     }
 
+    let beneficiaryRestoreResult: { restoredCaseCount: number } | { error: string } | null = null;
+    let visitRestoreResult: { restoredVisitCount: number } | { error: string } | null = null;
+
     if (dto.decision === 'APPROVE') {
       await this.userClient.reactivateUser(existing.requestedByUserId, authorizationHeader);
+
+      try {
+        beneficiaryRestoreResult = await this.beneficiaryClient.restoreForSakhi(
+          existing.requestedByUserId,
+          authorizationHeader,
+        );
+      } catch (err) {
+        console.error(
+          `DATA_RESTORE card ${cardId} approved (account reactivated) but restoring ` +
+            `beneficiary data failed (needs manual follow-up):`,
+          err,
+        );
+        beneficiaryRestoreResult = { error: err instanceof Error ? err.message : String(err) };
+      }
+
+      try {
+        visitRestoreResult = await this.visitClient.restoreForSakhi(
+          existing.requestedByUserId,
+          authorizationHeader,
+        );
+      } catch (err) {
+        console.error(
+          `DATA_RESTORE card ${cardId} approved (account reactivated) but restoring ` +
+            `visit/form data failed (needs manual follow-up):`,
+          err,
+        );
+        visitRestoreResult = { error: err instanceof Error ? err.message : String(err) };
+      }
+    }
+
+    try {
+      if (dto.decision === 'APPROVE') {
+        await this.auditClient.log(
+          decidedByUserId,
+          'DATA_RESTORE_APPROVED',
+          'User',
+          existing.requestedByUserId,
+          { beneficiaryRestoreResult, visitRestoreResult },
+          authorizationHeader,
+        );
+      } else {
+        await this.auditClient.log(
+          decidedByUserId,
+          'DATA_RESTORE_REJECTED',
+          'User',
+          existing.requestedByUserId,
+          { decision: 'REJECTED', reason: dto.decisionNotes ?? null },
+          authorizationHeader,
+        );
+      }
+    } catch (err) {
+      console.error(
+        `DATA_RESTORE card ${cardId} was decided (${dto.decision}) but writing the audit entry failed:`,
+        err,
+      );
     }
 
     try {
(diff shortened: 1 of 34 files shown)


apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
```diff
@@ -1412,4 +1412,16 @@ export class BeneficiaryService {
...
+
+  /**
+   * Server-to-server only — see restore-for-sakhi.dto.ts and the
+   * repository method's own doc comment for the full rationale. No
+   * caller-scoping check here (unlike reactivateCase's
+   * assertCallerCanTouchCase): SYSTEM/ADMIN-only route access is the
+   * authorization boundary, since this restores everything a Sakhi owned
+   * rather than one caller-chosen case.
+   */
+  async restoreForSakhi(sakhiUserId: string) {
```
palani-ND: **Missing roster/geography scope check — any Supervisor can restore another team's soft-deleted beneficiary data (IDOR).**

`restoreForSakhi` is only gated by `requireRoles('SUPERVISOR', 'SYSTEM', 'ADMIN')` at the route (beneficiary.routes.ts) — there is no ownership/roster check, unlike every other mutation in this file. The controller doesn't even forward `req.user` to the service, so the service has nothing to scope with.

Compare `reactivateCase` (same file), whose own doc comment states: *"A SUPERVISOR caller may only reactivate a case in their own Sakhi roster ... this endpoint is reachable by a human Supervisor role, not just server-to-server, so it needs the same IDOR guard as any other single-case mutation."* `restoreForSakhi` is reachable by that exact same human SUPERVISOR role via the same kind of route gate, yet skips `assertCallerCanTouchCase` entirely.

**Failure scenario:** A Supervisor outside Sakhi X's roster calls `PATCH /api/v1/beneficiaries/restore` with `{ sakhiUserId: X }` directly (bypassing approval-service's properly-scoped decide flow, which does check `sakhiIds.includes(existing.requestedByUserId)`). `requireRoles` passes, `restoreForSakhi(sakhiUserId)` runs with no ownership check, and case/PII/mother/child/consent records for a Sakhi outside the caller's roster get restored.

Also directly violates `.claude/CLAUDE.md`'s security rule: *"RBAC + geography scope enforced on the server (`requireRoles` middleware), not just UI."*

(Self-disclosed by the PR author in the PR description as a known blocking issue — confirmed here by direct code inspection.)

apps/visit-form-service/src/visits/visitInstance.service.ts
```diff
@@ -453,6 +453,18 @@ export class VisitInstanceService {
...
+
+  /**
+   * Server-to-server only — see restore-for-sakhi.dto.ts and the
+   * repository method's own doc comment for the full rationale. No
+   * caller-scoping check here (unlike updateStatus's roster/own-record
+   * checks): SYSTEM/ADMIN-only route access is the authorization boundary,
+   * since this restores everything a Sakhi owned rather than one
+   * caller-chosen visit.
+   */
+  async restoreForSakhi(sakhiUserId: string) {
```
palani-ND: **Same missing-roster-scope defect as beneficiary-service's restore endpoint (IDOR), mirrored here.**

`restoreForSakhi` has no caller-scoping check, unlike `updateStatus` in this same file (which enforces "SAKHI own visits, SUPERVISOR only a visit whose Sakhi is assigned to her"). The route (visitInstance.routes.ts) is gated only by `requireRoles('SUPERVISOR', 'SYSTEM', 'ADMIN')`.

**Failure scenario:** Any Supervisor calls `PATCH /api/v1/visits/restore` with an arbitrary `sakhiUserId` outside their own roster and successfully restores that Sakhi's soft-deleted visit/schedule/form data — the same IDOR gap as the sibling beneficiary-service endpoint, just in visit-form-service.

Also violates `.claude/CLAUDE.md`'s rule: *"RBAC + geography scope enforced on the server (`requireRoles` middleware), not just UI."*

apps/approval-service/src/quick-response/quick-response.service.ts
```diff
@@ -1343,31 +1349,109 @@ export class QuickResponseService {
...
+        beneficiaryRestoreResult = { error: err instanceof Error ? err.message : String(err) };
+      }
+
+      try {
+        visitRestoreResult = await this.visitClient.restoreForSakhi(
+          existing.requestedByUserId,
+          authorizationHeader,
+        );
+      } catch (err) {
+        console.error(
```
palani-ND, line 1424: **Doc-comment overclaims durability of partial-restore-failure recording — it can still be silently lost.**

The rewritten doc comment above `decideDataRestoreCard` claims restore failures are recorded "verbatim into the audit-log entry's afterJson (queryable, durable — not just a console.error) so a partial failure is visible for manual follow-up rather than silently lost." But the audit-log write itself (lines ~1430-1455) is wrapped in its own try/catch that just `console.error`s on failure — it is not tolerant-with-a-durable-fallback, it's best-effort like everything else.

**Failure scenario:** `beneficiaryClient.restoreForSakhi` throws (beneficiary-service down) so `beneficiaryRestoreResult = { error: ... }`, and concurrently `auditClient.log(...)` for `DATA_RESTORE_APPROVED` also throws (audit-service down). The catch around the audit call swallows this — `decide()` still returns `{ decision: 'APPROVE' }`, the account is reactivated, and the only trace of the beneficiary-restore failure is a transient console line, exactly the outcome the doc comment says this design avoids.

(The PR body already flags a version of this as a non-blocking follow-up — this comment is about the doc comment's claim not matching what the code actually guarantees, for whoever reads it later as documentation of the guarantee.)

apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
```diff
@@ -930,4 +930,60 @@ export class BeneficiaryRepository {
...
+   * Deliberately does not touch `currentStatus` or write a
+   * BeneficiaryStatusHistory row: unlike reactivateCase() (a CLOSED->ACTIVE
+   * status transition on an otherwise-live row), this only reverses
+   * isDeleted/deletedAt on rows that were removed outright — a distinct
+   * operation as far as this service's own data model is concerned, and one
+   * plain enough that a case's `toStatus` here would be undefined. Audit
+   * logging for this action lives centrally in approval-service's decide
+   * path (AuditClient), matching how REOPEN's approval-level audit entry is
+   * not duplicated as a local status-history row either.
+   */
```
palani-ND, line 953: **Altitude: hardcoded per-table restore list duplicates a cascading soft-delete mapping that doesn't exist anywhere yet.**

This method (and its twin `VisitInstanceRepository.restoreForSakhi`) hardcodes its own list of tables to restore. As the doc comment on `decideDataRestoreCard` admits, there is currently no deactivation trigger anywhere that soft-deletes this data in the first place — confirmed by grep: no code in beneficiary-service or visit-form-service sets `isDeleted: true` today. So this restore logic and the (not-yet-built) soft-delete logic will each independently need to enumerate the same set of cascading tables, in two different places, with nothing linking them.

**Cost:** when the soft-delete trigger is eventually implemented, its author must independently re-derive this exact table list; a table added to one side and missed on the other means a future table stays permanently soft-deleted after a restore (or un-deleted when it shouldn't be), discovered only when a Sakhi notices missing/wrong data. Not blocking for this PR (nothing exercises it yet), but worth a shared single source of truth (e.g. a table-relationship map both the future delete-trigger and this restore read from) before the delete side is built.

apps/approval-service/src/quick-response/quick-response.service.ts
```diff
@@ -1343,31 +1349,109 @@ export class QuickResponseService {
...
       throw badRequest('decision: Must be APPROVE or REJECT for a DATA_RESTORE card.');
     }
 
+    let beneficiaryRestoreResult: { restoredCaseCount: number } | { error: string } | null = null;
+    let visitRestoreResult: { restoredVisitCount: number } | { error: string } | null = null;
+
     if (dto.decision === 'APPROVE') {
       await this.userClient.reactivateUser(existing.requestedByUserId, authorizationHeader);
+
+      try {
```
palani-ND, line 1404: **Minor: identical copy-paste try/catch for beneficiary vs. visit restore.**

The two restore try/catch blocks in `decideDataRestoreCard` (beneficiary just above, visit just below) are structurally identical: call client method, on error `console.error` a near-duplicate message and stash `{ error: message }`. Unlike the other card types' genuinely bespoke tolerance rules in this same file, this pair is pure duplication.

**Cost:** low today, but a third restore target (e.g. a future incentive-wages restore, if this flow grows) would triple this exact boilerplate rather than extracting a small local helper (e.g. `async function tryRestore(label, fn)`.) Not blocking — just flagging before a third copy gets pasted in.

apps/beneficiary-service/src/beneficiary/beneficiary.service.ts
Dharanishwaran-ND: Confirmed and fixed in ca93cf799e7316c459d2bbffb0bfc7294f09f71d. `restoreForSakhi` now takes `caller`/`authorizationHeader` and runs `assertCallerCanTouchCase(sakhiUserId, caller, authorizationHeader)` before restoring — a SUPERVISOR is now scoped to their own roster, SYSTEM/ADMIN remain unscoped. Controller updated to forward `req.user`. Added 403 (out-of-roster, no-project-scope) and 200 (in-roster) test cases to beneficiary.service.spec.ts.

apps/visit-form-service/src/visits/visitInstance.service.ts
Dharanishwaran-ND: Confirmed and fixed in ca93cf799e7316c459d2bbffb0bfc7294f09f71d, same pattern as the beneficiary-service sibling. `restoreForSakhi` now takes `caller`/`authorizationHeader`, checks the SUPERVISOR-roster membership of `sakhiUserId` via `listSakhiIdsForSupervisor` (mirroring `assertCallerOwnsBeneficiary`'s own SUPERVISOR branch), SYSTEM/ADMIN unscoped. Controller updated to forward `req.user`. Added matching 403/200 test cases to visitInstance.service.spec.ts.

apps/approval-service/src/quick-response/quick-response.service.ts
Dharanishwaran-ND: Confirmed — the audit-log write is indeed just best-effort (own try/catch, console.error only), so the doc comment's 'not just a console.error' claim was wrong. Corrected the wording in ca93cf799e7316c459d2bbffb0bfc7294f09f71d: it now states the restore failure is recorded in afterJson only when that audit-log write itself succeeds, and that a coincident audit-log failure leaves only the console line — no durable fallback claimed anymore. No behavior change, doc-only.

apps/beneficiary-service/src/beneficiary/beneficiary.repository.ts
Dharanishwaran-ND: Agreed this is a real forward-looking risk, but per the comment's own framing ('not blocking for this PR — nothing exercises it yet') I'm leaving this as-is rather than introducing a shared table-relationship map speculatively. Flagging for whoever builds the soft-delete trigger to read this comment thread and this method's doc comment before hardcoding a second, possibly-divergent table list.

apps/approval-service/src/quick-response/quick-response.service.ts
Dharanishwaran-ND: Agreed it's duplication, but per the comment's own framing ('not blocking — just flagging before a third copy gets pasted in') I'm leaving the two try/catch blocks as-is for this PR rather than extracting a helper for a pattern used only twice today. Worth revisiting (e.g. a small `tryRestore(label, fn)` helper) if/when a third restore target is added.

PR #222, opened 2026-09-04
feat(health-education): Map SRS health-education content to real triggers

### LINKED ISSUE #200: Contextual health-education content — mapping, stage vocabulary, precedence, and delivery mechanism — clarifications needed

Dharanishwaran-ND:
## Summary

Resolves #200's items 1 and 2 for the 5 risk-graded conditions. Per SRS FR-S-5.2(c)/FR-S-13, wires all 21 SRS-specified health-education conditions (`docs/Revised_App_Form_Final_20.3.26.xlsx.md`, "Health education message" table) to real ARMMAN content instead of the `COMING_SOON` placeholder — split into the SRS's two genuinely different mechanisms:

**Group A — risk-graded, "as soon as detected" (5 conditions):** Anemia, Gestational Hypertension, Gestational Diabetes, Inadequate Gestational Weight Gain, Previous Pregnancy Complication.
- Adds `RiskAssessment.riskPhase` (migration) so `GET /beneficiaries/:id/risk` can return the stage-appropriate message(s) for each condition — ANC-phase gets every non-postpartum message, PP-phase gets only the postpartum one (riskPhase alone can't distinguish finer ANC sub-stages without a GA-week calculation, which belongs to Group B).
- Corrects an earlier unmerged draft's condition-code map, which mistakenly used `DANGER_SIGNS`/`INFANT_DANGER_SIGNS` (Group B conditions) instead of the correct `BAD_OBSTETRIC_HISTORY`.
- `educationContent` is now an array (was a single nullable object) — **behavior change**, documented in the route's OpenAPI description.

**Group B — stage-based, not risk-graded (16 conditions):** Primigravida, Birth Preparedness, Danger Signs, Substance Use, Micronutrient Supplementation, Family Planning, Nutrition, Breastfeeding, Neonatal Care, Infant Care ×3, Malnutrition in Infants, POSTPARTUM Counselling, Post-loss, Dehydration.
- New `healthEducationStage.resolver.ts` gates each condition on gestational week, infant age in months, first-visit-only, or closure/delivery outcome (stillbirth on `DELIVERY_VISIT` or miscarriage/abortion on `ANC_CLOSURE_VISIT`) — resolved independently of any risk flag.
- New `stageEducationContent` field on `POST /forms/:formCode/submissions`'s response.

## Bugs found and fixed along the way (via live end-to-end testing)

1. **cms-content-service**: `healthEducation.controller.ts` never extracted `conditionLabel` from the query string — every `conditionLabel`-filtered call silently returned all 32 seeded messages instead of the 1-4 that matched. This blocked Group A entirely. Added controller-level tests (only service-level tests existed before, which couldn't catch this).
2. **tools/prisma-seed-foreach.js**: never applied each service's own `DATABASE_URL` schema override (unlike its sibling `prisma-foreach.js`), so `npm run seed` silently seeded against the wrong Postgres schema for every service with per-schema content. Fixed to mirror `prisma-foreach.js`'s `withSchema()` exactly.

## Out of scope (explicitly, not silently dropped)

- Risk grading is still not wired onto DELIVERY_VISIT/INFANT_VISIT/REFERRAL_VISIT — only DELIVERY_VISIT's specific stillbirth→Post-loss stage trigger is added here.
- Issue #200's items 3 (risk-vs-GA-based message precedence) and 4 (Marathi/media delivery mechanism) remain open — no rule/date exists yet to build against.

## Test plan

- [x] `npx nx run-many -t build test --projects=risk-referral-service,visit-form-service,cms-content-service` — all green (766 + 286 + 14 = 1066 tests)
- [x] `npm run migrate` — clean across all 15 services after this change
- [x] `npm run seed` — clean for every schema-affected service after the tooling fix
- [x] **Live-verified end-to-end** against the dev DB with a disposable SAKHI account and beneficiary (created/deleted via the real API, per repo convention): submitted a real `ANC_VISIT` with Hb=6.0 (SEVERE anemia) at 26 weeks gestational age — confirmed the risk-graded Anemia content (1 correct message) and all 4 applicable stage-based conditions (Birth Preparedness, Danger Signs, Substance Use, Dehydration) returned correctly, with 5 out-of-window conditions correctly excluded.

Closes #200 (partially — items 1/2 resolved, items 3/4 remain and are called out above).

### DIFF
--- a/apps/risk-referral-service/src/beneficiary-risk/beneficiaryRisk.service.ts
+++ b/apps/risk-referral-service/src/beneficiary-risk/beneficiaryRisk.service.ts
@@ -4,6 +4,7 @@ import { BeneficiaryClient } from './beneficiary.client';
 import { listSakhiIdsForSupervisor } from './sakhi.client';
 import { resolveRiskGrades } from './riskGrade.client';
 import { resolveEducationContent, type EducationContent } from './educationContent.client';
+import { resolveHealthEducationMessages } from './healthEducation.client';
 
 type StateSnapshotRow = Awaited<
   ReturnType<BeneficiaryRiskRepository['findStateSnapshots']>
@@ -12,6 +13,42 @@ type StateSnapshotRow = Awaited<
 type AssessmentWithFlagsRow = Awaited<
   ReturnType<BeneficiaryRiskRepository['findAssessmentsWithFlags']>
 >[number];
+
+/**
+ * RiskCondition.conditionCode -> HealthEducationMessage.conditionLabel, for
+ * the 5 SRS-specified "as soon as detected" risk-graded conditions
+ * (docs/Revised_App_Form_Final_20.3.26.xlsx.md, "Health education message"
+ * table, rows 1-5) that ARMMAN's delivered content actually covers. A flag
+ * whose conditionCode isn't listed here falls back to the COMING_SOON
+ * placeholder, same as before this map existed — this is additive, not a
+ * replacement for every condition. Every other condition in that same SRS
+ * table (Danger Signs, Neonatal Care, POSTPARTUM Counselling, etc.) is
+ * stage-based rather than risk-graded and is served by visit-form-service's
+ * health-education stage resolver instead — see that resolver's own doc
+ * comment for why those don't belong in this map.
+ */
+const CONDITION_CODE_TO_LABEL: Record<string, string> = {
+  ANEMIA: 'Anemia',
+  HYPERTENSION: 'Gestational Hypertension',
+  HYPERGLYCEMIA: 'Gestational Diabetes',
+  GESTATIONAL_WEIGHT_GAIN: 'Inadequate Gestational weight gain',
+  BAD_OBSTETRIC_HISTORY: 'Previous pregnancy complication',
+};
+
+/**
+ * Every seeded stage string for these 5 conditions contains the word
+ * "postpartum" exactly when — and only when — it's the PP-phase message
+ * (verified against health-education-messages.json's seed data). This is
+ * the only stage-precision riskPhase alone can support: ANC assessments
+ * can't be told apart into "as soon as detected" vs "2nd trimester" vs
+ * "3rd trimester" sub-stages without a gestational-week calculation (that
+ * belongs to the stage resolver, not this risk-flag path) — so an ANC-phase
+ * assessment gets every non-postpartum message for the condition, and a
+ * PP-phase assessment gets only the postpartum one.
+ */
+function isPostpartumStage(stage: string): boolean {
+  return stage.toLowerCase().includes('postpartum');
+}
 
 interface RiskConditionSummaryAcc {
   riskConditionId: string;
@@ -67,19 +104,67 @@ export class BeneficiaryRiskService {
     ]);
 
     const hasTriggeredFlag = assessments.some((a) => a.riskFlags.some((f) => f.isEducationTrigger));
-    // Resolved once per call, not once per flag — every isEducationTrigger
-    // flag maps to the same COMING_SOON topic today (see this feature's
-    // implementation plan doc for why there's no per-condition mapping yet).
+    // Resolved once per call, not once per flag — the COMING_SOON fallback
+    // is identical content regardless of which unmapped condition triggered
+    // it (mapped conditions get real content below instead).
     const comingSoonContent = hasTriggeredFlag
       ? await resolveEducationContent('COMING_SOON', authorizationHeader)
       : null;
 
+    // Resolved once per distinct (conditionCode, isPostpartum) pair across
+    // the whole call, not once per flag/assessment — several assessments
+    // commonly re-trigger the same condition (e.g. Anemia flagged at every
+    // ANC visit until it resolves), and each pair maps to the same content
+    // regardless of which assessment it came from. Caches the in-flight
+    // Promise itself, not just its resolved value — toAssessmentView below
+    // resolves every flag concurrently via Promise.all, so two flags for the
+    // same (conditionCode, phase) pair can both reach this function before
+    // either has finished; caching only the settled value would let both
+    // still call resolveHealthEducationMessages.
+    const contentCache = new Map<string, Promise<EducationContent[]>>();
+    function resolveMappedContent(
+      conditionCode: string,
+      riskPhase: string | null,
+    ): Promise<EducationContent[]> {
+      const conditionLabel = CONDITION_CODE_TO_LABEL[conditionCode];
+      if (!conditionLabel) return Promise.resolve(comingSoonContent ? [comingSoonContent] : []);
+
+      const isPostpartum = riskPhase === 'PP';
+      const cacheKey = `${conditionCode}:${isPostpartum}`;
+      const cached = contentCache.get(cacheKey);
+      if (cached) return cached;
+
+      const pending = (async (): Promise<EducationContent[]> => {
+        const messages = await resolveHealthEducationMessages(conditionLabel, authorizationHeader);
+        const filtered = riskPhase
+          ? messages.filter((m) => isPostpartumStage(m.stage) === isPostpartum)
+          : messages; // riskPhase null (pre-migration row) — return every message, undifferentiated.
+        const sorted = [...filtered].sort((a, b) => a.messageOrder - b.messageOrder);
+
+        return sorted.length > 0
+          ? sorted.map((m) => ({
+              topicCode: conditionCode,
+              topicName: m.titleEn ?? m.conditionLabel,
+              mediaType: m.mediaType,
+              contentUrl: m.mediaFile,
+            }))
+          : comingSoonContent
+            ? [comingSoonContent]
+            : [];
+      })();
+
+      contentCache.set(cacheKey, pending);
+      return pending;
+    }
+
+    const assessmentViews = await Promise.all(
+      assessments.map((assessment) => this.toAssessmentView(assessment, resolveMappedContent)),
+    );
+
     return {
       beneficiaryId,
       currentState: this.toCurrentStatePerPhase(snapshots),
-      assessments: assessments.map((assessment) =>
-        this.toAssessmentView(assessment, comingSoonContent),
-      ),
+      assessments: assessmentViews,
     };
   }
 
@@ -229,17 +314,15 @@ export class BeneficiaryRiskService {
   }
 
   /** Flattens each RiskFlag's nested RiskCondition into conditionCode/conditionName. */
-  private toAssessmentView(
+  private async toAssessmentView(
     assessment: AssessmentWithFlagsRow,
-    comingSoonContent: EducationContent | null,
+    resolveMappedContent: (
+      conditionCode: string,
+      riskPhase: string | null,
+    ) => Promise<EducationContent[]>,
   ) {
-    return {
-      id: assessment.id,
-      evaluatedAt: assessment.evaluatedAt,
-      overallRiskCategory: assessment.overallRiskCategory,
-      overallHighRiskFlag: assessment.overallHighRiskFlag,
-      hrDetectedFlag: assessment.hrDetectedFlag,
-      flags: assessment.riskFlags.map((flag) => ({
+    const flags = await Promise.all(
+      assessment.riskFlags.map(async (flag) => ({
         id: flag.id,
         conditionCode: flag.riskCondition.conditionCode,
         conditionName: flag.riskCondition.conditionName,
@@ -248,8 +331,19 @@ export class BeneficiaryRiskService {
         isReferralTrigger: flag.isReferralTrigger,
         isEducationTrigger: flag.isEducationTrigger,
         isHrVisitTrigger: flag.isHrVisitTrigger,
-        educationContent: flag.isEducationTrigger ? comingSoonContent : null,
+        educationContent: flag.isEducationTrigger
+          ? await resolveMappedContent(flag.riskCondition.conditionCode, assessment.riskPhase)
+          : [],
       })),
+    );
+
+    return {
+      id: assessment.id,
+      evaluatedAt: assessment.evaluatedAt,
+      overallRiskCategory: assessment.overallRiskCategory,
+      overallHighRiskFlag: assessment.overallHighRiskFlag,
+      hrDetectedFlag: assessment.hrDetectedFlag,
+      flags,
     };
   }
 }
--- a/apps/risk-referral-service/src/beneficiary-risk/beneficiaryRisk.service.ts
+++ b/apps/risk-referral-service/src/beneficiary-risk/beneficiaryRisk.service.ts
@@ -36,18 +36,26 @@ const CONDITION_CODE_TO_LABEL: Record<string, string> = {
 };
 
 /**
- * Every seeded stage string for these 5 conditions contains the word
- * "postpartum" exactly when — and only when — it's the PP-phase message
- * (verified against health-education-messages.json's seed data). This is
- * the only stage-precision riskPhase alone can support: ANC assessments
- * can't be told apart into "as soon as detected" vs "2nd trimester" vs
- * "3rd trimester" sub-stages without a gestational-week calculation (that
- * belongs to the stage resolver, not this risk-flag path) — so an ANC-phase
- * assessment gets every non-postpartum message for the condition, and a
- * PP-phase assessment gets only the postpartum one.
+ * The exact seeded `stage` string that marks the PP-phase message, for each
+ * of the 5 CONDITION_CODE_TO_LABEL conditions (verified against
+ * health-education-messages.json's seed data as of this writing — every
+ * other stage string for these conditions is an ANC sub-stage: "as soon as
+ * detected during ANC visit", "2nd trimester (6th month)", etc.). Matched
+ * exactly, not by substring — a looser `stage.includes('postpartum')` match
+ * was flagged in review as fragile against a future content reword (e.g.
+ * cms-content-service's seed data changing the exact phrasing), silently
+ * degrading a PP-phase beneficiary to the COMING_SOON placeholder with no
+ * error anywhere in the pipeline. This still shares that fragility (a
+ * reworded stage string breaks the exact match the same way it would break
+ * a substring match) — the real fix is a typed `phase` column on
+ * HealthEducationMessage in cms-content-service, mirroring the RiskPhase
+ * enum this PR already added to risk_assessments, which would make this
+ * function unnecessary. Tracked as a followup, not done here (cross-service
+ * migration, out of scope for this PR's review-comment fixes).
  */
+const POSTPARTUM_STAGE = 'postpartum (PP1 or PP2 whichever is attended)';
 function isPostpartumStage(stage: string): boolean {
-  return stage.toLowerCase().includes('postpartum');
+  return stage === POSTPARTUM_STAGE;
 }
 
 interface RiskConditionSummaryAcc {
@@ -129,11 +137,18 @@ export class BeneficiaryRiskService {
       const conditionLabel = CONDITION_CODE_TO_LABEL[conditionCode];
       if (!conditionLabel) return Promise.resolve(comingSoonContent ? [comingSoonContent] : []);
 
-      const isPostpartum = riskPhase === 'PP';
-      const cacheKey = `${conditionCode}:${isPostpartum}`;
+      // 3-way discriminant, not a collapsed isPostpartum boolean — riskPhase
+      // null (legacy pre-migration row, "return every message
+      // undifferentiated") and riskPhase 'ANC' (non-postpartum) both map to
+      // isPostpartum: false, which would otherwise share one cache bucket
+      // and let whichever resolves first silently overwrite the other's
+      // filtering intent for the rest of this call.
+      const phaseKey = riskPhase === null ? 'null' : riskPhase === 'PP' ? 'PP' : 'other';
+      const cacheKey = `${conditionCode}:${phaseKey}`;
       const cached = contentCache.get(cacheKey);
       if (cached) return cached;
 
+      const isPostpartum = riskPhase === 'PP';
       const pending = (async (): Promise<EducationContent[]> => {
         const messages = await resolveHealthEducationMessages(conditionLabel, authorizationHeader);
         const filtered = riskPhase
--- a/apps/risk-referral-service/src/beneficiary-risk/healthEducation.client.ts
+++ b/apps/risk-referral-service/src/beneficiary-risk/healthEducation.client.ts
@@ -1,6 +1,14 @@
 // Read directly (not via appConfig) — see beneficiary.client.ts for why.
 const API_GATEWAY_BASE_URL = process.env.API_GATEWAY_BASE_URL ?? 'http://localhost:3000';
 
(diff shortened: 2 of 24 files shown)


apps/risk-referral-service/src/beneficiary-risk/healthEducation.client.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+export async function resolveHealthEducationMessages(
+  conditionLabel: string,
+  authorizationHeader: string,
+  stage?: string,
+): Promise<HealthEducationMessage[]> {
+  try {
+    const params = new URLSearchParams({ conditionLabel });
+    if (stage) params.set('stage', stage);
+    const res = await fetch(
```
palani-ND, line 49: **Correctness (high): missing request timeout.**

This `fetch` call has no `AbortSignal.timeout`, unlike its sibling `educationContent.client.ts` in the same directory, which sets `REQUEST_TIMEOUT_MS = 3000` with an explicit comment: *"A slow (not down) dependency must degrade the same way an unreachable one does — without this, a hung gateway/cms-content-service call blocks GET /beneficiaries/:beneficiaryId/risk indefinitely."* The visit-form-service counterpart (`apps/visit-form-service/src/forms/healthEducation.client.ts`) does include the timeout guard — only this one omits it.

**Failure scenario:** cms-content-service or the gateway becomes slow (not down) while resolving one of the 5 mapped conditions' content — `GET /beneficiaries/:id/risk` hangs indefinitely instead of degrading gracefully.

Suggested fix: add the same `REQUEST_TIMEOUT_MS`/`AbortSignal.timeout` guard used in the sibling client.

apps/risk-referral-service/src/beneficiary-risk/beneficiaryRisk.service.ts
```diff
@@ -67,19 +104,67 @@ export class BeneficiaryRiskService {
...
+    const contentCache = new Map<string, Promise<EducationContent[]>>();
+    function resolveMappedContent(
+      conditionCode: string,
+      riskPhase: string | null,
+    ): Promise<EducationContent[]> {
+      const conditionLabel = CONDITION_CODE_TO_LABEL[conditionCode];
+      if (!conditionLabel) return Promise.resolve(comingSoonContent ? [comingSoonContent] : []);
+
+      const isPostpartum = riskPhase === 'PP';
+      const cacheKey = `${conditionCode}:${isPostpartum}`;
```
palani-ND: **Correctness (high): cache-key collision between riskPhase=null and riskPhase='ANC'.**

`resolveMappedContent`'s cache key is `${conditionCode}:${isPostpartum}` where `isPostpartum = riskPhase === 'PP'`. This is `false` for *both* `riskPhase === null` (legacy pre-migration rows, meant to return every message undifferentiated) and `riskPhase === 'ANC'` (meant to return only non-postpartum messages) — so both collapse onto the same cache bucket, e.g. `ANEMIA:false`.

**Failure scenario:** A beneficiary has two ANEMIA-triggering assessments in one `getRiskProfile()` call — one legacy row with `riskPhase: null` and one newer row with `riskPhase: 'ANC'`. `toAssessmentView` resolves all flags concurrently via `Promise.all`, so whichever flag reaches `resolveMappedContent` first populates the shared cache key for both:
- If the ANC flag resolves first, the null-phase flag wrongly loses its postpartum message (should get every message, undifferentiated).
- If the null-phase flag resolves first, the ANC flag wrongly gets the postpartum message mixed into its 'as soon as detected' view — contradicting the OpenAPI-documented per-phase contract.

No existing test exercises both phases for the same condition in the same call, so this gap isn't caught by CI. Suggested fix: key the cache on the actual `riskPhase` value (or a 3-way discriminant: 'PP' | 'other' | 'null') rather than the collapsed `isPostpartum` boolean.

apps/visit-form-service/src/forms/form.service.ts
```diff
@@ -623,6 +652,86 @@ export class FormService {
...
+  ): Promise<StageEducationContent[]> {
+    try {
+      const visitDateIso = submittedAt.toISOString();
+
+      let gestationalWeeks: number | undefined;
+      let isFirstVisitOfFormCode: boolean | undefined;
+      if (formCode === 'ANC_VISIT') {
+        const [registration, submissionCount] = await Promise.all([
+          this.resolveAncRiskRegistrationAnswers(dto.beneficiaryId),
+          this.repository.countSubmissionsByBeneficiaryAndFormCode(dto.beneficiaryId, 'ANC_VISIT'),
```
palani-ND: **Correctness (medium): `isFirstVisitOfFormCode` recomputed at replay time can drift from the truth at original-submission time.**

`countSubmissionsByBeneficiaryAndFormCode`'s own doc comment states its correctness depends on being called right after the current submission's own insert (so count==1 means 'this is the first visit'). That holds on the fresh-insert path, but on the idempotent-replay branch (line ~311-317) added by this PR, the same count is *recomputed* at replay time rather than reused from the original submission.

**Failure scenario:** A beneficiary's true first `ANC_VISIT` is submitted (count=1 at that time, so Primigravida's first-visit-only content is correctly included in the response). Later, more `ANC_VISIT` submissions land for the same beneficiary. A delayed/retried replay of the original `localSubmissionUuid` (e.g. an offline device re-syncing a queued request) now sees count>1, so `isFirstVisitOfFormCode` evaluates `false` and Primigravida content is silently dropped from the replay response even though this genuinely was the first visit.

apps/visit-form-service/src/forms/form.service.ts
```diff
@@ -623,6 +652,86 @@ export class FormService {
...
+      if (formCode === 'INC_VISIT') {
+        const beneficiary = await findBeneficiaryById(dto.beneficiaryId, authorizationHeader);
+        if (beneficiary?.childDateOfBirth) {
+          ageInMonths = ageInMonthsAt(beneficiary.childDateOfBirth, visitDateIso);
+        }
+      }
+
+      let closureReasonCode: string | undefined;
+      if (formCode === 'ANC_CLOSURE_VISIT') {
+        const reason = dto.formData.closure_reason;
```
palani-ND, line 782: **Correctness (medium): `closure_reason` is read without gating on `continue_with_closure`, unlike the established pattern for the same field elsewhere in this file.**

`resolveClosureRequest` (line 862) explicitly gates: `if (dto.formData.continue_with_closure !== 'yes') return;` before trusting `closure_reason`. This new code reads `dto.formData.closure_reason` directly with no such gate, and `validateSubmission`/`isVisible` doesn't strip values submitted for hidden fields.

**Failure scenario:** A Sakhi selects `continue_with_closure='yes'` and `closure_reason='miscarriage'`, then reconsiders and sets `continue_with_closure='no'` before submitting — but the mobile client doesn't clear the now-hidden `closure_reason` field (a common hidden-field-state UX gap). No closure is actually created (`resolveClosureRequest` correctly gates and returns early), but this code still reads `closure_reason` directly and fires 'Post miscarriage/abortion/still birth' health-education content for a beneficiary who was never actually closed for that reason — misleading content in the submission response.

apps/visit-form-service/src/forms/form.service.ts
```diff
@@ -623,6 +652,86 @@ export class FormService {
...
+    authorizationHeader: string,
+  ): Promise<StageEducationContent[]> {
+    try {
+      const visitDateIso = submittedAt.toISOString();
+
+      let gestationalWeeks: number | undefined;
+      let isFirstVisitOfFormCode: boolean | undefined;
+      if (formCode === 'ANC_VISIT') {
+        const [registration, submissionCount] = await Promise.all([
+          this.resolveAncRiskRegistrationAnswers(dto.beneficiaryId),
```
palani-ND: **Efficiency: duplicate DB read of the same MOTHER_REGISTRATION row within one request.**

`resolveAncRiskRegistrationAnswers(dto.beneficiaryId)` is called twice in one `createSubmission` invocation for `ANC_VISIT`: once at line 518 building the risk-trigger `answers` merge, and again here inside `resolveStageEducationContentForSubmission` just to extract `lmpDate`. Each call does its own `findLatestSubmissionByBeneficiaryAndFormCode` DB lookup plus identical field extraction from the same row.

**Cost:** every `ANC_VISIT` submission with a visitId and a configured `riskRuleSetId` (the standard case) does 2 identical DB round trips instead of 1, on the submission hot path.

Suggested fix: compute the registration answers once and thread the result (or just `lmpDate`) into `resolveStageEducationContentForSubmission` instead of re-fetching.

apps/risk-referral-service/src/beneficiary-risk/beneficiaryRisk.service.ts
```diff
@@ -13,6 +14,42 @@ type AssessmentWithFlagsRow = Awaited<
...
+ * "postpartum" exactly when — and only when — it's the PP-phase message
+ * (verified against health-education-messages.json's seed data). This is
+ * the only stage-precision riskPhase alone can support: ANC assessments
+ * can't be told apart into "as soon as detected" vs "2nd trimester" vs
+ * "3rd trimester" sub-stages without a gestational-week calculation (that
+ * belongs to the stage resolver, not this risk-flag path) — so an ANC-phase
+ * assessment gets every non-postpartum message for the condition, and a
+ * PP-phase assessment gets only the postpartum one.
+ */
+function isPostpartumStage(stage: string): boolean {
```
palani-ND, line 57: **Altitude: phase is derived by substring-matching free-text content instead of a typed field.**

`isPostpartumStage(stage) { return stage.toLowerCase().includes('postpartum'); }` reverse-engineers a structured fact (is this the postpartum-phase message?) from cms-content-service's free-text `stage` column. Notably, this same PR added a proper typed `RiskPhase` enum column to `risk_assessments` for the *beneficiary* side of 'phase' — but didn't mirror that pattern onto the *content* side.

**Failure scenario:** A future content edit to the seeded `stage` string (e.g. rewording 'postpartum (PP1 or PP2 whichever is attended)') breaks the substring match. A PP-phase beneficiary then silently gets zero filtered messages, falls through to the generic COMING_SOON placeholder, and nothing errors — `resolveHealthEducationMessages` swallows failures/empties by design, so there's no signal anywhere in the pipeline.

Suggested deeper fix: add a typed `phase` column to `HealthEducationMessage` in cms-content-service (mirroring the `RiskPhase` enum this PR already added to `risk_assessments`) and filter on that instead of string-matching.

apps/risk-referral-service/src/beneficiary-risk/healthEducation.client.ts
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * Deliberately swallows every failure and returns an empty array rather
+ * than throwing — same rationale as resolveEducationContent: this is
+ * reference content, not critical data, so a beneficiary's risk profile
+ * must stay readable even if cms-content-service is unreachable.
+ */
+export async function resolveHealthEducationMessages(
+  conditionLabel: string,
+  authorizationHeader: string,
+  stage?: string,
```
palani-ND: **Simplification: dead parameter.**

The `stage` parameter on `resolveHealthEducationMessages` is never passed by its only caller (`beneficiaryRisk.service.ts`), which instead does its own post-hoc JS filtering via `isPostpartumStage()`. The doc comment above this function claims `stage` is 'used for the riskPhase-is-null degradation case' — but that's not what actually happens; the caller always omits it and filters client-side.

Not a runtime bug, but the dead `if (stage) params.set('stage', stage)` branch and the misleading doc comment could lead a future maintainer to believe server-side filtering is engaged when it isn't. Consider removing the parameter or actually wiring it up as documented.

apps/visit-form-service/src/forms/form.service.ts
```diff
@@ -523,7 +541,18 @@ export class FormService {
...
+    // above, not gated on visitId/riskRuleSetId; resolveStageEducationContentForSubmission
+    // itself no-ops (returns []) for a formCode with no applicable stages.
+    const stageEducationContent = await this.resolveStageEducationContentForSubmission(
+      formCode,
+      dto,
+      created.submittedAt,
+      authorizationHeader,
+    );
+
+    return toApiFormSubmission(created, childBeneficiaryIds, stageEducationContent);
```
palani-ND, line 604: **Correctness (medium-high): the P2002 race-loser return path (line ~393, not shown in this diff's hunks) silently omits `stageEducationContent`.**

This PR wires `stageEducationContent` into the idempotent-replay return (line ~317) and this normal create return, but there's a *third* return path in `createSubmission` this PR didn't touch: the `catch` block's `P2002` unique-constraint race handler (`if (isPrismaErrorCode(err, 'P2002')) { ... if (existingRow) return toApiFormSubmission(existingRow); }`, around line 393). Since `toApiFormSubmission`'s third argument is only included in the response when not `undefined`, that path's response drops the `stageEducationContent` key entirely (not even an empty array), unlike this line and line 317.

**Failure scenario:** Two devices double-submit the same offline-queued `ANC_VISIT` with the same `localSubmissionUuid`. The winner's response (via this line) includes `stageEducationContent`; the loser's response (via the P2002 catch handler) is missing the key entirely — an inconsistent response shape for what should be an idempotent replay of the same logical submission. The existing P2002-race test only asserts `toMatchObject({ id: 'sub-winner' })`, so it doesn't catch the gap.

Suggested fix: resolve `stageEducationContent` in the P2002 catch handler too, mirroring this path and the replay path.

apps/visit-form-service/src/forms/form.service.ts
```diff
@@ -523,7 +541,18 @@ export class FormService {
       );
     }
 
-    return toApiFormSubmission(created, childBeneficiaryIds);
+    // SRS's 16 stage-based health-education conditions (Danger Signs,
+    // Neonatal Care, POSTPARTUM Counselling, etc.) — unlike risk grading
+    // above, not gated on visitId/riskRuleSetId; resolveStageEducationContentForSubmission
+    // itself no-ops (returns []) for a formCode with no applicable stages.
+    const stageEducationContent = await this.resolveStageEducationContentForSubmission(
```
palani-ND: **Efficiency: this resolution and the risk-assessment trigger above (line ~499-542, not shown in this diff's hunks) are independent but run sequentially.**

The risk-assessment trigger block above (its own DB read + outbound POST via `triggerRiskAssessment`) and this `resolveStageEducationContentForSubmission` call (its own DB reads + outbound GET(s) to cms-content-service) are data-independent of each other, but are `await`ed one after another rather than concurrently (e.g. via `Promise.all`).

**Cost:** on every visit-linked submission with a risk rule set configured, the Sakhi-facing submission response latency is the *sum* of both blocks' DB+HTTP time instead of the *max*, adding avoidable wall-clock delay with no correctness dependency between them.

apps/risk-referral-service/src/beneficiary-risk/healthEducation.client.ts
Dharanishwaran-ND: Fixed in 0da67ed621a8b19e2ea49bc4303287318458d651 — added the same `AbortSignal.timeout(REQUEST_TIMEOUT_MS)` guard (3s) that `educationContent.client.ts` already uses in this same directory. A slow cms-content-service/gateway now degrades to an empty array instead of hanging `GET /beneficiaries/:id/risk` indefinitely.

apps/risk-referral-service/src/beneficiary-risk/beneficiaryRisk.service.ts
Dharanishwaran-ND: Confirmed and fixed in 0da67ed621a8b19e2ea49bc4303287318458d651. Cache key is now a 3-way discriminant (`null | 'PP' | 'other'`) instead of the collapsed `isPostpartum` boolean, so `riskPhase === null` and `riskPhase === 'ANC'` no longer share a bucket. Added a regression test — I verified it fails against the old (reverted) logic and passes against the fix by literally reverting the cache-key change, re-running, confirming the failure, then restoring the fix and re-running clean.

apps/visit-form-service/src/forms/form.service.ts
Dharanishwaran-ND: Confirmed and fixed in 0da67ed621a8b19e2ea49bc4303287318458d651. `countSubmissionsByBeneficiaryAndFormCode` now takes an explicit `asOfCreatedAt` cutoff instead of counting against 'now' — the fresh-submission path passes the just-inserted row's own `createdAt`, the idempotent-replay path (and the P2002 race-loser path, same class of bug) passes the ORIGINAL row's `createdAt`, so `isFirstVisitOfFormCode` reflects the truth at original-submission time regardless of how much later a replay lands.

apps/visit-form-service/src/forms/form.service.ts
Dharanishwaran-ND: Confirmed and fixed in 0da67ed621a8b19e2ea49bc4303287318458d651 — `closure_reason` is now only trusted when `continue_with_closure === 'yes'`, matching `resolveClosureRequest`'s own established gate on this exact field. Added a regression test reproducing your hidden-field-state scenario (continue_with_closure='no' + a stale closure_reason='miscarriage' still present in formData) and confirmed Post-loss content is correctly NOT fired.

apps/visit-form-service/src/forms/form.service.ts
Dharanishwaran-ND: Confirmed and fixed in 0da67ed621a8b19e2ea49bc4303287318458d651. Registration answers are now resolved once (guarded to only run for ANC_VISIT) and threaded into both the risk-trigger `answers` merge and `resolveStageEducationContentForSubmission` via a new optional `preResolvedRegistrationAnswers` parameter, instead of each doing its own identical `findLatestSubmissionByBeneficiaryAndFormCode` lookup.

apps/risk-referral-service/src/beneficiary-risk/beneficiaryRisk.service.ts
Dharanishwaran-ND: Agreed on the underlying fragility, and agreed a typed `phase` column on `HealthEducationMessage` (mirroring the `RiskPhase` enum this PR added to `risk_assessments`) is the real fix — that's a cross-service migration on cms-content-service, which I don't think should ride along silently inside this PR's review-comment fixes. In 0da67ed621a8b19e2ea49bc4303287318458d651 I tightened the current approach from a loose `.includes('postpartum')` substring match to an exact match against the one seeded PP-phase stage string, which at least removes the 'partial reword still matches' false-positive risk, though it doesn't remove the 'exact reword breaks silently' risk you're flagging. Documented the deeper fix as a followup in the code comment. Happy to open a tracking issue for the typed-column migration if you'd like it tracked separately — let me know.

apps/risk-referral-service/src/beneficiary-risk/healthEducation.client.ts
Dharanishwaran-ND: Confirmed and fixed in 0da67ed621a8b19e2ea49bc4303287318458d651 — removed the dead `stage` parameter and the inaccurate doc-comment claim about it, and corrected the doc comment to describe what the function actually does (always returns every message for the condition, unfiltered; the caller does its own phase-appropriate filtering client-side via `isPostpartumStage`).

apps/visit-form-service/src/forms/form.service.ts
Dharanishwaran-ND: Confirmed and fixed in 0da67ed621a8b19e2ea49bc4303287318458d651 — the P2002 catch handler now resolves `stageEducationContent` the same way the sequential replay path does (using the existing row's own `submittedAt`/`createdAt`) before returning, instead of omitting the key entirely. Added a regression test using NEONATAL_VISIT (unconditional Group B content) that exercises this exact race-loser path and confirms the response now includes the correct content.

apps/visit-form-service/src/forms/form.service.ts
Dharanishwaran-ND: Confirmed and fixed in 0da67ed621a8b19e2ea49bc4303287318458d651 — the risk-trigger block and `resolveStageEducationContentForSubmission` now run concurrently via `Promise.all` (the risk-trigger block wrapped in an IIFE to preserve its early-return control flow inside the array). They're genuinely data-independent (neither reads the other's result), so response latency is now the max of the two instead of the sum.

PR #224, opened 2026-09-07
fix(visit-form-service): Correct ANC_VISIT USG follow-up fields' visibleWhen

### LINKED ISSUE #223: fix(visit-form-service): ANC_VISIT USG follow-up fields gated on wrong visibleWhen field
## Summary
The ANC_VISIT form schema's three USG follow-up fields — `if_yes_date_of_usg`, `type_of_usg`, `usg_finding` — have an incorrect `visibleWhen.field`, gated on `have_you_been_able_to_meet_the_beneficiary_for_the_visit` instead of `have_you_done_usg_since_last_visit`.

Per SRS skip logic: selecting "No" on "Have you done a USG since last visit?" should skip these three fields entirely. Currently they stay visible/answerable regardless of that answer.

## Root cause
An emergency fix (commit 431a15f, 2026-08-07) flattened a compound `visibleWhen` array condition (`[USG=yes, met_beneficiary=yes]`) down to a single condition, because mobile's parser couldn't handle array-form `visibleWhen`. The next day's cleanup (commit 40de159) restored the correct single-field gate on most affected fields, but missed these three — leaving them stuck on the emergency placeholder condition.

## Fix
`visibleWhen.field` on all three fields corrected to `have_you_done_usg_since_last_visit`. This is behaviorally equivalent to the original compound condition, since the parent question (`have_you_done_usg_since_last_visit`) is itself still gated on `have_you_been_able_to_meet_the_beneficiary_for_the_visit = yes`.

Confirmed (via apps/visit-form-service/src/forms/form-validation.ts) that submission-time required-field validation reads this same `visibleWhen` block — no separate fix needed for server-side validation.

## Reported by
Bharath (Sakhi App — Frontend), who shipped a client-side stopgap patch (`KnownSchemaGapPatch`) pending this server-side fix.

Dharanishwaran-ND:
## Summary

**⚠️ Scope note:** This PR now bundles the original ANC USG fix together with the `feature/cr-analytics-events-ingestion` branch (merged in) plus two additional, unrelated fixes. Reviewers should treat this as several independent changes rather than one cohesive diff — flagging for split-review or separate PRs if preferred.

### 1. ANC_VISIT USG `visibleWhen` fix (original scope)
- `if_yes_date_of_usg`, `type_of_usg`, and `usg_finding` in the ANC_VISIT form schema had `visibleWhen.field` incorrectly set to `have_you_been_able_to_meet_the_beneficiary_for_the_visit` instead of `have_you_done_usg_since_last_visit` — so selecting "No" on the USG question did not hide these fields, contradicting the SRS skip logic (Q52).
- Root cause: an emergency fix (431a15f, 2026-08-07) flattened a compound `visibleWhen` array condition on these fields down to a single condition, because mobile's parser couldn't handle arrays. The next day's cleanup (40de159) restored the correct field-specific gate on other affected fields but missed these three.
- Fix is behaviorally equivalent to the original compound condition, since the parent question (`have_you_done_usg_since_last_visit`) is itself still gated on `have_you_been_able_to_meet_the_beneficiary_for_the_visit = yes`.
- Confirmed submission-time required-field validation (`form-validation.ts`) reads this same `visibleWhen` block — no separate fix needed for server-side validation.
- Added a dedicated test (`seed-data.spec.ts`) matching this file's existing "one-step-removed gate" test pattern, and excluded the three fields from the blanket "every Q5+ field gates on met-beneficiary" assertion, same as other one-step-removed fields already documented there.
- Fixes #223

### 2. Merged: `feature/cr-analytics-events-ingestion`
Brings in the full Sakhi-app analytics-event ingestion feature (SRS Section 9), merged into this branch because change #3 below (an audit-service bug fix) only applies on top of it:
- `audit-service`: `POST /analytics/events/batch` (Sakhi-app event batch ingestion) and `GET /analytics/events` (SYSTEM-only, feeds reporting-etl-service's aggregation job), Prometheus metrics.
- `reporting-etl-service`: analytics-event client, metric-aggregation job, aggregated-metric repository/schema, enrollment metrics.
- `api-gateway`: routing for the new endpoints.
- Tooling: `prisma-foreach.js` `--service` filter fix, stale schema comment fix.

### 3. Fix: audit-service OpenAPI doc-generation crash
`audit-service` failed to boot with `MissingParameterDataError` — `GET /analytics/events`'s query schema is wrapped in `.refine()` (making it a `ZodEffects`, not a plain object), which `zod-to-openapi`'s query-param walker cannot introspect the same way. Fixed by passing an explicit `query:` doc override built from the unrefined base schema (same pattern already used by `visit-form-service`'s `visitHistoryQuerySchema` doc override) — runtime validation via `validate()` is unaffected; this only changes what's documented.

### 4. Fix: missing "Other, specify" fields on two forms
Two seed-data forms had an `"other"`/`"other_please_specify"` multiselect option with no corresponding free-text field to actually capture the specified value — added, following the same `REQUIRED_IF_SELECTED` conditional pattern already used elsewhere in these forms:
- `child-registration.json`: `did_the_baby_have_any_complications_at_the_time_of_birth_other_specify`
- `delivery-visit.json`: `did_mother_experience_complications_other_specify`, plus per-child (`child1`/`child2`/`child3`) `..._related_complications_other_specify`

## Test plan
- [x] `nx test visit-form-service --testPathPattern=seed-data` — 92/92 pass, including new dedicated USG gating test
- [x] `nx test visit-form-service` (full suite) — 769/769 pass
- [x] Verified USG fix independently on the dev environment (api.armman.org): created + patched + published ANC_VISIT v4 with the identical fix, confirmed `active-version` now serves the corrected `visibleWhen`, confirmed no collateral changes to any other field, confirmed the working precedent field (`if_yes_enter_date_of_latest_anc_visit_at_the_health_facility`) untouched.
- [x] Verified audit-service boots cleanly after the OpenAPI fix and `GET /analytics/events` appears correctly in the generated `/docs.json`.
- [ ] "Other, specify" fields (#4) not yet exercised end-to-end against a running stack — schema-level change following an established pattern in the same files.


apps/audit-service/src/analytics/analyticsEvent.service.ts
```diff
@@ -0,0 +1,151 @@
...
+ */
+function isSameLogicalEvent(
+  existing: ExistingAnalyticsEvent,
+  toCreate: AnalyticsEventInput,
+): boolean {
+  return (
+    existing !== null &&
+    existing.featureArea === toCreate.featureArea &&
+    existing.eventName === toCreate.eventName &&
+    existing.occurredAt.toISOString() === toCreate.occurredAt
```
palani-ND: **Correctness (high confidence): idempotency check is missing an actor check — cross-actor IDOR**

`isSameLogicalEvent` never compares `sakhiUserId`, even though `localEventUuid` is globally unique across ALL Sakhis (not per-actor). The sibling `auditLog.service.ts`’s `isSameLogicalEntry` explicitly compares `actorUserId` for exactly this reason (per its own doc comment, referencing a prior security review finding), and this function’s own doc comment claims to replicate that concern — but the implementation does not.

If two different Sakhis ever produce the same `localEventUuid` (buggy/non-cryptographic client UUID gen) and their events also share `featureArea`+`eventName`+`occurredAt`, the second Sakhi's event silently no-ops (reported as `created`, not `failed`) and her real event is permanently lost with no visible error.

Suggest adding `existing.sakhiUserId === toCreate.sakhiUserId` to the comparison (with a cross-actor test mirroring `auditLog.service.spec.ts`).

apps/audit-service/src/analytics/analyticsEvent.routes.ts
```diff
@@ -0,0 +1,170 @@
...
+        403: { description: 'Caller role not permitted', schema: apiErrorSchema },
+      },
+    },
+    trustGatewayIdentity,
+    // SAKHI-only: this endpoint captures Sakhi mobile-app usage per SRS
+    // Section 9's own scope ("Applies to: Sakhi Mobile App. Supervisor app
+    // metrics not defined in PRD.") — no Supervisor/Manager caller has
+    // events to report here.
+    requireRoles('SAKHI'),
+    validateBody(createAnalyticsEventBatchRequestSchema),
```
palani-ND: **Correctness: outer batch schema loses `.strict()`, silently accepting unknown top-level fields**

This route validates against a locally-built `createAnalyticsEventBatchRequestSchema` (defined above for OpenAPI examples), not the exported `.strict()` `createAnalyticsEventBatchSchema` from the DTO file. The per-event objects stay strict (`.extend()` preserves inner `unknownKeys`), but the outer `{ events: [...] }` wrapper does not, so `{"events":[...], "extraField":"x"}` now passes with `extraField` silently dropped instead of a 400 — contradicting the DTO file's own comment that `.strict()` matches every other DTO in this codebase. No test currently covers this (no unknown-top-level-field test in analyticsEvent.routes.spec.ts).

apps/audit-service/src/analytics/analyticsEvent.controller.ts
```diff
@@ -0,0 +1,27 @@
...
+export function createAnalyticsEventController(service: AnalyticsEventService) {
+  return {
+    createBatch: asyncHandler(async (req, res, next) => {
+      if (!req.user) return next(unauthorized());
+      const result = await service.createBatch(req.user.id, req.body);
+      // 201 even when some events in the batch failed (result.failed is
+      // non-empty) — the batch as a whole was accepted and processed; a
+      // per-event failure is reported in the body for the caller to inspect
+      // and retry just those, not a reason to fail the whole HTTP response.
+      res.status(201).json(ok(result));
```
palani-ND: **API contract: HTTP 201 returned even when every event in the batch fails**

When `result.created === 0` and all events land in `failed`, this still responds `201` with `success: true`. The comment above reasons about *some* events failing, but a caller that gates on HTTP status alone will treat a wholly-rejected batch as a successful ingest. Worth an explicit check for the all-failed case (e.g. a distinct status).

apps/reporting-etl-service/src/metrics/enrollmentMetrics.ts
```diff
@@ -0,0 +1,81 @@
...
+ * this is a contract with the (not-yet-built) mobile instrumentation, not
+ * something this function can enforce; an event missing sessionId is
+ * excluded from the completion-time calculation but still counted for
+ * drop-off rate and per-Sakhi counts.
+ */
+export function computeEnrollmentMetrics(events: AnalyticsEventRecord[]): EnrollmentMetrics {
+  const opens = events.filter((e) => e.eventName === ENROLLMENT_EVENT_NAMES.FORM_OPEN);
+  const submits = events.filter((e) => e.eventName === ENROLLMENT_EVENT_NAMES.FORM_SUBMIT);
+
+  const dropOffRate = opens.length > 0 ? (opens.length - submits.length) / opens.length : null;
```
palani-ND: **Correctness: `dropOffRate` is not clamped and can go negative**

Aggregation buckets events into fixed UTC calendar-day windows. A FORM_OPEN near 23:58 UTC can land in day N's window while its FORM_SUBMIT (synced later) lands in day N+1's — pushing `submits.length` above `opens.length` for a window and producing a negative `dropOffRate`, which is nonsensical for a rate metric and would corrupt `aggregated_metrics.value` / any downstream dashboard. Consider `Math.max(0, ...)` or excluding submits whose matched open falls outside the window.

apps/audit-service/src/analytics/analyticsEvent.service.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    for (const [index, event] of dto.events.entries()) {
+      try {
+        await this.createOne(sakhiUserId, event);
+        result.created += 1;
+      } catch (err) {
+        result.failed.push({
+          index,
+          localEventUuid: event.localEventUuid,
+          message: err instanceof Error ? err.message : 'Unknown error persisting this event.',
```
palani-ND: **Security/CLAUDE.md: raw error message can leak DB/internal error text to the client**

`err.message` is placed straight into the `failed[]` array of a normal 201 response, bypassing the global `errorHandler` middleware that would otherwise sanitize it. `createOne` only special-cases the P2002 path explicitly — any other exception (Prisma validation error, connection error, etc.) has its raw message returned to the caller, in tension with this repo's "never leak stack traces, DB errors, or internal fields to the client" rule.

apps/audit-service/src/analytics/analyticsEvent.service.ts
```diff
@@ -0,0 +1,151 @@
...
+   * localEventUuid is caught by the column's own unique constraint and
+   * resolved the same way as a sequential retry.
+   */
+  async createBatch(
+    sakhiUserId: string,
+    dto: CreateAnalyticsEventBatchInput,
+  ): Promise<AnalyticsEventBatchResult> {
+    const result: AnalyticsEventBatchResult = { created: 0, failed: [] };
+
+    for (const [index, event] of dto.events.entries()) {
```
palani-ND: **Efficiency: batch ingestion is fully sequential — no batched existence-check or bulk insert**

`createBatch` loops `for...of` with sequential `await this.createOne(...)`, and each `createOne` does an extra `await findByLocalEventUuid` SELECT before the INSERT when `localEventUuid` is present. A max-size batch (`MAX_BATCH_EVENTS`) can issue ~400 sequential round trips on one connection. Consider pre-fetching all existing rows for the batch's `localEventUuid`s in one `findMany({ where: { localEventUuid: { in: [...] } } })`, then `createMany({ data, skipDuplicates: true })` for the non-colliding new events, falling back to per-row logic only for actual collisions.

apps/audit-service/prisma/schema.prisma
```diff
@@ -42,3 +42,57 @@ model AuditLog {
...
+
+  createdAt DateTime @default(now()) @map("created_at")
+
+  // Idempotency key for this offline-first, batch-synced caller — a
+  // dropped-connection retry of a batch resubmits the same client-generated
+  // value per event. Same nullable-unique pattern as
+  // audit_log.local_audit_uuid / approval_requests.local_request_uuid.
+  localEventUuid String? @unique @map("local_event_uuid") @db.VarChar(80)
+
+  @@index([featureArea, eventName, occurredAt])
```
palani-ND: **Efficiency: composite index doesn't match the actual query pattern**

The ETL list query (`findByFeatureAreaAndWindow`) filters on `featureArea` (equality) + `occurredAt` (range) and orders by `(occurredAt, id)` for cursor pagination — it never filters on `eventName`. Because `eventName` sits between `featureArea` and `occurredAt` here, Postgres can use this index for the `featureArea` prefix but not for a contiguous `occurredAt` range scan or the `(occurredAt, id)` sort. Suggest `@@index([featureArea, occurredAt, id])` to match the real access pattern (add alongside this one if `eventName`-scoped queries are needed elsewhere).

palani-ND:
**Convention: `reporting-etl-service` now owns 4 Postgres tables, not the 2 its own `.claude/CLAUDE.md` authorizes**

That file states the service's entire data-ownership footprint as `etl_runs` and `report_exports` via Prisma. This PR adds `job_runs` (cron-lock table for the new aggregation job) and `aggregated_metrics` (computed SRS §9.9 metrics) — doubling the table count. `AggregatedMetric`'s own comment even concedes it's "NOT a permanent addition to this service's Postgres-only mandate." Please update that CLAUDE.md's ownership line to match, or reconsider whether these tables belong in this service.

---

**Ops gap: `apps/reporting-etl-service/.env.example` was not updated for the new service-account/cron vars**

`app-config.ts` adds `SERVICE_ACCOUNT_CLIENT_ID`, `SERVICE_ACCOUNT_CLIENT_SECRET` (both optional) and `ANALYTICS_AGGREGATION_JOB_CRON`, but `.env.example` is untouched — the only one of the three services adding this same service-token pattern in this PR whose `.env.example` wasn't updated (risk-referral-service and visit-form-service both list these vars). Because both vars are optional, a fresh environment deployed from `.env.example` boots fine but the job silently rejects on every cron tick with "SERVICE_ACCOUNT_CLIENT_ID/SECRET not configured" — the whole ETL feature never ingests any data, discoverable only via error logs, not a startup failure.

PR #226, opened 2026-09-15
feat(supervisor-operations-service): add groupId to inventory transactions, allow editing transactionType

logeshn-1:
## Summary

Adds a durable `groupId` to inventory transactions so multi-item submissions can be reliably grouped and later appended to, and allows `transactionType` to be edited via `PUT /inventory-transactions/{id}`. Both are extensions beyond what FR-SV-1.1/FR-SV-1 currently document in the SRS — flagging that up front for product visibility, not just as a technical change.

Related to #227 — tracks the open product sign-off questions below (does not auto-close on merge; #227 is a product decision, not something this code merging resolves).

## What changed

**1. `groupId` on inventory transactions** (schema + migration)
- New `group_id` column on `inventory_transactions`, backfilled to each row's own `inventory_txn_id` for existing data, indexed on `(sakhi_id, group_id)`.
- **Tradeoff, deliberately accepted:** the backfill does *not* attempt to reconstruct historical multi-item submissions from the old same-millisecond `createdAt` heuristic — that heuristic was never exact, so re-deriving groups from it risked silently merging unrelated rows or missing real ones. Pre-migration multi-item submissions therefore read as N separate one-row groups after this migration and can't be appended to as a group; only new submissions get a real, durable `groupId` going forward.

**2. `POST /inventory-transactions` — stamps a single groupId per submission**
Replaces the previous implicit "same-millisecond `createdAt`" grouping the client relied on (already flagged as fragile in the existing code comments) with an explicit, freshly generated UUID shared by every row created in one call.

**3. New `POST /inventory-transactions/{groupId}/items`**
Appends a new item line to an existing group without mutating any existing row (keeps the append-only-ledger model; `itemId` immutability per row is preserved). Header fields (`projectId`/`sakhiId`/`transactionType`/`transactionDate`) are inherited from the group and cannot be re-specified by the client. `supervisorId` on the appended row always reflects whoever performs the append (not the original submitter) — this was a real bug caught in review and fixed before this PR (see commit `ff2529b`).

**4. `GET /inventory-transactions/by-sakhi/{sakhiId}` — now returns `groupId`**
So the client can group cards by real `groupId` instead of reconstructing from timestamps.

**5. `PUT /inventory-transactions/{id}` — `transactionType` is now editable**
The update schema previously omitted `transactionType` (silently dropped if sent by the client) and documented it as immutable, matching this repo's append-only-ledger convention. This PR relaxes that specifically for `transactionType`, so an edit-form correction (e.g. HANDOVER meant to be RETURNED) is actually persisted.
- **Tradeoff, deliberately accepted:** the ledger keeps no history of what `transactionType` was before an edit — only `updatedAt`/`updatedByUserId` record that something changed and who changed it, not what it changed from.

## Not in scope / flagged, not fixed

- Ownership-check logic (privileged-bypass + Sakhi lookup + supervisor match) is duplicated across `createInventoryTransactions`, `appendInventoryTransactionItem`, and `listInventoryTransactionsBySakhi` — pre-existing pattern in the codebase before this diff, not introduced here.
- The new append route excludes `MANAGER` at `requireRoles`, matching the pre-existing `createInventoryTransactions` route — even though both service methods' `isPrivileged()` logic and their unit tests treat `MANAGER` as allowed. This is a pre-existing repo-wide inconsistency (7+ other endpoints have the same route/service mismatch), not something newly introduced.

## Product-adjacent decisions worth confirming

- **groupId/append feature**: FR-SV-1.1 only documents creating a transaction with all items chosen upfront in one submission — it says nothing about appending to an existing transaction later. Worth confirming with ARMMAN/product whether this is in scope, and whether a transaction group should ever be considered "closed" (e.g. a time window after which appends are no longer allowed) — there is currently no such limit.
- **transactionType editability**: contradicts this repo's own documented append-only-ledger convention for audit-relevant records. Worth confirming this is acceptable, or whether a wrong transaction type should instead be corrected via void + recreate (soft-delete + new row), which would preserve the audit trail without any schema/immutability change.

## Test plan

- [x] `npm run lint` — clean
- [x] `nx run supervisor-operations-service:build` — succeeds
- [x] `nx run supervisor-operations-service:test` — 247/247 pass (includes new DTO validation cases, service-layer ownership/authorization cases, and a regression test for the supervisorId-attribution bug found in review)
- [x] Manual end-to-end verification against a running local instance (via Postman/curl, with a real login): create multi-item transaction → confirm shared `groupId` → append an item → confirm same `groupId` and inherited header fields → list by sakhi → confirm all rows grouped → `PUT` a `transactionType` change → confirm persisted via re-fetch.

## Migrations / env vars

- New migration: `20260915090000_add_inventory_transaction_group_id` — must be applied (`prisma migrate deploy`) before this code is deployed, since it writes `groupId` unconditionally on every create/append.
- No new env vars.


logeshn-1:
Product-adjacent decisions from this PR are tracked in #227 — appending items to an existing transaction group, and editing `transactionType` after creation. Both need product/ARMMAN confirmation that they're in scope beyond FR-SV-1/FR-SV-1.1.

apps/supervisor-operations-service/src/operations/dto/update-inventory-transaction.dto.ts
```diff
@@ -1,17 +1,21 @@
...
+ * RETURNED) is possible; note the ledger keeps no history of what a row's
+ * transactionType previously was, only `updatedAt`/`updatedByUserId` that
+ * something changed and who changed it. At least one field must be present.
  */
 export const updateInventoryTransactionSchema = z
   .object({
     quantity: z.number().int().positive(),
     transactionDate: z.coerce.date(),
     remarks: z.string().trim().min(1).nullable(),
+    transactionType: z.enum(['HANDOVER', 'RETURNED', 'PERMANENT_DAMAGED', 'MISPLACED', 'CONSUMED']),
```
palani-ND, line 25: **Group's transactionType/transactionDate can silently diverge across rows in the same groupId.**

`appendInventoryTransactionItem` derives the group's shared header fields via `findInventoryTransactionGroupHeader(groupId)` in `operations.repository.ts`, which picks whichever row is currently the oldest non-deleted row (`orderBy: { createdAt: 'asc' }`) and treats its `transactionType`/`transactionDate` as authoritative for the whole group. But `transactionType` (new in this PR) and `transactionDate` (already editable) can both be changed per-row via `PUT /inventory-transactions/{id}`, which has no groupId/sibling-row awareness at all.

Failure scenario: a 2-item group is created (both rows HANDOVER). The supervisor later PUTs one row to RETURNED. If that row happens to be the group's 'header' (oldest surviving row), a subsequent append inherits RETURNED while the sibling row stays HANDOVER — the group now has mixed transactionTypes under one groupId, silently violating the 'a group represents one transaction event' invariant the append feature (and its own doc comments) rely on. If the edited row isn't the header, the group is inconsistent in the other direction and the edit is invisible to future appends. Nothing in the PUT or append code path guards against this.

(Independently verified: CONFIRMED)

apps/supervisor-operations-service/src/operations/inventory.routes.ts
```diff
@@ -191,6 +199,31 @@ export function registerInventoryRoutes(doc: DocumentedRouter, service: Operatio
...
+        },
+        400: { description: 'Validation error', schema: apiErrorSchema },
+        401: { description: 'Unauthenticated', schema: apiErrorSchema },
+        403: { description: 'Caller does not have access to this Sakhi', schema: apiErrorSchema },
+        404: { description: 'Transaction group not found', schema: apiErrorSchema },
+        422: { description: 'Referenced item not found or inactive', schema: apiErrorSchema },
+      },
+    },
+    trustGatewayIdentity,
+    requireRoles('SUPERVISOR', 'ADMIN'),
```
palani-ND: **MANAGER is blocked by the route guard on this new endpoint, contradicting the service logic and this PR's own tests.**

This new route gates on `requireRoles('SUPERVISOR', 'ADMIN')`, which 403s a MANAGER caller (via `forbidden()` in `rbac.guard.ts`) before the request ever reaches `controller.appendTransactionItem`/the service. But `appendInventoryTransactionItem`'s `isPrivileged(caller)` check (operations.service.ts) explicitly treats MANAGER as privileged/unrestricted, and this PR adds a new unit test — 'allows a MANAGER regardless of Sakhi assignment, without calling the Sakhi client' — that calls the service directly with a `managerCaller` and asserts success.

Failure scenario: a user with only the MANAGER role calls `POST /inventory-transactions/:groupId/items` in the real deployed API and gets a 403 at the route layer, even though the service code path (and the test suite added in this very PR) was written and verified to allow them through. The MANAGER-success behavior the new tests assert is unreachable via the actual HTTP route.

(Note: the PR description flags this same requireRoles/isPrivileged mismatch as a pre-existing issue on the sibling `POST /inventory-transactions` route — but this is a brand-new route in this PR that reproduces the mismatch, directly contradicted by tests added in this same PR.)

(Independently verified: CONFIRMED)

apps/supervisor-operations-service/prisma/schema.prisma
```diff
@@ -179,6 +188,7 @@ model InventoryTransaction {
 
   item InventoryItem @relation(fields: [itemId], references: [id])
 
+  @@index([sakhiId, groupId])
```
palani-ND: **New composite index doesn't serve the query it was added for.**

This migration adds `@@index([sakhiId, groupId])` (mirrored as `CREATE INDEX ... ON inventory_transactions(sakhi_id, group_id)` in the migration.sql), with a comment claiming it supports 'all rows in a group' and 'a sakhi's groups' lookups. But the actual new hot-path query, `findInventoryTransactionGroupHeader(groupId)` in `operations.repository.ts`, filters by `{ groupId, isDeleted: false }` alone — no `sakhiId` predicate — and runs on every single `POST /inventory-transactions/:groupId/items` request.

A standard B-tree composite index on `(sakhi_id, group_id)` can only be used efficiently for queries that filter on the leading column (`sakhi_id`), or on both columns together. A query filtering on `group_id` alone can't use this index as a targeted lookup — as the table grows, this hot-path query degrades toward a full scan. No other query added in this diff filters on `(sakhiId, groupId)` together to justify this column order.

Suggest `@@index([groupId])` or `@@index([groupId, sakhiId])` instead, so `groupId`-only lookups (the actual access pattern) can use it.

(Independently verified: CONFIRMED)

palani-ND:
**Conventions — stale OpenAPI summary (not on a diff line, posting as a general comment)**

`apps/supervisor-operations-service/src/operations/inventory.routes.ts`, the `doc.put('/inventory-transactions/:id', ...)` route (around line 296): `summary: "Edit a transaction's quantity/date/remarks"`.

Per `.claude/CLAUDE.md` §17: "API documented via OpenAPI in `api-contracts`; keep it the source of truth." This PR adds `transactionType` to `updateInventoryTransactionSchema` (the whole point of feature #5 in the description) but didn't update this route's summary text, so the generated OpenAPI doc now understates what the endpoint accepts.

apps/supervisor-operations-service/src/operations/dto/update-inventory-transaction.dto.ts
logeshn-1: Fixed in 0e83a48. `updateInventoryTransaction` now propagates a `transactionType`/`transactionDate` change to every other non-deleted row sharing the same `groupId`, atomically in one `$transaction`. `quantity`/`remarks` remain per-row only. Added repository-level tests (`operations.repository.spec.ts`) proving the propagation, since a service-level mock can't verify this — it's the repository method's own internal write shape.

apps/supervisor-operations-service/src/operations/inventory.routes.ts
logeshn-1: Fixed in 0e83a48 — changed `requireRoles('SUPERVISOR', 'ADMIN')` to `requireRoles('SUPERVISOR', 'MANAGER', 'ADMIN')` on this route. Agreed this is different from the pre-existing sibling route: this route and its MANAGER-success test are both new in this PR, so there's no reason to leave them contradicting each other — fixing the route to match the tested behavior rather than leaving a dead code path.

apps/supervisor-operations-service/prisma/schema.prisma
logeshn-1: Fixed in 0e83a48. Replaced `@@index([sakhiId, groupId])` with two single-column indexes — `@@index([groupId])` and `@@index([sakhiId])` — matching the two real access patterns (`findInventoryTransactionGroupHeader` filters on `groupId` alone; `findInventoryTransactionsBySakhi` filters on `sakhiId` alone; neither filters on both together). Since the original composite index was already applied to the shared dev DB, this went through a new migration (`20260916060000_fix_inventory_transaction_group_id_index`) rather than editing the already-applied one in place.

logeshn-1:
Fixed in 0e83a48 — the PUT route's OpenAPI `summary` now reflects that `transactionType` (and by extension the group-wide propagation added in this same commit) is editable, not just quantity/date/remarks.

